PR #1, opened 2018-09-07
[ENG-416] Vectorize and simplify raster predictions

feynmanliang:


danielrasmuson:
Tested and it worked for me. Nice work. 👏

PR #20, opened 2019-11-26
Add Project.bulk_delete_labels.

florijanstamenkovic:
Several issues unresolved:
* Now there are two ways to bulk delete `Label` objects.
* The newly added method's `label_contains` filter does not work as I'd expect.
* Changelog not updated (need to decided what stays first).


rllin:
closed for now, but we can reopen if we want when there is a need

PR #27, opened 2020-07-10
Added BulkImportRequest integration

gszpak:
Changes:
- Added `BulkImportRequest` entity
- Implemented method to create `BulkImportRequest` from python dictionaries
- Implemented method to upload local ndjson file with predictions to Labelbox' object storage

### DIFF
--- a/labelbox/schema/bulk_import_request.py
+++ b/labelbox/schema/bulk_import_request.py
@@ -1,11 +1,4 @@
-import json
-from pathlib import Path
-from typing import Iterable
-
-import ndjson
-
 from labelbox import Client
-from labelbox.exceptions import LabelboxError
 from labelbox.orm import query
 from labelbox.orm.db_object import DbObject
 from labelbox.orm.model import Field
@@ -23,19 +16,16 @@ class BulkImportRequest(DbObject):
     status_file_url = Field.String("status_file_url")
     state = Field.Enum(BulkImportRequestState, "state")
 
-    @staticmethod
-    def create(
-            client: Client, project_id: str, name: str,
-            predictions: Iterable[dict]) -> 'BulkImportRequest':
-        data_str = '\n'.join(json.dumps(prediction) for prediction in predictions)
-        data = data_str.encode('utf-8')
-        input_file_url = client.upload_data(data)
+    @classmethod
+    def create_from_url(
+            cls, client: Client, project_id: str, name: str,
+            url: str) -> 'BulkImportRequest':
         query_str = """
         mutation CreateBulkImportRequestPyApi {
             createBulkImportRequest(data: {
-                projectId: %s,
-                name: %s,
-                fileUrl: %s
+                projectId: "%s",
+                name: "%s",
+                fileUrl: "%s"
             }) {
                 %s
             }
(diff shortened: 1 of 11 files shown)


labelbox/schema/bulk_import_request.py
```diff
@@ -0,0 +1,76 @@
...
+            local_predictions_file_path (str): local NDJSON file containing predictions
+        Returns:
+            A URL of uploaded NDJSON file
+        Raises:
+            LabelboxError: if local file is not a valid NDJSON file
+        """
+        with local_predictions_file_path.open("rb") as f:
+            try:
+                data = ndjson.load(f)
+                return client.upload_data(
```
rllin, line 72: nit: move to after except or put under a else

labelbox/schema/bulk_import_request.py
```diff
@@ -0,0 +1,76 @@
...
+from labelbox.exceptions import LabelboxError
+from labelbox.orm import query
+from labelbox.orm.db_object import DbObject
+from labelbox.orm.model import Field
+from labelbox.orm.model import Relationship
+from labelbox.schema.enums import BulkImportRequestState
+from labelbox.schema.enums import UploadedFileType
+
+
+class BulkImportRequest(DbObject):
```
rllin, line 17: i think worth adding tests for this

gszpak: Definitely:) Will work on this after API pr is merged

labelbox/schema/bulk_import_request.py
```diff
@@ -0,0 +1,151 @@
...
+            client, request_data, file_name, file_data)
+        return BulkImportRequest(client, response_data["createBulkImportRequest"])
+
+    @classmethod
+    def create_from_local_file(
+            cls, client: Client, project_id: str, name: str,
+            file: Path) -> 'BulkImportRequest':
+        file_name = cls.__make_file_name(project_id, name)
+        request_data = cls.__make_request_data(project_id, name, file_name)
+        with file.open('rb') as f:
```
rllin, line 73: are we polluting user's local fs?  change to tempfile https://docs.python.org/3.7/library/tempfile.html to self clean up?

rllin, line 73: if we are exposing this method, let's throw in a quick ndjson.load to validate the file format?

we get enough cases where each annotation row is split across multiple lines, I think this will be a big win for cheap.

labelbox/schema/bulk_import_request.py
```diff
@@ -0,0 +1,151 @@
...
+            query.results_query_part(BulkImportRequest)
+        )
+        bulk_import_request_kwargs = client.execute(query_str)["createBulkImportRequest"]
+        return BulkImportRequest(client, bulk_import_request_kwargs)
+
+    @classmethod
+    def create_from_objects(
+            cls, client: Client, project_id: str, name: str,
+            predictions: Iterable[dict]) -> 'BulkImportRequest':
+        data_str = '\n'.join(json.dumps(prediction) for prediction in predictions)
```
rllin, line 58: nit: if you're installing the `ndjson`, should we just use that to write?

labelbox/schema/bulk_import_request.py
```diff
@@ -0,0 +1,335 @@
...
+        periodically refreshing object's state.
+
+        Args:
+            sleep_time_seconds (str): a time to block between subsequent API calls
+        """
+        while self.state == BulkImportRequestState.RUNNING:
+            logger.info(f"Sleeping for {sleep_time_seconds}...")
+            self.__exponential_backoff_refresh()
+
+    @backoff.on_exception(
```
rllin, line 206: did you find that `backoff` is better suited than google's `retry`?  https://googleapis.dev/python/google-api-core/latest/retry.html

labelbox/schema/bulk_import_request.py
```diff
@@ -180,12 +183,39 @@ def get(cls, client: Client, project_id: str, name: str) -> 'BulkImportRequest':
...
-       Synchronizes values of all fields with the database.
-       """
+        Synchronizes values of all fields with the database.
+        """
         bulk_import_request = self.get(self.client, self.project().uid, self.name)
         for field in self.fields():
             setattr(self, field.name, getattr(bulk_import_request, field.name))
 
+    def wait_until_done(
+            self, sleep_time_seconds: int = 30) -> None:
```
rllin, line 193: where does `sleep_time_seconds` get used?  i only see it in the log line

labelbox/schema/bulk_import_request.py
gszpak: Worked pretty well - that was the first thing I saw when I googled "exponential backoff python", tried it with local API down and logs seemed reasonable

labelbox/schema/bulk_import_request.py
gszpak: Whoops. Good catch!

setup.py
```diff
@@ -5,15 +5,15 @@
...
+    version="2.4.3",
     author="Labelbox",
     author_email="engineering@labelbox.com",
     description="Labelbox Python API",
     long_description=long_description,
     long_description_content_type="text/markdown",
     url="https://labelbox.com",
     packages=setuptools.find_packages(),
-    install_requires=["requests>=2.22.0"],
+    install_requires=["backoff>=1.10.0", "ndjson>=0.3.1", "requests>=2.22.0"],
```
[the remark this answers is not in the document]
rllin: addressed this myself :)

PR #30, opened 2020-07-30
[BACKEND-829] fix content length

rllin:
fixing content length being calculated incorrectly


labelbox/client.py
```diff
@@ -188,21 +188,33 @@ def upload_file(self, path):
         content_type, _ = mimetypes.guess_type(path)
         basename = os.path.basename(path)
         with open(path, "rb") as f:
-            return self.upload_data(data=(basename, f.read(), content_type))
+            return self.upload_data(
+                content=f.read(),
+                filename=basename,
```
yiweig, line 193: style nit: why not rename this var at its declaration on line 189 above? imo having variables change names just because they get passed to another method makes it difficult for readers that are new to the codebase

rllin: good call!

PR #34, opened 2020-08-01
[BACKEND-828] github actions publish package upon release creation

rllin:
1. configured secrets manually via github repo settings
2. this workflow will trigger when a release is created

tested on https://github.com/Labelbox/labelbox-python/runs/934122349?check_suite_focus=true


rllin:
merge because just adding workflow that will not be triggered until release creation

PR #36, opened 2020-08-05
clean up extra bespoke handling of bir queries

rllin:


labelbox/schema/project.py
```diff
@@ -352,6 +352,47 @@ def create_prediction(self, label, data_row, prediction_model=None):
         res = self.client.execute(query_str, params)
         return Prediction(self.client, res["createPrediction"])
 
+    def upload_annotations(
+            self,
+            name: str,
+            url: str = None,
+            path: str = None,
+            annotations: Iterable[dict] = None
```
gszpak, line 360: In general, imho  separate methods are much easier to understand than exclusive args. 
<Java mode on>
How about having an internal AnnotationUploader class with methods `from_url`, `from_path`, `from_object`? Then you'd do `self.annotation_uploader(name).from_url(url)`
<Java mode off>

labelbox/schema/project.py
```diff
@@ -352,6 +352,47 @@ def create_prediction(self, label, data_row, prediction_model=None):
...
+    ) -> None:
+
+        exclusionary_args = iter((url, path, annotations))
+        # any will stop iteration when it finds a single truthy value
+        # not any will continue with where the iteration left off and
+        # check the rest are falsy
+        assert (any(exclusionary_args)
+                and not any(exclusionary_args)), (
+            'Only one of url, path, or annotations can be specified.'
+            )
```
gszpak, line 370: 1. I prefer raising `ValueError` over assertions for invalid args. 
2. `any(exclusionary_args) and not any(exclusionary_args)` <- it'll be hard to make it true:D

labelbox/schema/project.py
```diff
@@ -352,6 +352,47 @@ def create_prediction(self, label, data_row, prediction_model=None):
...
 
+    def upload_annotations(
+            self,
+            name: str,
+            url: str = None,
+            path: str = None,
+            annotations: Iterable[dict] = None
+    ) -> None:
+
+        exclusionary_args = iter((url, path, annotations))
```
gszpak, line 363: Why iter?

labelbox/schema/project.py
```diff
@@ -352,6 +352,47 @@ def create_prediction(self, label, data_row, prediction_model=None):
         res = self.client.execute(query_str, params)
         return Prediction(self.client, res["createPrediction"])
 
+    def upload_annotations(
+            self,
+            name: str,
+            url: str = None,
+            path: str = None,
+            annotations: Iterable[dict] = None
+    ) -> None:
```
gszpak, line 361: `-> BulkImportRequest` - we will have a circular dependency though...

rllin: the way to resolve this with mypy is to just do the string version `'BulkImportRequest'`

labelbox/schema/project.py
rllin: updated

labelbox/schema/project.py
```diff
@@ -352,6 +353,68 @@ def create_prediction(self, label, data_row, prediction_model=None):
         res = self.client.execute(query_str, params)
         return Prediction(self.client, res["createPrediction"])
 
+    def upload_annotations(
+        self,
+        name: str,
+        annotations: Union[str, Iterable[dict]],
```
gszpak, line 359: Let's use pathlib.Path for local file - that's what BIR expects

labelbox/schema/project.py
```diff
@@ -352,6 +353,68 @@ def create_prediction(self, label, data_row, prediction_model=None):
...
+        validate_file: bool = True,
+    ) -> 'BulkImportRequest':
+        """ Uploads annotations to a project.
+
+        Args:
+            name: name of the BulkImportRequest job
+            annotations:
+                url that is publically accessible by Labelbox containing an
```
gszpak, line 367: typo: publicly

labelbox/schema/project.py
```diff
@@ -352,6 +353,68 @@ def create_prediction(self, label, data_row, prediction_model=None):
...
+                """ Verifies that the given string is a valid url.
+
+                Args:
+                    url: string to be checked
+                Returns:
+                    True if the given url is valid otherwise False
+
+                """
+                parsed = urlparse(url)
+                return parsed.http and parsed.netloc
```
gszpak, line 395: Let's move this validation to `BulkImportRequest.create_from_url`

labelbox/schema/project.py
```diff
@@ -352,6 +353,68 @@ def create_prediction(self, label, data_row, prediction_model=None):
         res = self.client.execute(query_str, params)
         return Prediction(self.client, res["createPrediction"])
 
+    def upload_annotations(
+        self,
+        name: str,
+        annotations: Union[str, Iterable[dict]],
+        validate_file: bool = True,
```
gszpak, line 360: Nit: having a flag that's applicable to only one type of arg isn't ideal, can we avoid it?

labelbox/schema/project.py
rllin: keeping as str because this is customer facing and I don't want to be on a call explaining the `Path` lib

i do make sure to cast before passing it to BIR now tho

labelbox/schema/bulk_import_request.py
```diff
@@ -318,3 +203,115 @@ def __build_bulk_import_request_from_result(
...
+        validate_file (bool): a flag indicating if there should be a validation
+            if `file` is a valid ndjson file
+    Returns:
+        BulkImportRequest object
+    """
+    file_name = __make_file_name(project_id, name)
+    content_length = file.stat().st_size
+    request_data = __make_request_data(project_id, name, content_length,
+                                       file_name)
+    if validate_file:
```
rllin, line 301: @gszpak ptal at the changes here

this way we don't read into memory completely for validation

labelbox/schema/bulk_import_request.py
```diff
@@ -213,108 +130,116 @@ def wait_until_done(self, sleep_time_seconds: int = 30) -> None:
...
+        with file.open('rb') as f:
```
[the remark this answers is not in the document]
gszpak: Let's also make sure we fix the docs, i.e. delete the line about loading the whole file to mem

labelbox/schema/bulk_import_request.py
```diff
@@ -25,6 +23,71 @@
...
+
+    if not response_data.get("createBulkImportRequest", None):
+        raise labelbox.exceptions.LabelboxError(
+            "Failed to create BulkImportRequest, message: %s" %
+            response_json.get("errors", None) or
+            response_data.get("error", None))
+
+    return response_data
+
+
```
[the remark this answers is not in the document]
rllin: they didn't need to be `classmethod`s as did not utilize the `cls` itself.

in which case it came down to if they should remain `staticmethod`s rather than module level private functions.

I leaned towards the latter as matter of style

PR #47, opened 2020-08-19
raise error for repeat uuid

rllin:


labelbox/schema/bulk_import_request.py
```diff
@@ -274,3 +277,26 @@ def create_from_local_file(cls,
...
                                                       file_name, file_data)
         return cls(client, response_data["createBulkImportRequest"])
+
+
+class NdjsonError(Exception):
+    pass
+
+
+class UuidError(NdjsonError):
+    pass
```
gszpak, line 287: I'd be careful here - I guess we will add more validation to the SDK, so how about we make a base class more generic? And inherit from LabelboxError?

labelbox/schema/bulk_import_request.py
```diff
@@ -274,3 +277,26 @@ def create_from_local_file(cls,
...
+        - verifies that uuids are unique
+
+    """
+    uuids: Set[str] = set()
+    for line in lines:
+        uuid = line['uuid']
+        if uuid in uuids:
+            raise UuidError(f'{uuid} already used in this import job, '
+                            'must be unique for the project.')
+        uuids.add(line['uuid'])
```
gszpak, line 302: nit: `uuids.add(uuid)`

gszpak (changes requested):

labelbox/schema/bulk_import_request.py
rllin: ah good call, updated

labelbox/schema/bulk_import_request.py
rllin: ah good call, updated

gszpak (approved):
lgtm

PR #57, opened 2020-09-02
adding global logger and unit tests

jtsodapop:


rllin:
reopening through main repo rather than fork

PR #58, opened 2020-09-02
added logger and unit tests

jtsodapop:


labelbox/client.py
```diff
@@ -52,6 +53,11 @@ def __init__(self,
             api_key = os.environ[_LABELBOX_API_KEY]
         self.api_key = api_key
 
+        if verbose:
```
jtsodapop, line 56: verbose is now an arg that will help specify which logger the logger is. lmk if this is the right way to do this!

labelbox/orm/db_object.py
```diff
@@ -7,7 +7,9 @@
 from labelbox.orm.model import Field, Relationship, Entity
 from labelbox.pagination import PaginatedCollection
 
-logger = logging.getLogger(__name__)
+from lb_logger import logger
```
jtsodapop, line 10: changing where logger is coming from

labelbox/schema/bulk_import_request.py
```diff
@@ -22,8 +22,10 @@
 from labelbox.orm.model import Relationship
 from labelbox.schema.enums import BulkImportRequestState
 
+from lb_logger import logger
+
 NDJSON_MIME_TYPE = "application/x-ndjson"
-logger = logging.getLogger(__name__)
+logger = logger
```
jtsodapop, line 28: changing where logger is coming from

labelbox/schema/task.py
```diff
@@ -5,7 +5,9 @@
 from labelbox.orm.db_object import DbObject
 from labelbox.orm.model import Field, Relationship
 
-logger = logging.getLogger(__name__)
+from lb_logger import logger
```
jtsodapop, line 8: changing where logger is coming from

lb_logger.py
```diff
@@ -0,0 +1,4 @@
+import logging
```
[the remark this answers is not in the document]
jtsodapop: added in!

lb_logger.py
```diff
@@ -0,0 +1,4 @@
...
+
+logging.basicConfig(level=logging.INFO)
```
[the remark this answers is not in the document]
jtsodapop: updated in above client.py file

lb_logger.py
```diff
@@ -0,0 +1,4 @@
...
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
jtsodapop: updated in other files which utilize "logger."

test_logger_msg.py
```diff
@@ -0,0 +1,31 @@
+from lb_logger import logger
+import unittest
+
+class TestLoggerMessages(unittest.TestCase):
```
jtsodapop, line 4: docstring added and removed if main...

labelbox/schema/task.py
```diff
@@ -5,7 +5,9 @@
...
+
+logger = logger
```
rllin, line 10: don't need this then

labelbox/client.py
rllin: i'd suggest something like:


```suggestion
        if verbose:
            logger.setLevel(level=logging.INFO)
```

and then where you declare `logger` in `lb_logger` just have it be a `logging.getLogger` call

labelbox/orm/db_object.py
```diff
@@ -7,7 +7,9 @@
...
+
+logger = logger
```
rllin, line 12: don't need this

lb_logger.py
```diff
@@ -0,0 +1,9 @@
+import logging
+"""
+Instantiates a logger object with level set to INFO
+
+This will be used in all other files as the default logger object
+"""
+
+logging.basicConfig(level=logging.INFO)
```
rllin, line 8: don't need this if we go with default level UNLESS client instantiation changes it

test_logger_msg.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_error_msg(self):
+        with self.assertLogs(self.test_logger) as assert_log:
+            self.test_logger.error(self.msg)
+        self.assertEqual(assert_log.output, ['ERROR:lb_logger:%s' %(self.msg)])    
+
+    def test_critical_msg(self):
+        with self.assertLogs(self.test_logger) as assert_log:
+            self.test_logger.critical(self.msg)
+        self.assertEqual(assert_log.output, ['CRITICAL:lb_logger:%s' %(self.msg)])
```
rllin, line 31: i'd add another test that's something like

```
def test_verbosity() -> None:
    assert logger.level == logging.WARNING
    client = Client(verbose=True)
    assert logger.level == logging.INFO
```
    

```

rllin (changes requested):
remove these files

image

labelbox/client.py
jtsodapop: something I noticed is that .setLevel is not sufficient to change the logging output. maybe instead of 
`logger.setLevel(level=logging.INFO`
we can go with
`logging.basicConfig(level=logging.INFO)`

since this originated from `bulk_import_request` not being `level=logging.INFO`, what is the best way to have it so that that file will inherit if verbose == True?

lb_logger.py
jtsodapop: commented on this inside the client.py file conversation

labelbox/client.py
rllin: hmm setLevel should be sufficient as i understand it

and since we're using the same logger throughout, BIR should end up using the logger after the client has changed the level

jtsodapop: I see now, thanks for explaining BIR for me

setLevel unfortunately does not looks like it works. can provide an example file which indicates this

jtsodapop: hmm. will relook at this. test seems to show it works correctly but an external example test shows it does not

PR #60, opened 2020-09-03
Logger

jtsodapop:


labelbox/client.py
```diff
@@ -52,7 +53,10 @@ def __init__(self,
             api_key = os.environ[_LABELBOX_API_KEY]
         self.api_key = api_key
 
-        logger.info("Initializing Labelbox client at '%s'", endpoint)
+        if verbose:
+            logging.basicConfig(level=logging.INFO)
```
jtsodapop, line 57: found some interesting stuff on this one! so `logger.setLevel` will update the LOGGER's level, but because the logger is independent of the HANDLER, even by updating the logger itself, we will still not get any output.

default handler is set to warning and this occurs when we call absolutely no handler (handler of last resort, which has level set to warning)

`logging.basicConfig` is actually called whenever we do not have any set handler, so it defaults to warning. by updating this to logging.INFO, we are updating the default handler's level to INFO instead of WARNING

i suppose if we did not want to update the actual default handler's level, we can always go back to the idea of adding in a handler ourselves

jtsodapop: additional information provided [here](https://stackoverflow.com/questions/43109355/logging-setlevel-is-being-ignored) and here and here

test_logger_msg.py
```diff
@@ -0,0 +1,39 @@
...
+        with self.assertLogs(self.test_logger) as assert_log:
+            self.test_logger.error(self.msg)
+        self.assertEqual(assert_log.output, ['ERROR:lb_logger:%s' %(self.msg)])    
+
+    def test_critical_msg(self) -> None:      
+        with self.assertLogs(self.test_logger) as assert_log:
+            self.test_logger.critical(self.msg)
+        self.assertEqual(assert_log.output, ['CRITICAL:lb_logger:%s' %(self.msg)])     
+
+    def test_verbosity(self) -> None:
```
jtsodapop, line 36: i am a bit unsure on how to confirm this test. i had to insert my own `api_key` variable here to confirm the test. is there a good way to ensure that a call to Client does not need an actual api key when running with pytest?

lb_logger.py
```diff
@@ -0,0 +1,9 @@
+import logging
+"""
+Instantiates a logger object with level set to INFO
+
+This will be used in all other files as the default logger object
+"""
+
+# logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
```
jtsodapop, line 9: it is beginning to look like all of the files that are being redirected to lb_logger file is only a one liner change. if we are updating the basicConfig inside Client.py, is there still a need for a centralized logger object?

PR #63, opened 2020-09-08
adding in the test_logger file and updating the readme to include ins…

jtsodapop:
…talling Docker


tests/integration/test_logger.py
```diff
@@ -0,0 +1,12 @@
+from labelbox import Client
+import pytest
+import logging
+
+def test_client_log(caplog, project):
```
rllin, line 5: docstring

tests/integration/test_logger.py
```diff
@@ -0,0 +1,12 @@
+from labelbox import Client
```
rllin, line 1: docstring

rllin (approved):

tests/integration/test_logger.py
jtsodapop: added !

PR #70, opened 2020-09-24
uploading test.py file to showcase an issue with where clause in clie…

jtsodapop:
…nt.get_datasets, also adding in prints to query.py for visibility. no plans to merge, just for showcasing issue


jtsodapop:
killing PR

PR #82, opened 2020-11-06
Adding in function to toggle on and off model assisted labeling

jtsodapop:


labelbox/schema/project.py
```diff
@@ -367,6 +367,34 @@ def create_prediction(self, label, data_row, prediction_model=None):
         res = self.client.execute(query_str, params)
         return Prediction(self.client, res["createPrediction"])
 
+    def enable_model_assisted_labeling(self, toggle:bool=True) -> bool:
```
rllin, line 370:
```suggestion
    def enable_model_assisted_labeling(self, toggle: bool=True) -> bool:
```

labelbox/schema/project.py
```diff
@@ -367,6 +367,34 @@ def create_prediction(self, label, data_row, prediction_model=None):
...
+        """ Turns model assisted labeling either on or off based on input
+
+        Args:
+            toggle (Boolean): True or False boolean 
+        Returns:
+            True or False depending on the condition given
```
rllin, line 376: would be more correct and understandable if you said here instead that it returns True if turned on and False if turned off

labelbox/schema/project.py
```diff
@@ -367,6 +367,34 @@ def create_prediction(self, label, data_row, prediction_model=None):
...
+        """
+
+        project_param = "project_id"
```
rllin, line 379: i think these can just go into `params` rather than being pulled out up here

rllin (approved):

PR #91, opened 2021-01-12
Modify files for sphinx auto docs

alexandracota:
This PR excludes the sphinx documentation files.


labelbox/client.py
```diff
@@ -29,25 +29,24 @@ class Client:
...
-            >>> client = Client("<APIKEY>")
+        >>> import logger
+        >>> logging.basicConfig(level = logging.INFO)
+        >>> client = Client("<APIKEY>")
 
         Args:
-            api_key (str): API key. If None, the key is obtained from
-                the "LABELBOX_API_KEY" environment variable.
-            endpoint (str): URL of the Labelbox server to connect to.
+            api_key (String): API key. If None, the key is obtained from the "LABELBOX_API_KEY" environment variable.
```
florijanstamenkovic, line 48: You have been replacing "str" with `String` in most (all?) places... I don't think that's correct, generally speaking, as `str` is the proper name of a Python class representing a sequence of characters.

labelbox/client.py
```diff
@@ -434,16 +437,15 @@ def _create(self, db_object_type, data):
...
-            >>> dataset = client.create_dataset(name="<dataset_name>", projects=project)
+        >>> project = client.get_project("<project_uid>")
+        >>> dataset = client.create_dataset(name="<dataset_name>", projects=project)
 
-        Kwargs:
-            Keyword arguments with new Dataset attribute values.
-            Keys are attribute names (in Python, snake-case convention) and
-            values are desired attribute values.
+        Args:
+            kwargs: Keyword arguments with Dataset attribute values.
```
florijanstamenkovic, line 448: Are you sure this generates the correct thing? `kwargs` is not like other, normal args... It should be documented differently, but I'm not exactly sure which convetion you/Sphynx prefer.

Some quick googling gives: https://stackoverflow.com/questions/1137161/what-is-the-correct-way-to-document-a-kwargs-parameter

labelbox/client.py
```diff
@@ -453,15 +455,14 @@ def create_dataset(self, **kwargs):
...
 
-            >>> project = client.create_project(name="<project_name>", description="<project_description>")
+        >>> project = client.create_project(name="<project_name>", description="<project_description>")
 
-        Kwargs:
-            Keyword arguments with new Project attribute values.
-            Keys are attribute names (in Python, snake-case convention) and
-            values are desired attribute values.
+        Args:
+            kwargs: Keyword arguments with new Project attribute values.
```
florijanstamenkovic, line 465: Ditto.. This is important to get right.

labelbox/schema/benchmark.py
```diff
@@ -3,12 +3,21 @@
...
-    consistency of the training data. Benchmarks works by interspersing data
+    """ Represents a benchmark label.
+    
+    The Benchmarks tool works by interspersing data
     to be labeled, for which there is a benchmark label, to each person labeling.
     These labeled data are compared against their respective benchmark and an
     accuracy score between 0 and 100 percent is calculated.
+
+    Attributes:
+        created_at (DateTime)
```
florijanstamenkovic, line 14: Another issue with class names. Like with `str`, these attributes should probably be declared as python classes they belong to, namely `datetime`, `int` and `float`.

I know it's confusing as I'm declaring class variables like for example `Field.DateTime`, but that's a different thing.

.gitignore
```diff
@@ -131,3 +131,6 @@ dmypy.json
 
 # macos files
 .DS_STORE
+
+# Docs build
+labelbox/build/
```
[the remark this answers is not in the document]
alexandracota: Added newline

labelbox/client.py
```diff
@@ -29,25 +29,24 @@ class Client:
     """ A Labelbox client. Contains info necessary for connecting to
     a Labelbox server (URL, authentication key). Provides functions for
     querying and creating top-level data objects (Projects, Datasets).
+
```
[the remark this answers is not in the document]
alexandracota: Removed unnecessary newlines.

labelbox/client.py
alexandracota: Reverted all cases of String to back to str.

labelbox/client.py
alexandracota: Fixed the formatting for kwargs. Decided to go with [this format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) so we can leave kwargs unexpanded. Sphinx recognized Args as an alias of Parameters, [see docs](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html#sections). Decided to keep Parameters throughout for consistency.

labelbox/client.py
alexandracota: See comment above.

labelbox/schema/benchmark.py
alexandracota: Reverted back to str, datetime, int, and float.

labelbox/schema/project.py
```diff
@@ -403,15 +423,14 @@ def upload_annotations(
         """ Uploads annotations to a new Editor project.
 
         Args:
-            name: name of the BulkImportRequest job
-            annotations:
+            name (str): name of the BulkImportRequest job
+            annotations (str or dict):
```
florijanstamenkovic, line 427: This should be "str or Path or Iterable"

labelbox/client.py
florijanstamenkovic: Agreed regarding `kwargs`.

I don't understand what you mean by "decided to keep Parameters throughout for consistency"... We're not using "Parameters" in the documentation, but "Args"?

alexandracota: "Args" and "Arguments" are aliases for "Parameters". So, rather than replacing all instances of "Args" in our current documentation, I just kept "Args" as is and indicate all instances of kwargs with a double asterisk. When Sphinx generates the docs, all instances of "Args" become "Parameters". So, it will look like this:
Screen Shot 2021-01-14 at 8 47 53 AM

florijanstamenkovic: OK, got it.

PR #92, opened 2021-01-14
Fix build GitHub action

florijanstamenkovic:
* Fix GitHub action env var handling
* Fix yapf compliance

Tests still fail. This should be addressed, but is a larger issue unrelated to the fixes in this PR.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -27,9 +27,9 @@ jobs:
     - name: set environment for branch
       run: |
           if [[ "${{github.base_ref}}" == "master" || "${{github.ref}}" == "refs/heads/master" ]]; then
-            echo "::set-env name=LABELBOX_TEST_ENVIRON::prod"
+            echo "LABELBOX_TEST_ENVIRON=prod" >> $GITHUB_ENV
           else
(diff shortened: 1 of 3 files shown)


.github/workflows/python-package.yml
```diff
@@ -27,9 +27,9 @@ jobs:
     - name: set environment for branch
       run: |
           if [[ "${{github.base_ref}}" == "master" || "${{github.ref}}" == "refs/heads/master" ]]; then
-            echo "::set-env name=LABELBOX_TEST_ENVIRON::prod"
```
gszpak, line 30: seems like `set-env` is deprecated? https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

gszpak (changes requested):

.github/workflows/python-package.yml
florijanstamenkovic: Yes, this is the point of this PR, removing `set-env` in favor of concatenating to an env-file, as linked on the page you're referencing: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

gszpak (approved):

PR #96, opened 2021-01-19
Add .readthedocs.yml

alexandracota:
The .readthedocs.yml file is necessary for Readthedocs platform to build and display the sphinx-generated docs. Also added autodoc to requirements.txt.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,3 @@
 Sphinx==3.4.3
 sphinxcontrib-napoleon==0.7
 sphinx-rtd-theme==0.5.1
-autodoc==0.5.0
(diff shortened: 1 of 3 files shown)


docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
 Sphinx==3.4.3
 sphinxcontrib-napoleon==0.7
 sphinx-rtd-theme==0.5.1
+autodoc==0.5.0
```
florijanstamenkovic, line 4: Are you sure this is necessary?

I'm finding no mention of it in https://docs.readthedocs.io/en/stable/builds.html.... Also, the Autodoc project (https://pypi.org/project/autodoc/) seems to talk about documenting unit tests, which I'm pretty sure we don't want to do.

florijanstamenkovic (changes requested):

docs/requirements.txt
alexandracota: Yup, you are right. I removed the autodoc dependency.

PR #98, opened 2021-01-20
Fix minor bugs and tidy tests up

florijanstamenkovic:
* Fix client error test
* Improve `DbObject` equality testing
* Make to-one relationship fetching more robust
* Fix `Project.labels` ordering
* Mark tests as either `xfail` (for server-side errors we can't control in SDK) or `skip` (for things we should look into in the SDK)
* Skip slow tests when running `tox`
* Delete `test_logger` as it tests Pytest and not Labelbox


.github/workflows/python-package.yml
```diff
@@ -73,4 +73,4 @@ jobs:
         # randall+staging-python@labelbox.com
         LABELBOX_TEST_API_KEY_STAGING: ${{ secrets.STAGING_LABELBOX_API_KEY }}
       run: |
-        pytest -svv
+        tox -e py
```
gszpak, line 76: nit: could we pass pytest flags here too? `tox -e py -- -svv`?

labelbox/client.py
```diff
@@ -186,11 +186,27 @@ def check_errors(keywords, *path):
...
+
+        # A lot of different error situations are now labeled serverside
+        # as INTERNAL_SERVER_ERROR, when they are actually client errors.
+        # TODO: fix this in the server API
+        internal_server_error = check_errors(["INTERNAL_SERVER_ERROR"],
+                                             "extensions", "code")
+        if internal_server_error is not None:
+            message = internal_server_error.get("message")
+
+            if message.startswith("Syntax Error"):
```
gszpak, line 205: So the API is the reason why we check it this way right?

tests/integration/test_webhook.py
```diff
@@ -1,6 +1,10 @@
+import pytest
+
 from labelbox import Webhook
 
 
+# TODO investigate why this fails
```
gszpak, line 6: nit: xfail instead of skip?

tox.ini
```diff
@@ -8,4 +8,4 @@ deps =
     -rrequirements.txt
     pytest
 passenv = LABELBOX_TEST_API_KEY_PROD LABELBOX_TEST_API_KEY_STAGING LABELBOX_TEST_ENVIRON
-commands = pytest {posargs}
+commands = pytest -m "not slow" {posargs}
```
gszpak, line 11: we never run slow tests?

gszpak (approved):

tox.ini
florijanstamenkovic: It's a default in tox, if you just run `pytest` locally it will run them... So, not really sure when to run and when not :/ Wdyt?

tests/integration/test_webhook.py
florijanstamenkovic: I invented a convention of using `xfail` for situations where the server-side is causing the failure, but the PySDK is actually OK, and using `skip` for things we temporarily skip but should resolve (soon). And I added them as tasks. So, skips should be figured out and handled sometime soon.

labelbox/client.py
florijanstamenkovic: Yep. For syntax errors the API now responds with "INTERNAL_SEVER_ERROR", but the message contains "Syntax Error". This is wrong, but yeah...

tox.ini
gszpak: Is there a way to trigger build when the PR is approved? Then we could run all tests, including the slow ones?

florijanstamenkovic: Well, there is: https://github.community/t/feature-request-trigger-action-on-pull-request-approved/18413/2

But the logic should then be: if (push and was_approved_before) or (is_approved), which contains different GH action events and I don't want to get into that right now. Will first just remove the "not slow" flag and have everything run.

gszpak (approved):

PR #100, opened 2021-02-03
added sdk version to headers

msokoloff1:


setup.py
```diff
@@ -1,11 +1,14 @@
 import setuptools
+from labelbox._version import __version__
```
gszpak, line 2: @nmaswood a broader explanation of @msokoloff1 's answer: https://stackoverflow.com/a/24435742.

As I agree the import will work, it feels a bit tricky to me and can lead to problems - ptal at warning at point 6.: https://packaging.python.org/guides/single-sourcing-package-version/

I would lean towards the solution that @nmaswood suggested (or a simpler version without regexes, eg https://github.com/scikit-image/scikit-image/blob/dcec6c4c901072a813b0cd68e2589593a1e37cc9/setup.py#L123)

But I agree all of these solutions are kinda ugly.

gszpak (approved):

PR #112, opened 2021-03-04
MAL Validation

msokoloff1:
* Client side validation for MAL


tests/integration/test_bulk_import_request.py
```diff
@@ -1,154 +1,141 @@
 import uuid
+from attr import validate
```
florijanstamenkovic, line 2: It seems that `validate` isn't used in this file? Is this from the `attrs` library? It's not stated in the dependencies... Kinda confused what's going on here.

tests/integration/conftest.py
```diff
@@ -1,11 +1,12 @@
 from collections import namedtuple
 from enum import Enum
 from datetime import datetime
+from labelbox.schema.labeling_frontend import LabelingFrontend
```
florijanstamenkovic, line 4: Bad import order. Let's follow https://www.python.org/dev/peps/pep-0008/#imports. We also use a few other conventions (mostly driven by Greg's PyCharm) in other places, that we should sync on and decide what we want to propagate in the SDK.

tests/integration/conftest.py
```diff
@@ -123,3 +124,282 @@ def sample_video() -> str:
...
+    tools = [
+        bbox_tool, polygon_tool, polyline_tool, point_tool, entity_tool,
+        segmentation_tool
+    ]
+    classifications = [checklist, free_form_text]
+    return {"tools": tools, "classifications": classifications}
+
+
+@pytest.fixture
+def configured_project(client, rand_gen, ontology):
```
florijanstamenkovic, line 228: 1. I think `project_with_setup` is a better name.
2. I think this can reuse existing fixtures to be
```
@pytest.fixture
def project_with_setup(client, rand_gen, ontology, project, dataset):
    editor = list(client.get_labeling_frontends(
        where=LabelingFrontend.name == "editor"))[0]
    project.setup(editor, ontology)
    for _ in range(len(ontology['tools']) + len(ontology['classifications'])):
        dataset.create_data_row(row_data=IMG_URL)
    project.datasets.connect(dataset)
    yield project
```

tests/integration/conftest.py
```diff
@@ -123,3 +124,282 @@ def sample_video() -> str:
     path_to_video = 'tests/integration/media/cat.mp4'
     assert os.path.exists(path_to_video)
     return path_to_video
+
```
florijanstamenkovic, line 127: All of this hugely bloats `conftest.py`. I'd consider adding a subdirectory `tests/integration/bulk_import/`, and having a dedicated `conftest.py` there, along with the test files for that functionality. Not sure :/

tests/integration/test_ndjon_validation.py
```diff
@@ -0,0 +1,226 @@
+from labelbox.exceptions import NDJsonError, UuidError
```
florijanstamenkovic, line 1: Bad file name.

florijanstamenkovic: Also bad import order, please check across all files you edit.

tests/integration/test_ndjon_validation.py
```diff
@@ -0,0 +1,226 @@
...
+
+def test_subclassification_construction(rectangle_inference):
+    tool = NDTool.build(rectangle_inference)
+    assert len(tool.classifications) == 1, "Subclass was not constructed"
+    assert isinstance(tool.classifications[0], NDRadio)
+
+
+def test_tool_construction(polygon_inference, rectangle_inference,
+                           line_inference, entity_inference,
+                           segmentation_inference):
```
florijanstamenkovic, line 26: I'd parametrize:
```
@pytest.mark.parametrize("inference, expected_type", [
    (polygon_inference, NDPolygon), ...
])
def test_tool_construction(inference, expected_type):
    assert isinstance(NDTool.build(inference), expected_type)
```

Easier to parse for humans.

tests/integration/conftest.py
msokoloff1: I'll reuse the dataset fixture but not the project one. A test could the project to not yet be setup.

labelbox/exceptions.py
```diff
@@ -104,3 +104,8 @@ class MalformedQueryException(Exception):
 class UuidError(LabelboxError):
     """ Raised when there are repeat Uuid's in bulk import request."""
     pass
+
+
+class NDJsonError(LabelboxError):
```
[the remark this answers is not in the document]
msokoloff1: It's not a json error. This exception will be raised if we cannot construct a valid mal example from the line of json - even if the line is valid json.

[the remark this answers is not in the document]
florijanstamenkovic: Ah, OK. Then maybe we should call it `ValidationError` or similar? My understanding, when seeing this name, would be that the JSON part is what's causing the error...

tests/integration/conftest.py
florijanstamenkovic: The default fixture scope is "function". This means the fixture is invoked (objects created and destroyed) for each test function. So if you reuse the `project` fixture in `project_with_setup`, that project will be created and destroyed for each test, without the setup contaminating between tests.

labelbox/exceptions.py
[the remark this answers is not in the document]
msokoloff1: ValidationError sounds good

tests/integration/conftest.py
msokoloff1: You are right. I'll make the change

setup.py
```diff
@@ -22,10 +22,8 @@
     install_requires=[
         "backoff==1.10.0",
         "backports-datetime-fromisoformat==1.0.0; python_version < '3.7.0'",
-        "dataclasses==0.7; python_version < '3.7.0'",
-        "ndjson==0.3.1",
-        "requests>=2.22.0",
-        "google-api-core>=1.22.1",
+        "dataclasses==0.7; python_version < '3.7.0'", "ndjson==0.3.1",
```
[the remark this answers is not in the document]
msokoloff1: The code formatter does this. I agree we shouldn't have our requirements stored in two places.

tests/integration/bulk_import/conftest.py
```diff
@@ -0,0 +1,276 @@
...
+        'options': [{
+            'label': 'option1',
+            'value': 'option1'
+        }, {
+            'label': 'option2',
+            'value': 'option2'
+        }, {
+            'label': 'optionN',
+            'value': 'optionn'
+        }]
```
florijanstamenkovic, line 84: I'd say this formatting is so-so... Values can be inlined here most of the time, resulting in twice as less rows and improved readability. I guess it's a result of automatic formatting, and I'm against it.

tests/integration/bulk_import/conftest.py
```diff
@@ -0,0 +1,276 @@
...
+    }
+    free_form_text = {
+        'required': False,
+        'instructions': 'text',
+        'name': 'text',
+        'type': 'text',
+        'options': []
+    }
```
florijanstamenkovic, line 92: Hm, here we have better formatting.

tests/integration/bulk_import/test_bulk_import_request.py
```diff
@@ -0,0 +1,139 @@
...
+                                                                annotations=url,
+                                                                validate=False)
+
+    bulk_import_request.wait_until_done()
+
+    assert (bulk_import_request.state == BulkImportRequestState.FINISHED or
+            bulk_import_request.state == BulkImportRequestState.FAILED)
+
+
+def __assert_file_content(url: str, predictions):
```
florijanstamenkovic, line 137: Why the double underscore?

florijanstamenkovic (commented):
Still some things to fix up, please address.

tests/integration/bulk_import/test_bulk_import_request.py
msokoloff1: I didn't add this code. I just moved it to the new directory.

msokoloff1: I'm happy to change it if you want me to.

labelbox/schema/project.py
```diff
@@ -464,15 +465,15 @@ def _is_url_valid(url: Union[str, Path]) -> bool:
...
+                    validate_file=validate,
                 )
         elif isinstance(annotations, Iterable):
             return BulkImportRequest.create_from_objects(
                 client=self.client,
                 project_id=self.uid,
                 name=name,
                 predictions=annotations,  # type: ignore
-            )
+                validate=validate)
```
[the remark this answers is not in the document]
msokoloff1: I just resolved this as I was making the changes. Not sure why this wasn't saved.

Dockerfile
```diff
@@ -1,8 +1,12 @@
 FROM python:3.7
 
-RUN pip install pytest
+RUN pip install pytest pytest-cases
```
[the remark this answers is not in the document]
msokoloff1: I think it is fine. It does improve readability and might be useful in the future.

labelbox/schema/project.py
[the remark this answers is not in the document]
msokoloff1: Sorry I removed it earlier and then it came back :). We can't remove this check unless we also check if the first example of the iterable is a dict. Or we change the type hint to Iterable[Any]. So I think that the mypy ignore is probably the best option.

setup.py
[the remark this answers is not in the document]
florijanstamenkovic: Ah yes, YAPF is a build requirement for this repo, forgot about that, sorry... Not sure that's a good idea as it sometimes enforces a mess. But, back to the point. I guess we'll need to configure YAPF to ignore this file (https://github.com/google/yapf#excluding-files-from-formatting-yapfignore), and we might need to configure it a bit more (for the case below), but like you say, we can leave that out of this PR. I forgot about it being mandatory.

Dockerfile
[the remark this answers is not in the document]
florijanstamenkovic: OK.

tests/integration/bulk_import/conftest.py
florijanstamenkovic: Yes, I hope we can. Personally I hate this kind of unreadable mess. But, like I mentioned, I forgot it was mandatory for the build to pass. Let's consider outside of this PR.

tests/integration/bulk_import/test_bulk_import_request.py
florijanstamenkovic: Please do! I always try to do these kind of minor improvements in the code I touch, in time it accumulates.

florijanstamenkovic (approved):

gszpak (approved):

labelbox/schema/bulk_import_request.py
```diff
@@ -296,17 +313,437 @@ def create_from_local_file(cls,
...
+            uids.add(uuid)
+        except (ValidationError, ValueError, TypeError, KeyError) as e:
+            raise labelbox.exceptions.ValidationError(
+                f"Invalid NDJson on line {idx}") from e
+
+
+#The rest of this file contains objects for MAL validation
+def parse_classification(tool):
+    """
+    Parses a classification from an ontology. Only radio, checklist, and text are supported for mal
```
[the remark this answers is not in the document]
msokoloff1: So are we able to support dropdowns?

labelbox/schema/bulk_import_request.py
```diff
@@ -296,17 +313,437 @@ def create_from_local_file(cls,
...
+    for tool in ontology.normalized['classifications']:
+        valid_feature_schemas[tool['featureSchemaId']] = parse_classification(
+            tool)
+    return valid_feature_schemas
+
+
+LabelboxID: str = pydantic.Field(..., min_length=25, max_length=25)
+
+
+class Bbox(TypedDict):
```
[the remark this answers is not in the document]
msokoloff1: It just lets you specify the structure of the dictionary.

labelbox/schema/bulk_import_request.py
```diff
@@ -296,17 +313,437 @@ def create_from_local_file(cls,
...
+    id: str
+
+
+class NDFeatureSchema(BaseModel):
+    schemaId: str = LabelboxID
+
+
+class NDBase(NDFeatureSchema):
+    ontology_type: str
+    schemaId: str = LabelboxID
```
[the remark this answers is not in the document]
msokoloff1: yes that was an accident

labelbox/schema/bulk_import_request.py
```diff
@@ -296,17 +313,437 @@ def create_from_local_file(cls,
...
+    end: int
+    start: int
+
+
+class VideoSupported(BaseModel):
+    #Note that frames are only allowed as top level inferences for video
+    frames: Optional[List[FrameLocation]]
+
+
+class UnionConstructor:
```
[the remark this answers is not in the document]
msokoloff1: I'll add some comments to make it more clear in the code. Basically unions don't produce friendly error messages because there is no way to know which of the classes the user intended to construct. What this does is check for specific keys to know what the user intended. If none of those special keys exist then there is another error message alerting the user that they are missing the identifying keys.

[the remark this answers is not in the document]
msokoloff1: This object also has some weird functions to allow pydantic to make use of it.

labelbox/exceptions.py
```diff
@@ -104,3 +104,8 @@ class MalformedQueryException(Exception):
 class UuidError(LabelboxError):
     """ Raised when there are repeat Uuid's in bulk import request."""
     pass
+
+
+class ValidationError(LabelboxError):
+    """Raised when user input is invalid for MAL imports."""
+    ...
```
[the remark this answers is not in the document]
msokoloff1: That is kind of long. How is MALValidationError?

PR #124, opened 2021-03-19
catches errors properly

msokoloff1:
Error messages coming from the api were slightly changed. This update properly parses the new format.


nmaswood (approved):
lgtm

although do you mind including a short write up in the description

PR #127, opened 2021-03-25
fixed bugs, ready for release

msokoloff1:
* Some extra information from the mal validation process was being added to the upload. This caused mal for subclasses to fail. 
* Bumped the version and updated the changelog so we are ready to release a fix.
* Updated the README to include that the SDK works for python3.6 
* Fixed an incorrect reference to a notebook in the examples README.

### DIFF
--- a/labelbox/schema/bulk_import_request.py
+++ b/labelbox/schema/bulk_import_request.py
@@ -343,7 +343,7 @@ def _validate_ndjson(lines: Iterable[Dict[str, Any]],
     uids: Set[str] = set()
     for idx, line in enumerate(lines):
         try:
-            annotation = NDAnnotation(**line)
+            annotation = NDAnnotation(**line.copy())
             annotation.validate_instance(data_row_ids, feature_schemas)
             uuid = str(annotation.uuid)
             if uuid in uids:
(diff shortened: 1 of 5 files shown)


labelbox/schema/bulk_import_request.py
```diff
@@ -633,13 +633,12 @@ def validate_subclasses(cls, value, field):
         #This is caused by the fact that we require these ids for top level classifications but not for subclasses
         results = []
         for row in value:
-            row.update({
-                'dataRow': {
+            results.append({
+                **row, 'dataRow': {
                     'id': 'child'.center(25, '_')
```
nmaswood, line 638: I would maybe move `'child`.center(25, '_')` out of the loop, it makes it more clear that it's just a dummy fill in value.

nmaswood (approved):

labelbox/schema/bulk_import_request.py
msokoloff1: Good call!

nmaswood (approved):
if you like you could even put at module scope

```python
DUMMY_ID  =...
```

labelbox/schema/bulk_import_request.py
msokoloff1: Fixed.

PR #134, opened 2021-04-02
Add Export Issues to SDK

Rodion13:
https://labelbox.atlassian.net/browse/PLAT-1233

Blocked on: 
https://github.com/Labelbox/intelligence/pull/4063

All Issues
```
from labelbox import Client
client = Client(api_key='SOME KEY', endpoint='http://localhost:8080/graphql')
p = client.get_project('ckml5cezy009h0871u12kogd6')
p.export_issues()

Output: 'http://localhost:8080/v1/project/ckml5cezy009h0871u12kogd6/issues?bewit=bGFiZWxib3gtYXBpXDE2MTczOTg4MzdceWt4UWswUENBL0JPeXhCNVh4QldLdncvR3NyNVpqYm11YjlDU1kyZHZkTT1c'
```

Open Issues
```
client.get_project('ckml5cezy009h0871u12kogd6')
export_issues("Open")

Output: 'http://localhost:8080/v1/project/ckml5cezy009h0871u12kogd6/issues?status=Open&bewit=bGFiZWxib3gtYXBpXDE2MTczOTg4NDVcbUhLY0ZsU0tzKzMvWXZxT0hMQ3dtL3IyUzZvMHpSMmZyQ3JIRXNjT3VBdz1c'
```

Resolved Issues
```
p.export_issues("Resolved")

Output: 'http://localhost:8080/v1/project/ckml5cezy009h0871u12kogd6/issues?status=Resolved&bewit=bGFiZWxib3gtYXBpXDE2MTczOTkwNTdcRGtmNERlWWZDNVZCNzExK0plWGNhZ0R3czRWUmxyMjBRM1BvQjhKMU5Ccz1c'
```

Unknown Status:
```
p.export_issues("Resolve")



ValueError                                Traceback (most recent call last)
<ipython-input-4-5038dd8cc313> in <module>
      2 client = Client(api_key='eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VySWQiOiJja20zNnhqeDEwMG9pMDkwMXFxaXNtZ2tpIiwib3JnYW5pemF0aW9uSWQiOiJja20zNnhqdXEwMG9jMDkwMTVobzQ2NXQyIiwiYXBpS2V5SWQiOiJja216Y25tZm8wMnpqMDgzNmdyOXB3aGt2IiwiaWF0IjoxNjE3MzEwMDY3LCJleHAiOjIyNDg0NjIwNjd9.U2pug93HJedlrKUYhOAwZjmJWvpkLxvgNFLWW_D4K-w', endpoint='http://localhost:8080/graphql')
      3 p = client.get_project('ckml5cezy009h0871u12kogd6')
----> 4 p.export_issues("Resolve")

~/workspace/labelbox-python/labelbox/schema/project.py in export_issues(self, status, timeout_seconds)
    201
    202         if status not in valid_statuses:
--> 203             raise ValueError("status must be in {}. Found {}".format(
    204                 valid_statuses, status))
    205

ValueError: status must be in [None, 'Open', 'Resolved']. Found Resolve
```

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -177,6 +177,48 @@ class Project(DbObject, Updateable, Deletable):
                          self.uid)
             time.sleep(sleep_time)
 
+    
+    def export_issues(self, status=None, timeout_seconds=60):
+        """ Calls the server-side Issues exporting that generates a JSON
+        payload, and returns the URL to that payload.
+
+        Args:
+            timeout_seconds (float): Max waiting time, in seconds.
+            status (string): valid values: Open, Resolved
+        Returns:
+            URL of the data file with this Project's issues. If the server didn't
+            generate during the `timeout_seconds` period, None is returned.
+        """
+        sleep_time = 2
+        id_param = "projectId"
+        status_param = "status"
+        query_str = """query getProjectIssues($%s: ID!, $%s: IssueStatus) {
+            project(where: { id: $%s }) {
+                issueExportUrl(where: { status: $%s })
+            }
+        }""" % (id_param, status_param, id_param, status_param)
+
+        valid_statuses = [None, "Open", "Resolved"]
+
+        if status not in valid_statuses:
+            raise ValueError("status must be in {}. Found {}".format(valid_statuses, status))
+
+        while True:
+            res = self.client.execute(query_str, {id_param: self.uid, status_param: status})
+            print(res)
+            res = res.get('project').get('issueExportUrl')
+
+            if res:
+                return res
+
+            timeout_seconds -= sleep_time
+            if timeout_seconds <= 0:
+                return None
+
+            logger.debug("Project '%s' issues export, waiting for server...",
+                         self.uid)
+            time.sleep(sleep_time)
+
     def upsert_instructions(self, instructions_file: str):
         """
         * Uploads instructions to the UI. Running more than once will replace the instructions
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -96,8 +96,8 @@ class Project(DbObject, Updateable, Deletable):
         Label = Entity.Label
 
         kwargs[Label.project] = self
-        kwargs[Label.seconds_to_label] = kwargs.get(
-            Label.seconds_to_label.name, 0.0)
+        kwargs[Label.seconds_to_label] = kwargs.get(Label.seconds_to_label.name,
+                                                    0.0)
         data = {
             Label.attribute(attr) if isinstance(attr, str) else attr:
             value.uid if isinstance(value, DbObject) else value
@@ -142,9 +142,8 @@ class Project(DbObject, Updateable, Deletable):
             id_param, id_param, where, order_by_str,
             query.results_query_part(Label))
 
-        return PaginatedCollection(self.client, query_str,
-                                   {id_param: self.uid}, ["project", "labels"],
-                                   Label)
+        return PaginatedCollection(self.client, query_str, {id_param: self.uid},
+                                   ["project", "labels"], Label)
 
     def export_labels(self, timeout_seconds=60):
         """ Calls the server-side Label exporting that generates a JSON
@@ -305,13 +304,11 @@ class Project(DbObject, Updateable, Deletable):
             # python isoformat doesn't accept Z as utc timezone
             result["lastActivityTime"] = datetime.fromisoformat(
                 result["lastActivityTime"].replace('Z', '+00:00'))
-            return LabelerPerformance(**{
-                utils.snake_case(key): value
-                for key, value in result.items()
-            })
+            return LabelerPerformance(
+                **
+                {utils.snake_case(key): value for key, value in result.items()})
 
-        return PaginatedCollection(self.client, query_str,
-                                   {id_param: self.uid},
+        return PaginatedCollection(self.client, query_str, {id_param: self.uid},
                                    ["project", "labelerPerformance"],
                                    create_labeler_performance)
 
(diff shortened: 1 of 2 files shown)


labelbox/schema/project.py
```diff
@@ -177,6 +177,51 @@ def export_labels(self, timeout_seconds=60):
...
+        if status not in valid_statuses:
+            raise ValueError("status must be in {}. Found {}".format(
+                valid_statuses, status))
+
+        while True:
+            res = self.client.execute(query_str, {
+                id_param: self.uid,
+                status_param: status
+            })
+            print(res)
```
msokoloff1, line 211: This needs to be removed.

labelbox/schema/project.py
```diff
@@ -177,6 +177,51 @@ def export_labels(self, timeout_seconds=60):
...
+            res = res.get('project').get('issueExportUrl')
```
msokoloff1, line 211: I think there is potential for this to quietly break in the future. If the issueExportKey is missing for any reason this will silently fail. Is there a more specific way to see if we should retry? Also why not use res['project'] instead of res['project'] so that it throws a KeyError instead of an AttributeError?

msokoloff1 (changes requested):

msokoloff1:
Do you mind adding a test case for this?

labelbox/schema/project.py
Rodion13: I would retry if res.get('issueExportUrl') return None (key is missing) or say '' string is returned for URL

msokoloff1: I am just wondering if this can be specific about when to retry. Wouldn't there only be one condition to retry on? An empty string or a missing key? It doesn't make sense for both of those conditions. I am worried that checking for a missing key to retry could cause problems. If the key always exists and the retry depends on whether the string is empty or not, then you should change this to explicitly check for an empty string.

labelbox/schema/project.py
```diff
@@ -177,6 +177,50 @@ def export_labels(self, timeout_seconds=60):
...
+            timeout_seconds (float): Max waiting time, in seconds.
+            status (string): valid values: Open, Resolved
+        Returns:
+            URL of the data file with this Project's issues. If the server didn't
+            generate during the `timeout_seconds` period, None is returned.
+        """
+        sleep_time = 2
+        id_param = "projectId"
+        status_param = "status"
+        query_str = """query getProjectIssues($%s: ID!, $%s: IssueStatus) {
```
msokoloff1, line 194: Please change this to `getProjectIssuesPyApi` this naming convention is used on the backend to record metrics.

labelbox/schema/project.py
msokoloff1: E.g. the[project.export_labels](https://github.com/Labelbox/labelbox-python/blob/0cea76258db6fedbef54d4b69518b5e320121fc5/labelbox/schema/project.py#L169)  checks for the server to respond with whether or not it should poll. Is this not available for issues? I assume the implementation of the exporter is similar.

labelbox/schema/project.py
Rodion13: Oh, I am reusing the same query that is used in product. Looks like I need to create a separete API just for SDK to make sure the convention is followed? This way I can be more specific about retries as well

Rodion13:
> Do you mind adding a test case for this?
Added a test. We don't have a way to create issues via an API for now, so I am not testing the content of the issues export file. Hope that's ok

msokoloff1:
This query is marked as internal and isn't going to work until that is changed. https://github.com/Labelbox/intelligence/blob/504ee77b516f7783a4c87f86d4b9674f6eb7ae4d/apps/api/src/issues-comments/issues-comments.graphql#L119

Rodion13:
> The test you added is fine.
> 
> but the query is marked as internal and isn't going to work until that is changed. https://github.com/Labelbox/intelligence/blob/504ee77b516f7783a4c87f86d4b9674f6eb7ae4d/apps/api/src/issues-comments/issues-comments.graphql#L119

Thanks! I have a PR open to remove the @internal flag: https://github.com/Labelbox/intelligence/pull/4063. I am waiting for it to get reviewed

tests/integration/test_issues.py
```diff
@@ -0,0 +1,25 @@
...
+
+    exported_issues_url = project.export_issues("Open")
+    assert exported_issues_url is not None
+    assert exported_issues_url is not ''
+    assert "?status=Open" in exported_issues_url
+
+    exported_issues_url = project.export_issues("Resolved")
+    assert exported_issues_url is not None
+    assert exported_issues_url is not ''
+    assert "?status=Resolved" in exported_issues_url
```
msokoloff1, line 17: Each of these groups of two could be squashed into:
```
assert exported_issues_url
```
Since both None and '' are falsy

labelbox/schema/project.py
```diff
@@ -177,6 +177,40 @@ def export_labels(self, timeout_seconds=60):
...
+        """
+        id_param = "projectId"
+        status_param = "status"
+        query_str = """query GetProjectIssuesExportPyApi($%s: ID!, $%s: IssueStatus) {
+            project(where: { id: $%s }) {
+                issueExportUrl(where: { status: $%s })
+            }
+        }""" % (id_param, status_param, id_param, status_param)
+
+        valid_statuses = [None, "Open", "Resolved"]
```
msokoloff1, line 197: Just make this a set
valid_statuses = {None, "Open", "Resolved"}

msokoloff1 (changes requested):

msokoloff1 (approved):

PR #135, opened 2021-04-05
ms/datarows in queue

msokoloff1:
https://labelbox.atlassian.net/browse/PLAT-1539

* Users want to be able to list all DataRows that are currently in the queue. There is a new method called `queued_data_rows` for doing this.
* Added test case
* Updated examples to include new function.

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -147,7 +147,7 @@ class Project(DbObject, Updateable, Deletable):
 
     def queued_data_rows(self):
         """
-        This function returns all data_rows that have not been labeled yet. 
+        This function returns all data rows that have not yet been labeled in this project. 
         It is helpful for MAL if a user only wants to upload inferences to labels that are still in the queue.
         This is updated after attaching a dataset. This will be accurate even if the project isn't set up yet.
 
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -152,7 +152,7 @@ class Project(DbObject, Updateable, Deletable):
         This is updated after attaching a dataset. This will be accurate even if the project isn't set up yet.
 
         Returns:
-            datarows (List[dataRows]): All datarows that are currently in the queue.
+            All DataRows that are currently in the queue as a PaginatedCollection
 
         """
         id_param = "projectId"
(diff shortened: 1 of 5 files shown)


labelbox/schema/project.py
```diff
@@ -145,6 +145,28 @@ def labels(self, datasets=None, order_by=None):
...
+        This function returns all DataRows that have not yet been labeled in this Project. 
+        It is helpful for MAL if a user only wants to upload inferences to DataRows that are still in the queue.
+        This is updated after attaching a Dataset. This will be accurate even if the Project isn't set up yet.
+
+        Returns:
+            All DataRows that are currently in the queue as a PaginatedCollection
+
+        """
+        id_param = "projectId"
+        query_str = """query GetProjectQueuedDatarowsPyApi($%s: ID!){
```
nmaswood, line 159: Is there a reason the API use `strf` style string formatting as opposed to format strings or `.format`

I feel like either one would be better.

labelbox/schema/project.py
```diff
@@ -145,6 +145,28 @@ def labels(self, datasets=None, order_by=None):
...
+        project(where:{id: $%s}){
+            labelReservations(where:{status:IN_COMPLETE, reservedBy:{id_starts_with:""}} skip: %%d first: %%d)
+            {dataRow  {%s}}}}""" % (id_param, id_param,
+                                    query.results_query_part(DataRow))
+
+        return PaginatedCollection(
```
nmaswood, line 165: Something that I cannot confirm for which it may be better to pull someone else in is:

is this anything that we ought to not expose to users, like will exposing this particular method be taxing on our servers or cause us headaches down the line?

nmaswood:
Meta:

I am not great about this myself, but I think it would be useful if you formatted the PR description according to some consistent standard a la:

# Changes

# Head ups

as we use for our internal PR.

Honestly, use of JIRA I don't think would be a bad thing either, as much as I hate on JIRA. I think it could help visually organize bodies of work too. e.g. like a documentation epic or something like that

labelbox/schema/project.py
msokoloff1: I think the reason that it uses strf is because you don't need to add extra `{}` to support the other approaches. I just kept this pattern for consistency.

msokoloff1: E.g. to represent `{ id: 1}` you have to write `"{{ id: {}}}".format(1)` or `print(f"{{ id: {1}}}")` which is kind of gross.

labelbox/schema/project.py
msokoloff1: I don't think that this is a big deal but I will confirm.

nmaswood (approved):
LGTM, I would just ask that you dig a little to make sure there is no harm in exposing this method. 

I can help you by looking at datadog / asking around as well.

msokoloff1:
Waiting on backend changes. Will reopen when that is complete.

PR #138, opened 2021-04-12
Ms/user access management

msokoloff1:
### [PLAT-1224](https://labelbox.atlassian.net/browse/PLAT-1224)

The purpose of this pr is to allow customers to manage organization invites and user permissions from the SDK. The following functionality was added:

* create / ~~update / revoke invite~~
* ~~query for active invites~~
* query for remaining allowed invites to an organization
* query for remaining available seats in an organization
* set and update organization roles
* assign users to projects
    * set / update / revoke project role
* delete users from org
* Example notebook

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -126,7 +126,7 @@ class Client:
 
             if experimental:
                 endpoint = self.endpoint.replace('graphql', '_gql')
-                logger.warn(
+                logger.info(
                     "Experimental queries/mutations aren't guarenteed to maintain their interface."
                 )
 
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -3,7 +3,6 @@ import json
 import logging
 import mimetypes
 import os
-from typing import Tuple
 
 from google.api_core import retry
 import requests
@@ -63,13 +62,13 @@ class Client:
                     "Labelbox API key not provided")
             api_key = os.environ[_LABELBOX_API_KEY]
         self.api_key = api_key
-        self.enable_beta = enable_beta
-
+    
+        self.enable_beta = True
         if enable_beta:
             logger.info("Beta features have been enabled")
+        
 
         logger.info("Initializing Labelbox client at '%s'", endpoint)
-
         self.endpoint = endpoint
         self.headers = {
             'Accept': 'application/json',
@@ -77,10 +76,11 @@ class Client:
             'Authorization': 'Bearer %s' % api_key,
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
+    
 
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
-    def execute(self, query, params=None, timeout=30.0):
+    def execute(self, query, params=None, timeout=30.0, beta = False):
         """ Sends a request to the server for the execution of the
         given query.
 
@@ -126,7 +126,11 @@ class Client:
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
-            response = requests.post(self.endpoint,
+            endpoint = self.endpoint
+            if beta:
+                endpoint = endpoint.replace('/graphql', '/_gql')
+
+            response = requests.post(endpoint,
                                      data=data,
                                      headers=self.headers,
                                      timeout=timeout)
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -37,7 +37,7 @@ class Client:
     def __init__(self,
                  api_key=None,
                  endpoint='https://api.labelbox.com/graphql',
-                 enable_beta = False):
+                 enable_beta=False):
         """ Creates and initializes a Labelbox Client.
 
         Logging is defaulted to level WARNING. To receive more verbose
@@ -62,11 +62,10 @@ class Client:
                     "Labelbox API key not provided")
             api_key = os.environ[_LABELBOX_API_KEY]
         self.api_key = api_key
-    
+
         self.enable_beta = True
         if enable_beta:
             logger.info("Beta features have been enabled")
-        
 
         logger.info("Initializing Labelbox client at '%s'", endpoint)
         self.endpoint = endpoint
@@ -76,11 +75,10 @@ class Client:
             'Authorization': 'Bearer %s' % api_key,
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
-    
 
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
-    def execute(self, query, params=None, timeout=30.0, beta = False):
+    def execute(self, query, params=None, timeout=30.0, beta=False):
         """ Sends a request to the server for the execution of the
         given query.
 
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -37,7 +37,7 @@ class Client:
     def __init__(self,
                  api_key=None,
                  endpoint='https://api.labelbox.com/graphql',
-                 enable_beta=False):
+                 enable_experimental=False):
         """ Creates and initializes a Labelbox Client.
 
         Logging is defaulted to level WARNING. To receive more verbose
@@ -50,7 +50,7 @@ class Client:
         Args:
             api_key (str): API key. If None, the key is obtained from the "LABELBOX_API_KEY" environment variable.
             endpoint (str): URL of the Labelbox server to connect to.
-            enable_beta (bool): Indicated whether or not to use beta features
+            enable_experimental (bool): Indicated whether or not to use experimental features
         Raises:
             labelbox.exceptions.AuthenticationError: If no `api_key`
                 is provided as an argument or via the environment
@@ -63,9 +63,9 @@ class Client:
             api_key = os.environ[_LABELBOX_API_KEY]
         self.api_key = api_key
 
-        self.enable_beta = True
-        if enable_beta:
-            logger.info("Beta features have been enabled")
+        self.enable_experimental = enable_experimental
+        if enable_experimental:
+            logger.info("Experimental features have been enabled")
 
         logger.info("Initializing Labelbox client at '%s'", endpoint)
         self.endpoint = endpoint
@@ -78,7 +78,7 @@ class Client:
 
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
-    def execute(self, query, params=None, timeout=30.0, beta=False):
+    def execute(self, query, params=None, timeout=30.0, experimental=False):
         """ Sends a request to the server for the execution of the
         given query.
 
@@ -125,7 +125,7 @@ class Client:
 
         try:
             response = requests.post(self.endpoint.replace('/graphql', '/_gql')
-                                     if beta else self.endpoint,
+                                     if experimental else self.endpoint,
                                      data=data,
                                      headers=self.headers,
                                      timeout=timeout)
@@ -213,9 +213,9 @@ class Client:
         if internal_server_error is not None:
             message = internal_server_error.get("message")
 
-            if message.startswith("Syntax Error"):
+            if message.startswith("Syntax Error") or message.startswith(
+                    "Invite(s) cannot be sent"):
                 raise labelbox.exceptions.InvalidQueryError(message)
-
             else:
                 raise labelbox.exceptions.InternalServerError(message)
 
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -213,8 +213,7 @@ class Client:
         if internal_server_error is not None:
             message = internal_server_error.get("message")
 
-            if message.startswith("Syntax Error") or message.startswith(
-                    "Invite(s) cannot be sent"):
+            if message.startswith(("Syntax Error", "Invite(s) cannot be sent")):
                 raise labelbox.exceptions.InvalidQueryError(message)
             else:
                 raise labelbox.exceptions.InternalServerError(message)
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -505,8 +505,8 @@ class Client:
 
     def get_roles(self):
         """
-        
         Returns:
-            Roles: Object containing all valid user and org level permissions
+            Roles: Provides information on available roles within an organization. 
+            Roles are used for user management.
         """
         return Roles(self)
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -396,7 +396,6 @@ class Client:
         where = not_deleted if where is None else where & not_deleted
         query_str, params = query.get_all(db_object_type, where)
 
-        print(query_str, params)
         return PaginatedCollection(
             self, query_str, params,
             [utils.camel_case(db_object_type.type_name()) + "s"],
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -72,8 +72,8 @@ class Client:
         }
 
     #TODO: Add exponential backoff so we don'tt overwhelm the api
-    @retry.Retry(predicate=retry.if_exception_type(
-        labelbox.exceptions.InternalServerError))
+    #@retry.Retry(predicate=retry.if_exception_type(
+    #    labelbox.exceptions.InternalServerError))
     def execute(self, query, params=None, timeout=30.0, experimental = False):
         """ Sends a request to the server for the execution of the
         given query.
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -1,5 +1,6 @@
 from datetime import datetime, timezone
 import json
+from labelbox.schema.role import Role, Roles
 import logging
 import mimetypes
 import os
@@ -86,6 +87,8 @@ class Client:
             params (dict): Query parameters referenced within the query.
             timeout (float): Max allowed time for query execution,
                 in seconds.
+            experimental (bool): Enables users to query experimental features.
+                Users should be aware that these features could change slightly over time.
         Returns:
             dict, parsed JSON response.
         Raises:
@@ -120,7 +123,12 @@ class Client:
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
-            endpoint = self.endpoint.replace('graphql', '_gql') if experimental else self.endpoint 
+            endpoint = self.endpoint
+
+            if experimental:
+                endpoint = self.endpoint.replace('graphql', '_gql')
+                logger.warn("Experimental queries/mutations aren't guarenteed to maintain their interface.")
+            
             response = requests.post(endpoint,
                                      data=data,
                                      headers=self.headers,
@@ -386,6 +394,8 @@ class Client:
         not_deleted = db_object_type.deleted == False
         where = not_deleted if where is None else where & not_deleted
         query_str, params = query.get_all(db_object_type, where)
+
+        print(query_str, params)
         return PaginatedCollection(
             self, query_str, params,
             [utils.camel_case(db_object_type.type_name()) + "s"],
@@ -491,3 +501,12 @@ class Client:
                 any of the attribute names given in kwargs.
         """
         return self._create(Project, kwargs)
+
+    def get_roles(self):
+        """
+        
+        Returns:
+            Roles: Object containing all valid user and org level permissions
+        """
+        return Roles(self)
+
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -124,11 +124,8 @@ class Client:
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
-            endpoint = self.endpoint
-            if beta:
-                endpoint = endpoint.replace('/graphql', '/_gql')
-
-            response = requests.post(endpoint,
+            response = requests.post(self.endpoint.replace('/graphql', '/_gql')
+                                     if beta else self.endpoint,
                                      data=data,
                                      headers=self.headers,
                                      timeout=timeout)
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -50,7 +50,7 @@ class Client:
         Args:
             api_key (str): API key. If None, the key is obtained from the "LABELBOX_API_KEY" environment variable.
             endpoint (str): URL of the Labelbox server to connect to.
-            enable_experimental (bool): Indicated whether or not to use experimental features
+            enable_experimental (bool): Indicates whether or not to use experimental features
         Raises:
             labelbox.exceptions.AuthenticationError: If no `api_key`
                 is provided as an argument or via the environment
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -74,7 +74,7 @@ class Client:
     #TODO: Add exponential backoff so we don'tt overwhelm the api
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
-    def execute(self, query, params=None, timeout=30.0):
+    def execute(self, query, params=None, timeout=30.0, experimental = False):
         """ Sends a request to the server for the execution of the
         given query.
 
@@ -120,7 +120,8 @@ class Client:
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
-            response = requests.post(self.endpoint,
+            endpoint = self.endpoint.replace('graphql', '_gql') if experimental else self.endpoint 
+            response = requests.post(endpoint,
                                      data=data,
                                      headers=self.headers,
                                      timeout=timeout)
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -18,7 +18,7 @@ from labelbox.schema.dataset import Dataset
 from labelbox.schema.user import User
 from labelbox.schema.organization import Organization
 from labelbox.schema.labeling_frontend import LabelingFrontend
-from labelbox.schema.role import Roles
+from labelbox.schema import role
 from labelbox import __version__ as SDK_VERSION
 
 logger = logging.getLogger(__name__)
@@ -503,4 +503,4 @@ class Client:
             Roles: Provides information on available roles within an organization. 
             Roles are used for user management.
         """
-        return Roles(self)
+        return role.get_roles(self)
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -1,6 +1,5 @@
 from datetime import datetime, timezone
 import json
-from labelbox.schema.role import Role, Roles
 import logging
 import mimetypes
 import os
@@ -20,6 +19,7 @@ from labelbox.schema.dataset import Dataset
 from labelbox.schema.user import User
 from labelbox.schema.organization import Organization
 from labelbox.schema.labeling_frontend import LabelingFrontend
+from labelbox.schema.role import Roles
 from labelbox import __version__ as SDK_VERSION
 
 logger = logging.getLogger(__name__)
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -37,7 +37,8 @@ class Client:
 
     def __init__(self,
                  api_key=None,
-                 endpoint='https://api.labelbox.com/graphql'):
+                 endpoint='https://api.labelbox.com/graphql',
+                 enable_beta = False):
         """ Creates and initializes a Labelbox Client.
 
         Logging is defaulted to level WARNING. To receive more verbose
@@ -50,6 +51,7 @@ class Client:
         Args:
             api_key (str): API key. If None, the key is obtained from the "LABELBOX_API_KEY" environment variable.
             endpoint (str): URL of the Labelbox server to connect to.
+            enable_beta (bool): Indicated whether or not to use beta features
         Raises:
             labelbox.exceptions.AuthenticationError: If no `api_key`
                 is provided as an argument or via the environment
@@ -61,6 +63,10 @@ class Client:
                     "Labelbox API key not provided")
             api_key = os.environ[_LABELBOX_API_KEY]
         self.api_key = api_key
+        self.enable_beta = enable_beta
+
+        if enable_beta:
+            logger.info("Beta features have been enabled")
 
         logger.info("Initializing Labelbox client at '%s'", endpoint)
 
@@ -74,7 +80,7 @@ class Client:
 
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
-    def execute(self, query, params=None, timeout=30.0, experimental=False):
+    def execute(self, query, params=None, timeout=30.0):
         """ Sends a request to the server for the execution of the
         given query.
 
@@ -86,8 +92,6 @@ class Client:
             params (dict): Query parameters referenced within the query.
             timeout (float): Max allowed time for query execution,
                 in seconds.
-            experimental (bool): Enables users to query experimental features.
-                Users should be aware that these features could change slightly over time.
         Returns:
             dict, parsed JSON response.
         Raises:
@@ -122,15 +126,7 @@ class Client:
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
-            endpoint = self.endpoint
-
-            if experimental:
-                endpoint = self.endpoint.replace('graphql', '_gql')
-                logger.info(
-                    "Experimental queries/mutations aren't guarenteed to maintain their interface."
-                )
-
-            response = requests.post(endpoint,
+            response = requests.post(self.endpoint,
                                      data=data,
                                      headers=self.headers,
                                      timeout=timeout)
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -72,10 +72,9 @@ class Client:
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
 
-    #TODO: Add exponential backoff so we don'tt overwhelm the api
-    #@retry.Retry(predicate=retry.if_exception_type(
-    #    labelbox.exceptions.InternalServerError))
-    def execute(self, query, params=None, timeout=30.0, experimental = False):
+    @retry.Retry(predicate=retry.if_exception_type(
+        labelbox.exceptions.InternalServerError))
+    def execute(self, query, params=None, timeout=30.0, experimental=False):
         """ Sends a request to the server for the execution of the
         given query.
 
@@ -127,8 +126,10 @@ class Client:
 
             if experimental:
                 endpoint = self.endpoint.replace('graphql', '_gql')
-                logger.warn("Experimental queries/mutations aren't guarenteed to maintain their interface.")
-            
+                logger.warn(
+                    "Experimental queries/mutations aren't guarenteed to maintain their interface."
+                )
+
             response = requests.post(endpoint,
                                      data=data,
                                      headers=self.headers,
@@ -509,4 +510,3 @@ class Client:
             Roles: Object containing all valid user and org level permissions
         """
         return Roles(self)
-
--- a/labelbox/orm/db_object.py
+++ b/labelbox/orm/db_object.py
@@ -1,5 +1,5 @@
 from datetime import datetime, timezone
-from functools import partial
+from functools import wraps
 import logging
 
 from labelbox import utils
@@ -258,11 +258,12 @@ class BulkDeletable:
 
 def beta(fn):
 
+    @wraps(fn)
     def wrapper(self, *args, **kwargs):
         if not self.client.enable_beta:
             raise Exception(
                 f"This function {fn.__name__} relies on a beta feature in the api. This means that the interface could change."
-                " Set `enable_beta=True` in the client to enable use of these functions."
+                " Set `enable_beta=True` in the client to enable use of beta functions."
             )
         return fn(self, *args, **kwargs)
 
--- a/labelbox/orm/db_object.py
+++ b/labelbox/orm/db_object.py
@@ -252,3 +252,20 @@ class BulkDeletable:
         a call to this you should not use this DB object anymore.
         """
         type(self).bulk_delete([self])
+
+
+
+def beta(fn):
+    def wrapper(self, *args, **kwargs):
+        if not isinstance(self, DbObject):
+            raise TypeError("Cannot decorate functions that are not functions of `DbOjects` with `beta` decorator")
+        if not self.client.enable_beta:
+            logger.warning(
+                f"This function {fn.__name__} relies on a beta feature in the api. This means that the interface could change."
+                " Set `enable_beta=True` in the client to silence this warning.")
+            self.client.endpoint = self.client.endpoint.replace("/graphql", "/_gql")
+            result = fn(self, *args, **kwargs)
+            self.client.endpoint = self.client.endpoint.replace("/_gql", "/graphql")
+            return result
+    return wrapper
+    
--- a/labelbox/orm/db_object.py
+++ b/labelbox/orm/db_object.py
@@ -256,14 +256,14 @@ class BulkDeletable:
         type(self).bulk_delete([self])
 
 
-def beta(fn):
+def experimental(fn):
 
     @wraps(fn)
     def wrapper(self, *args, **kwargs):
-        if not self.client.enable_beta:
+        if not self.client.enable_experimental:
             raise Exception(
-                f"This function {fn.__name__} relies on a beta feature in the api. This means that the interface could change."
-                " Set `enable_beta=True` in the client to enable use of beta functions."
+                f"This function {fn.__name__} relies on a experimental feature in the api. This means that the interface could change."
+                " Set `enable_experimental=True` in the client to enable use of experimental functions."
             )
         return fn(self, *args, **kwargs)
 
--- a/labelbox/orm/db_object.py
+++ b/labelbox/orm/db_object.py
@@ -43,6 +43,7 @@ class DbObject(Entity):
         self.client = client
         self._set_field_values(field_values)
 
+    
         for relationship in self.relationships():
             setattr(self, relationship.name,
                     RelationshipManager(self, relationship))
--- a/labelbox/orm/db_object.py
+++ b/labelbox/orm/db_object.py
@@ -42,7 +42,6 @@ class DbObject(Entity):
         """
         self.client = client
         self._set_field_values(field_values)
-
     
         for relationship in self.relationships():
             setattr(self, relationship.name,
--- a/labelbox/orm/db_object.py
+++ b/labelbox/orm/db_object.py
@@ -1,8 +1,6 @@
-from dataclasses import field
 from datetime import datetime, timezone
 from functools import partial
 import logging
-from os import EX_CANTCREAT
 
 from labelbox import utils
 from labelbox.exceptions import InvalidQueryError, InvalidAttributeError
@@ -48,7 +46,7 @@ class DbObject(Entity):
 
         for relationship in self.relationships():
             value = field_values.get(relationship.name)
-            if relationship.precompute and value is None:
+            if relationship.cache and value is None:
                 raise KeyError(
                     f"Expected field  values for {relationship.name}")
 
--- a/labelbox/orm/db_object.py
+++ b/labelbox/orm/db_object.py
@@ -259,23 +259,11 @@ class BulkDeletable:
 def beta(fn):
 
     def wrapper(self, *args, **kwargs):
-        if not isinstance(self, DbObject):
-            raise TypeError(
-                "Cannot decorate functions that are not functions of `DbOjects` with `beta` decorator"
-            )
         if not self.client.enable_beta:
             raise Exception(
                 f"This function {fn.__name__} relies on a beta feature in the api. This means that the interface could change."
                 " Set `enable_beta=True` in the client to enable use of these functions."
             )
-        execute_fn = self.client.execute
-        try:
-            self.client.execute = partial(execute_fn, beta=True)
-            result = fn(self, *args, **kwargs)
-            if isinstance(result, PaginatedCollection):
-                result.beta = True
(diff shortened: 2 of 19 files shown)


labelbox/client.py
```diff
@@ -71,10 +72,9 @@ def __init__(self,
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
 
-    #TODO: Add exponential backoff so we don'tt overwhelm the api
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
-    def execute(self, query, params=None, timeout=30.0):
+    def execute(self, query, params=None, timeout=30.0, experimental=False):
```
nmaswood, line 77: Do you think this should happen at the function level or when a user init's their SDK?

What If I call execute a bunch of times and each time I have to remember to pass in experimental? Feel like it will make my code more messy as opposed to opting into it once

That will also let you just change the graphql endpoint exactly one

labelbox/client.py
```diff
@@ -120,7 +122,15 @@ def convert_value(value):
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
-            response = requests.post(self.endpoint,
+            endpoint = self.endpoint
+
+            if experimental:
+                endpoint = self.endpoint.replace('graphql', '_gql')
+                logger.info(
+                    "Experimental queries/mutations aren't guarenteed to maintain their interface."
```
nmaswood, line 130: We might want to pass a header in and log it in our analtics so we know how often users are making experimental queries.

nmaswood: Also you want  to do `logger.warn` and then explain that their code can break. Also you might want to only do this once per session as to not crowd the users logs

labelbox/client.py
```diff
@@ -490,3 +501,11 @@ def create_project(self, **kwargs):
                 any of the attribute names given in kwargs.
         """
         return self._create(Project, kwargs)
+
+    def get_roles(self):
```
nmaswood, line 505: Can this just be `roles`

and then be a class property?

labelbox/orm/db_object.py
```diff
@@ -44,8 +45,11 @@ def __init__(self, client, field_values):
         self._set_field_values(field_values)
 
         for relationship in self.relationships():
+            value = field_values[
```
nmaswood, line 48: can you do `field_values.get(relationship.name)` instead?

labelbox/orm/db_object.py
```diff
@@ -107,6 +111,7 @@ def __init__(self, source, relationship):
         self.supports_filtering = True
         self.supports_sorting = True
         self.filter_on_id = True
+        self.value = value
```
nmaswood, line 114: Can you explain this change / this new instance var?

labelbox/pagination.py
```diff
@@ -64,6 +92,11 @@ def __next__(self):
             if len(page_data) == 0:
                 raise StopIteration()
 
+            if self.cursor_path:
```
nmaswood, line 95: Can we push all internal cursor state into a single object s.t.

```
class CursorState:
      fetched_all: boolean
      path: something

class OffState:
    _fetched_page: number

PaginationState = Union[CursorState, OffsetState]
```
s.t.  you can't get confused by having the state for both mutually exclusive members floating around

labelbox/schema/invite.py
```diff
@@ -0,0 +1,62 @@
...
+    email = Field.String("email", "inviteeEmail")
+
+    def __init__(self, client, invite_response):
+        project_roles = invite_response.pop("projectInvites", [])
+        super().__init__(client, invite_response)
+
+        self.project_roles = [
+            ProjectRole(
+                project=client.get_project(r['projectId']),
+                role=client.get_roles()[r['projectRoleName'].upper().replace(
```
nmaswood, line 50: this snippet might be more readable if you move `upper.replace` stuff into a seperate named helper method

labelbox/schema/invite.py
```diff
@@ -0,0 +1,62 @@
...
+                    " ", "_")]) for r in project_roles
+        ]
+
+    def revoke(self):
```
nmaswood, line 54: Is this whole class advertised to API users as experimental?

labelbox/schema/organization.py
```diff
@@ -34,3 +42,155 @@ def __init__(self, *args, **kwargs):
...
+
+        data_param = "data"
+        query_str = """mutation createInvitesPyApi($%s: [CreateInviteInput!]){
+                    createInvites(data: $%s){  invite { id createdAt organizationRoleName inviteeEmail}}}""" % (
+            data_param, data_param)
+
+        projects = [{
+            "projectId": x.project.uid,
+            "projectRoleId": x.role.uid
+        } for x in project_roles]
```
nmaswood, line 83: would prefer `project_role` to `x`

labelbox/schema/organization.py
```diff
@@ -34,3 +42,155 @@ def __init__(self, *args, **kwargs):
...
+            },
+            experimental=True)  # We prob want to return an invite
+        # Could support bulk ops in the future
+        invite_info = res['createInvites'][0]['invite']
+        return invite_info
+
+    def invite_user(self,
+                    email: str,
+                    role: Role,
+                    project_roles: List[ProjectRole] = []) -> Invite:
```
nmaswood, line 103: default argument as array can become tricky because this array is a singleton instance for the whole module

`project_roles = Union[List, None]` might be better

labelbox/schema/organization.py
```diff
@@ -34,3 +42,155 @@ def __init__(self, *args, **kwargs):
...
+            role (Role): Role to assign to the user
+            project_roles (List[ProjectRoles]): List of project roles to assign to the User (if they have a project based org role).
+
+        Returns:
+            Invite for the user
+
+        """
+        remaining_invites = self.invite_limit().remaining
+        if remaining_invites == 0:
+            if next(self.users((User.email == email) &
```
nmaswood, line 118: this if statement logic is a handleful might be nice to break into another function

labelbox/schema/role.py
```diff
@@ -0,0 +1,72 @@
...
+class Roles:
+    """
+    Object that manages org and user roles
+
+        >>> roles = client.get_roles()
+        >>> roles.valid_roles # lists all valid roles
+        >>> roles['ADMIN'] # returns the admin Role
+
+    """
+    _instance = None
```
nmaswood, line 20: Is this `_instance` value suppoed to be static?

labelbox/schema/role.py
```diff
@@ -0,0 +1,72 @@
...
+                valid_roles.add(_name)
+            cls._instance.valid_roles = valid_roles
+        cls._instance.index = 0
+        return cls._instance
+
+    def __repr__(self):
+        return str({k: getattr(self, k) for k in self.valid_roles})
+
+    def __getitem__(self, name):
+        if name in self.valid_roles:
```
nmaswood, line 41: If you flip this logic it is more concise:


```
if not in names:
    throw
return getattr(self, name)
```

labelbox/schema/role.py
```diff
@@ -0,0 +1,72 @@
...
+    ...
+
+
+class UserRole(Role):
+    ...
+
+
+@dataclass
+class ProjectRole:
+    project: "Project"
```
nmaswood, line 71: why is this harded codeD?

labelbox/schema/user.py
```diff
@@ -35,3 +38,69 @@ class User(DbObject):
...
+    def upsert_project_role(self, project: Project, role: Role):
+        """ Updates or replaces a User's role in a project.
+
+        Args:
+            project (Project): The project to update the users permissions for
+            role (Role): The role to assign to this user in this project.
+        
+        """
+        org_role = self.org_role()
+        if org_role.name.upper() != 'NONE':
```
nmaswood, line 83: is the value `NONE` being returned by the backend?

nmaswood (commented):
Going to take a second look tommorow when I am fresh!

I might ask Greg to also take a look, because I think `experimental` mode is a big change and we should be cautious in how we roll it out in terms of the API

also I think Grant/Collin might be helpful in providing some context of how the invite api ought to be used.

labelbox/client.py
msokoloff1: That would be a different access pattern than the other db objects. Although this one behaves differently because it can't be modified. What do you think?

labelbox/orm/db_object.py
msokoloff1: It should throw a KeyError if `precompute = True` and the key does not exist. I could break this out into 
```
value =   field_values.get(relationship.name)
if relationship.precompute and value is None:
    raise KeyError(f"Expected field  values for {relationship.name}"
```
Does that make more sense?

labelbox/orm/db_object.py
msokoloff1: I added a comment on the relationship manager 
``` 
Whether or not to precompute the relationship.
           Useful for objects that aren't directly queryable from the api (relationship query builder won't work)
          Also useful for expensive ToOne relationships 
```
In this case we get `member`s back from a custom query. Since a `member` isn't directly queryable it can't be represented as a dbobject even though we have all of the information to construct the object. This enables us to pass the information to create this object without an additional query.

This is also useful if you have an object that has a relationship that needs to be used for every result. E.g. labeling parameter overrides usually need data rows to do anything useful. However, you need query for each individual data row which. This enables us to do the join on the backend and drastically cut the number of requests.

labelbox/pagination.py
msokoloff1: Yes, I like this idea.

labelbox/schema/organization.py
msokoloff1: fixed

labelbox/schema/organization.py
msokoloff1: fixed. Changed it to `project_roles: Optional[List[ProjectRole]] = None`

labelbox/schema/role.py
msokoloff1: yes. This will never change during a single user session.

labelbox/schema/role.py
msokoloff1: fixed

labelbox/schema/user.py
msokoloff1: Yes

labelbox/schema/role.py
msokoloff1: This isn't hard coded, it is just setting the type using a string. There is a circular import between project and role. The type hinting still works with ides, it is just a touch uglier.

labelbox/schema/invite.py
msokoloff1: fixed

labelbox/schema/organization.py
msokoloff1: This actually isn't necessary anymore. I removed it.

labelbox/client.py
```diff
@@ -49,6 +51,7 @@ def __init__(self,
         Args:
             api_key (str): API key. If None, the key is obtained from the "LABELBOX_API_KEY" environment variable.
             endpoint (str): URL of the Labelbox server to connect to.
+            enable_beta (bool): Indicated whether or not to use beta features
```
nmaswood, line 54: In the comment I would explain the danger of using beta features.

The name beta scares me as well because it implies that we are seriously considering making them G.A., but I'm not sure if that is the case here.

labelbox/orm/db_object.py
nmaswood: I think it might, this statement is a little hard to grok

labelbox/orm/db_object.py
```diff
@@ -244,3 +252,20 @@ def delete(self):
...
+
+def beta(fn):
+    def wrapper(self, *args, **kwargs):
+        if not isinstance(self, DbObject):
+            raise TypeError("Cannot decorate functions that are not functions of `DbOjects` with `beta` decorator")
+        if not self.client.enable_beta:
+            logger.warning(
+                f"This function {fn.__name__} relies on a beta feature in the api. This means that the interface could change."
+                " Set `enable_beta=True` in the client to silence this warning.")
+            self.client.endpoint = self.client.endpoint.replace("/graphql", "/_gql")
```
nmaswood, line 266: I do not like modifying `self.client`

1. on principal seems a little hacky
2. if `fn` throws and exception the clean up code `self.client.endpoint` won't get run and restored


we could fix that in a try/finally of some sort but even then I would prefer

that `beta` functions don't execute and instead utilize their own method `execute_beta` or something like that which has the proper end point. Since `beta` methods are all annotated with a decotator it shouldn't be too hard to go back and change them to regular `execute` if we need to.


Also if they don't have `enable_beta` as true should we just throw?

The current behavior seems more like `warn_on_beta`

nmaswood: Maybe cursor pagination can take an execute function then as opposed to being given the client

labelbox/pagination.py
```diff
@@ -25,45 +32,90 @@ def __init__(self, client, query, params, dereferencing, obj_class):
...
+        self.next_cursor = None
+
+    def get_next_cursor(self, results):
+        for path in self.cursor_path:
+            results = results[path]
+        return results
+
+    def fetched_all(self, n_items, results):
+        self.next_cursor = self.get_next_cursor(results)
+        if self.next_cursor is None:
```
nmaswood, line 98: You can just do `return self.next_cursor is None`

labelbox/orm/db_object.py
msokoloff1: ok, I'll update it

labelbox/client.py
msokoloff1: We are going to do a full release of these features. We are only holding off until we can make sure they are working as expected.

labelbox/client.py
msokoloff1: this is all being managed by the `beta` decorator. Adding here for context.

labelbox/client.py
msokoloff1: This doesn't exist anymore.

labelbox/pagination.py
msokoloff1: fixed.

labelbox/schema/invite.py
msokoloff1: This has been removed.

labelbox/orm/db_object.py
msokoloff1: this is all being managed by the beta decorator. Adding here for context.

labelbox/pagination.py
msokoloff1: Fixed.

msokoloff1:
After speaking with Grant, we are going to remove the ability to query for active invites and revoke them since this might behave unexpectedly. He also said we need to remove the old user_limit query and just use the new invite_limit query. All of these requests have been made.

labelbox/pagination.py
```diff
@@ -25,45 +31,88 @@ def __init__(self, client, query, params, dereferencing, obj_class):
...
+    def fetched_all(self, n_items, results):
+        self.next_cursor = self.get_next_cursor(results)
+        return self.next_cursor is None
+
+    def fetch_results(self, query, params, beta):
+        params.update({'from': self.next_cursor, 'first': _PAGE_SIZE})
+        return self.client.execute(query, params, beta=beta)
+
+
+class _OffsetPagination:
```
nmaswood, line 104: I know that duck typing is more common in python, but I think defining a pagination interface with abc.meta could be nice to see what common methods are shared

labelbox/pagination.py
```diff
@@ -25,45 +31,88 @@ def __init__(self, client, query, params, dereferencing, obj_class):
...
+class _CursorPagination:
+
+    def __init__(self, client, cursor_path):
+        self.client = client
+        self.cursor_path = cursor_path
+        self.next_cursor = None
+
+    def get_next_cursor(self, results):
+        for path in self.cursor_path:
+            results = results[path]
```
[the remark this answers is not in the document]
msokoloff1: This is right. It is traversing the nested json.

labelbox/pagination.py
msokoloff1: I thought about this but then got lazy :). I'll make the change.

labelbox/orm/db_object.py
```diff
@@ -244,3 +254,28 @@ def delete(self):
...
+                "Cannot decorate functions that are not functions of `DbOjects` with `beta` decorator"
+            )
+        if not self.client.enable_beta:
+            raise Exception(
+                f"This function {fn.__name__} relies on a beta feature in the api. This means that the interface could change."
+                " Set `enable_beta=True` in the client to enable use of these functions."
+            )
+        execute_fn = self.client.execute
+        try:
+            self.client.execute = partial(execute_fn, beta=True)
```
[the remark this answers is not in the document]
msokoloff1: the decision to use the beta endpoint is determined by the parameter in the execute function. This is so that we can support calling beta endpoints from paginated collections.

msokoloff1:
From @christopheramata
```
Yes the bulk option is becoming a much much more asked question
Especially at enterprise convos
```
I am going to update the invite_users to support bulk updates.

labelbox/pagination.py
msokoloff1: Updated.

labelbox/pagination.py
```diff
@@ -25,45 +31,88 @@ def __init__(self, client, query, params, dereferencing, obj_class):
...
+            self._fetched_all = self.paginator.fetched_all(n_items, results)
+
         rval = self._data[self._data_ind]
         self._data_ind += 1
         return rval
+
+
+class _CursorPagination:
+
+    def __init__(self, client, cursor_path):
```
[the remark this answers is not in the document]
msokoloff1: Updated.

labelbox/client.py
```diff
@@ -71,10 +76,9 @@ def __init__(self,
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
 
-    #TODO: Add exponential backoff so we don'tt overwhelm the api
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
-    def execute(self, query, params=None, timeout=30.0):
+    def execute(self, query, params=None, timeout=30.0, beta=False):
```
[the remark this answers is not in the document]
msokoloff1: only if they write their own graphql queries which they shouldn't be doing.

labelbox/pagination.py
msokoloff1: Updated.

labelbox/orm/db_object.py
```diff
@@ -244,3 +254,28 @@ def delete(self):
         a call to this you should not use this DB object anymore.
         """
         type(self).bulk_delete([self])
+
+
+def beta(fn):
+
+    def wrapper(self, *args, **kwargs):
```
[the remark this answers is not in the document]
msokoloff1: Updated.

labelbox/client.py
```diff
@@ -120,7 +124,11 @@ def convert_value(value):
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
-            response = requests.post(self.endpoint,
+            endpoint = self.endpoint
+            if beta:
+                endpoint = endpoint.replace('/graphql', '/_gql')
+
+            response = requests.post(endpoint,
```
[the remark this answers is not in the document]
msokoloff1: Updated.

labelbox/orm/db_object.py
```diff
@@ -244,3 +254,28 @@ def delete(self):
...
+        if not isinstance(self, DbObject):
+            raise TypeError(
+                "Cannot decorate functions that are not functions of `DbOjects` with `beta` decorator"
+            )
+        if not self.client.enable_beta:
+            raise Exception(
+                f"This function {fn.__name__} relies on a beta feature in the api. This means that the interface could change."
+                " Set `enable_beta=True` in the client to enable use of these functions."
```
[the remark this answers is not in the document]
msokoloff1: Changed to enable use of beta functions. Since it enable this function and others.

labelbox/orm/db_object.py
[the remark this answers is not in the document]
msokoloff1: This has been removed because it wasn't thread safe

labelbox/pagination.py
```diff
@@ -25,45 +40,101 @@ def __init__(self, client, query, params, dereferencing, obj_class):
...
+        self.cursor_path = cursor_path
+        self.next_cursor: Optional[str] = None
+
+    def get_next_cursor(self, results) -> Optional[str]:
+        for path in self.cursor_path:
+            results = results[path]
+        return results
+
+    def fetched_all(self, n_items: int, results: List[Dict[str, Any]]) -> bool:
+        self.next_cursor = self.get_next_cursor(results)
```
gszpak, line 116: Imho it's a bit confusing that `fetched_all` updates the cursor. Judging by the name, it should be an idempotent operation. Could we add a new method to the interface and move updating cursor there?

labelbox/schema/project.py
```diff
@@ -78,9 +78,25 @@ class Project(DbObject, Updateable, Deletable):
...
+        """
+        id_param = "projectId"
+        query_str = """query ProjectMemberOverviewPyApi($%s: ID!) {
+             project(where: {id : $%s}) { id members(skip: %%d first: %%d){ id user { %s } role { id name } }
+           }
+        }""" % (id_param, id_param, query.results_query_part(Entity.User))
+        return PaginatedCollection(
+            self.client, query_str, {id_param: str(self.uid)},
+            ["project", "members"],
+            lambda client, res: ProjectMember(client, res))
```
gszpak, line 96: Why does `PaginatedCollection` accept a lambda? Don't see a `Callable` in `PaginatedCollection`'s __init__

labelbox/schema/role.py
```diff
@@ -0,0 +1,69 @@
...
+            res = client.execute(query_str)
+            valid_roles = set()
+            for result in res['roles']:
+                _name = result['name'].upper().replace(' ', '_')
+                result['name'] = _name
+                setattr(cls._instance, _name, Role(client, result))
+                valid_roles.add(_name)
+            cls._instance.valid_roles = valid_roles
+        cls._instance.index = 0
+        return cls._instance
```
gszpak, line 32: In general I'm not a huge fan of overriding `__new__`. Imho the code gets overly complicated. If we want to make them singletons, could we instead declare the roles as lazy initalized module vars - sth like here https://github.com/Labelbox/python-monorepo/blob/6c6bfebb68fc6f876c3abf29c9ce727a6c5744dc/services/prediction_import/prediction_import/db/client.py#L357

labelbox/schema/user.py
```diff
@@ -35,3 +38,69 @@ class User(DbObject):
...
+            raise ValueError(
+                "User is not project based and has access to all projects")
+
+        if not isinstance(role, Role):
+            raise TypeError(f"Must provide a `Role` object. Found {role}")
+
+        if not isinstance(project, Project):
+            raise TypeError(f"Must provide a `Project` object. Found {project}")
```
gszpak, line 91: Same here about type checking

labelbox/orm/db_object.py
```diff
@@ -244,3 +254,28 @@ def delete(self):
...
+            )
+        execute_fn = self.client.execute
+        try:
+            self.client.execute = partial(execute_fn, beta=True)
+            result = fn(self, *args, **kwargs)
```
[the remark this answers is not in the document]
msokoloff1: If beta = True it just calls our _gql endpoint. It won't cause any problems. I'll add a test for it.

labelbox/orm/db_object.py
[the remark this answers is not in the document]
msokoloff1: This is out of date. I changed it to be thread safe.

labelbox/orm/db_object.py
```diff
@@ -244,3 +254,28 @@ def delete(self):
...
+            if isinstance(result, PaginatedCollection):
+                result.beta = True
```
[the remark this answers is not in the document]
msokoloff1: This was so that the PaginatedCollection would keep track of whether or not to call the beta endpoint even when the function decorated with the beta decorator was out of scope. This is no longer used.

labelbox/pagination.py
```diff
@@ -25,45 +31,88 @@ def __init__(self, client, query, params, dereferencing, obj_class):
                 reach the paginated objects of interest.
             obj_class (type): The class of object to be instantiated with each
                 dict containing db values.
+            cursor_path: If not None, this is used to find the cursor
+            experimental: Used to call experimental endpoints
```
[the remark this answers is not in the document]
msokoloff1: Will update

labelbox/pagination.py
```diff
@@ -25,45 +31,88 @@ def __init__(self, client, query, params, dereferencing, obj_class):
...
+        self.cursor_path = cursor_path
+        self.next_cursor = None
+
+    def get_next_cursor(self, results):
+        for path in self.cursor_path:
+            results = results[path]
+        return results
+
+    def fetched_all(self, n_items, results):
+        self.next_cursor = self.get_next_cursor(results)
```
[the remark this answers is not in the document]
msokoloff1: Ok makes sense. Will change.

labelbox/schema/invite.py
```diff
@@ -0,0 +1,37 @@
...
+
+@dataclass
+class InviteLimit:
+    """
+    remaining (int): Number of invites remaining in the org
+    used (int): Number of invites used in the org
+    limit (int): Maximum number of invites available to the org
+    """
+
+    remaining: int
```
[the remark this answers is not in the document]
msokoloff1: It is provided by the backend

labelbox/schema/organization.py
```diff
@@ -34,3 +44,109 @@ def __init__(self, *args, **kwargs):
...
+        """
+        remaining_invites = self.invite_limit().remaining
+        if remaining_invites == 0:
+            raise LabelboxError(
+                "Invite(s) cannot be sent because you do not have enough available seats in your organization. "
+                "Please upgrade your account, revoke pending invitations or remove other users."
+            )
+
+        if not isinstance(role, Role):
+            raise TypeError(f"role must be Role type. Found {role}")
```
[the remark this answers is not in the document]
msokoloff1: This often helps me. I find myself passing ids instead of objects. It might not be pythonic but I think that it adds value. Especially when users feel overwhelmed by the SDK, it is nice to make the errors super clear.

labelbox/schema/organization.py
```diff
@@ -34,3 +44,109 @@ def __init__(self, *args, **kwargs):
...
+            raise ValueError(
+                "Project roles cannot be set for a user with organization level permissions. Found role name `{role.name}`, expected `NONE`"
+            )
+
+        _project_roles = [] if project_roles is None else project_roles
+        for project_role in _project_roles:
+            if not isinstance(project_role, ProjectRole):
+                raise TypeError(
+                    f"project_roles must be a list of `ProjectRole`s. Found {project_role}"
+                )
```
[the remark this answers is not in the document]
msokoloff1: Same comment as above:

This often helps me. I find myself passing ids instead of objects. It might not be pythonic but I think that it adds value. Especially when users feel overwhelmed by the SDK, it is nice to make the errors super clear.

labelbox/pagination.py
msokoloff1: Sounds good

labelbox/schema/project.py
msokoloff1: It accepts a DBObject which is a callable that accepts a client and json response. This pattern has been used elsewhere in the SDK to customize the parsing of the response string. However, it isn't necessary here anymore. I'll update this to just accept `ProjectMember`.

labelbox/schema/role.py
msokoloff1: This is a pretty standard singleton pattern. The code snippet you provided is much simpler but also because it is doing much less. This is going to basically look the same except swap __new__ for __init__ and add a new module level variable which feels like a lateral move rather than an improvement.

labelbox/orm/db_object.py
```diff
@@ -244,3 +254,28 @@ def delete(self):
         a call to this you should not use this DB object anymore.
         """
         type(self).bulk_delete([self])
+
+
+def beta(fn):
```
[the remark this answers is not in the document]
msokoloff1: It was experimental before. I have been bouncing back and force on the naming. If you like experimental I'll switch it back because I have no idea which is best :)

labelbox/schema/role.py
nmaswood: One thing I would say here, is it might be easier to grok this code if you moved the business logic into a named function
```python
def get_available_roles():
 ...
```

nmaswood: I lean more towards, structs/class should be dumb containers of data than locations of business logic.

In that sense I think you could seperate fetching `GetAvailableUserRolesPyApi` logic from the class `__new__` and move it out into a static method or function, imho it doesn make reading the class a little trickier.

```python
class RoleAccessor(self):
    def __init__(self):
         self.roles = get_The_roles(client)
    
```

or some other means to break up the logic

labelbox/schema/user.py
nmaswood: I agree with Greg in the sense of, if we runtime typecheck here, why not runtime type check everywhere in all places / how do you establish when one ought to?

tests/integration/conftest.py
```diff
@@ -46,6 +50,74 @@ def testing_api_key(environ: str) -> str:
     return os.environ["LABELBOX_TEST_API_KEY_STAGING"]
 
 
+def beta_endpoint(fn):
```
nmaswood, line 53: Is this beta_endpoint code copied from elsewhere? Why duplicate?

tests/integration/conftest.py
```diff
@@ -46,6 +50,74 @@ def testing_api_key(environ: str) -> str:
...
+            return fn(client, *args, **kwargs)
+        finally:
+            client.endpoint = client.endpoint.replace("/_gql", "/graphql")
+
+    return beta
+
+
+@beta_endpoint
+def cancel_invite(client, invite_id):
+    """
```
nmaswood, line 67: You could maybe turn this comments into application logic through introducing a dummy decorator that annotates like test method only or something like that.

or like throw if pytest is not running:

https://stackoverflow.com/questions/25188119/test-if-code-is-executed-from-within-a-py-test-session

tests/integration/conftest.py
```diff
@@ -46,6 +50,74 @@ def testing_api_key(environ: str) -> str:
...
+    Do not use. Only for testing.
+    """
+    id_param = "projectId"
+    query_str = """query GetProjectInvitationsPyApi($from: ID, $first: PageSize, $%s: ID!) {
+        project(where: {id: $%s}) {id
+        invites(from: $from, first: $first) { nodes { %s
+        projectInvites { projectId projectRoleName } } nextCursor}}}
+    """ % (id_param, id_param, results_query_part(Invite))
+    return list(
```
nmaswood, line 86: why use `list` here?

tests/integration/conftest.py
```diff
@@ -46,6 +50,74 @@ def testing_api_key(environ: str) -> str:
...
+        Invite,
+        cursor_path=['organization', 'invites', 'nextCursor'])
+    return list(
+        invites)  # list() so that it makes the request to the right endpoint.
+
+
+@pytest.fixture
+def queries():
+    return SimpleNamespace(
+        **{
```
nmaswood, line 114: Strictly counting chars, this might be more conccise without `**` syntax
```python
            cancel_invite=cancel_invite,
            get_project_invites=get_project_invites,
```

tests/integration/conftest.py
```diff
@@ -124,3 +196,47 @@ def sample_video() -> str:
     path_to_video = 'tests/integration/media/cat.mp4'
     assert os.path.exists(path_to_video)
     return path_to_video
+
+
+@pytest.fixture
+def organization(client):
+    # Must have at least one seat open in your org to run these tests
```
nmaswood, line 203: is this gonna make int. tests flakier?

labelbox/utils.py
```diff
@@ -1,3 +1,4 @@
+#from labelbox.orm.db_object import DbObject
```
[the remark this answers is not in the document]
msokoloff1: fixed.

labelbox/pagination.py
```diff
@@ -25,45 +31,88 @@ def __init__(self, client, query, params, dereferencing, obj_class):
...
+
+
+class _OffsetPagination:
+
+    def __init__(self, client):
+        self.client = client
+        self._fetched_pages = 0
+
+    def fetched_all(self, n_items, results):
+        self._fetched_pages += 1
```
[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/pagination.py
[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/schema/organization.py
```diff
@@ -34,3 +44,109 @@ def __init__(self, *args, **kwargs):
...
+
+        if not isinstance(role, Role):
+            raise TypeError(f"role must be Role type. Found {role}")
+
+        if project_roles and role.name != "NONE":
+            raise ValueError(
+                "Project roles cannot be set for a user with organization level permissions. Found role name `{role.name}`, expected `NONE`"
+            )
+
+        _project_roles = [] if project_roles is None else project_roles
```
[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/schema/organization.py
```diff
@@ -34,3 +44,109 @@ def __init__(self, *args, **kwargs):
...
+        for project_role in _project_roles:
+            if not isinstance(project_role, ProjectRole):
+                raise TypeError(
+                    f"project_roles must be a list of `ProjectRole`s. Found {project_role}"
+                )
+
+        if self.client.get_user().email == email:
+            raise ValueError("Cannot update your own role")
```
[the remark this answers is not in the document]
msokoloff1: Fixed

labelbox/schema/organization.py
```diff
@@ -34,3 +44,109 @@ def __init__(self, *args, **kwargs):
...
+        res = self.client.execute(query_str, {
+            data_param: [{
+                "inviterId": self.client.get_user().uid,
+                "inviteeEmail": email,
+                "organizationId": self.uid,
+                "organizationRoleId": role.uid,
+                "projects": projects
+            }]
+        },
+                                  beta=True)
```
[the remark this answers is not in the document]
msokoloff1: yapf does this. Do you want me to break this up into multiple statements?

labelbox/schema/project.py
msokoloff1: Fixed.

tests/integration/conftest.py
msokoloff1: This is slightly different and used just for testing. We are exposing functionality we don't want customers to use but only for testing.

tests/integration/conftest.py
msokoloff1: I want to fetch all results before the result is returned. Otherwise this query won't work since it will be outside of the experimental scope.

tests/integration/conftest.py
msokoloff1: Updated.

tests/integration/conftest.py
msokoloff1: No it clears invites before and after (only @labelbox.com invites so users don't accidentally mess up anything). Our test account is only used for testing so there shouldn't be any problem here. This will only matter if you are testing locally and don't have any open seats.

labelbox/schema/user.py
msokoloff1: Sometimes the error messages from the backend isn't so clear.. I guess I should make a PR to fix that instead of adding typechecks here. But the error here should be clear, e.g. attribute error if they passed a project id instead of a project would be pretty clear. I'll change it.

msokoloff1: Fixed.

labelbox/schema/organization.py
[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/schema/organization.py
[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/schema/organization.py
```diff
@@ -34,3 +44,109 @@ def __init__(self, *args, **kwargs):
...
+        Args:
+            email (str): email address of the user to invite
+            role (Role): Role to assign to the user
+            project_roles (Optional[List[ProjectRoles]]): List of project roles to assign to the User (if they have a project based org role).
+
+        Returns:
+            Invite for the user
+
+        Creates or updates users. This function shouldn't directly be called. 
+        Use `Organization.invite_user`
```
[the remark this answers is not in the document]
msokoloff1: From an old commit. I Removed it.

labelbox/schema/organization.py
```diff
@@ -34,3 +44,109 @@ def __init__(self, *args, **kwargs):
...
+
+        Note: that there is a really unclear foreign key error if you use an unsupported role.
+        - This means that the `Roles` object is not getting the right ids
+        """
+        remaining_invites = self.invite_limit().remaining
+        if remaining_invites == 0:
+            raise LabelboxError(
+                "Invite(s) cannot be sent because you do not have enough available seats in your organization. "
+                "Please upgrade your account, revoke pending invitations or remove other users."
+            )
```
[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/schema/invite.py
```diff
@@ -0,0 +1,37 @@
...
+    limit: int
+
+
+class Invite(DbObject):
+    """
+    An object representing a user invite
+    """
+    created_at = Field.DateTime("created_at")
+    organization_role_name = Field.String("organization_role_name")
+    email = Field.String("email", "inviteeEmail")
```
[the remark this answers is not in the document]
msokoloff1: It requires the cached relationship since user isn't queryable from an invite. I'll add it.

labelbox/pagination.py
msokoloff1: Fixed.

labelbox/orm/db_object.py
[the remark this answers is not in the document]
msokoloff1: Switched back.

labelbox/orm/db_object.py
```diff
@@ -44,8 +45,13 @@ def __init__(self, client, field_values):
         self._set_field_values(field_values)
 
         for relationship in self.relationships():
+            value = field_values.get(relationship.name)
+            if relationship.cache and value is None:
+                raise KeyError(
+                    f"Expected field  values for {relationship.name}")
```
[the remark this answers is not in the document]
msokoloff1: If we need that in the future we can make the change. I think making this more strict is better.

labelbox/schema/role.py
msokoloff1: I changed it to this for now. https://github.com/Labelbox/labelbox-python/blob/e6e4dd6d759e4c4dea61f21735863a1a6cbe4bd3/labelbox/schema/role.py#L8-L52. I'm not totally clear what Nasr is suggesting so we are going to sync tomorrow.

labelbox/orm/db_object.py
```diff
@@ -244,3 +254,28 @@ def delete(self):
...
         type(self).bulk_delete([self])
+
+
+def beta(fn):
+
+    def wrapper(self, *args, **kwargs):
+        if not isinstance(self, DbObject):
+            raise TypeError(
+                "Cannot decorate functions that are not functions of `DbOjects` with `beta` decorator"
+            )
```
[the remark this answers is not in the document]
msokoloff1: No longer checking this condition. This was removed because it wasn't thread safe.

labelbox/orm/db_object.py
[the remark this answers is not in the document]
msokoloff1: No longer used.

tests/integration/conftest.py
msokoloff1: Removed.

tests/integration/conftest.py
msokoloff1: The main point of the comment is to make it so that users don't copy this into their own application. I don't think there is a risk of someone importing from the conftest.

tests/integration/conftest.py
msokoloff1: Removed.

labelbox/schema/role.py
```diff
@@ -0,0 +1,41 @@
+from dataclasses import dataclass
+
+from labelbox.orm.model import Field
+from labelbox.orm.db_object import DbObject
+from labelbox.schema.project import Project
+
+_ROLES = None
+
+
+def get_roles(client):
```
nmaswood, line 10: 🥳 
I might recommend  typing annotations on `ROLES` or `get_roles` but not a big deal

labelbox/schema/role.py
```diff
@@ -0,0 +1,41 @@
...
+        query_str = """query GetAvailableUserRolesPyApi { roles { id name } }"""
+        res = client.execute(query_str)
+        _ROLES = {}
+        for role in res['roles']:
+            role['name'] = format_role(role['name'])
+            _ROLES[role['name']] = Role(client, role)
+    return _ROLES
+
+
+def format_role(name: str):
```
nmaswood, line 22: jw, do we even need to format_roles now that we are not emulating enum behaavior?

msokoloff1: Idk I don't like spaces and inconsistent casing in key names. It makes it harder to use.

labelbox/schema/role.py
msokoloff1: fixed.

labelbox/schema/organization.py
```diff
@@ -1,5 +1,15 @@
...
+from typing import List, Optional
+
+from labelbox.exceptions import LabelboxError
+from labelbox import utils
+from labelbox.orm.db_object import DbObject, experimental, query
 from labelbox.orm.model import Field, Relationship
+from labelbox.schema.invite import Invite, InviteLimit, ProjectRole
+from labelbox.schema.user import User
+from labelbox.schema.role import Role
+import logging
```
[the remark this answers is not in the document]
msokoloff1: yes we do... oops! I need to get my ide configured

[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/schema/invite.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def __init__(self, client, invite_response):
+        #invite_response.get('inviter')
```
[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/pagination.py
```diff
@@ -1,4 +1,12 @@
 # Size of a single page in a paginated query.
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type
+
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
msokoloff1: It is to avoid circular imports. There is not really a great solution here.

labelbox/client.py
```diff
@@ -49,6 +50,7 @@ def __init__(self,
         Args:
             api_key (str): API key. If None, the key is obtained from the "LABELBOX_API_KEY" environment variable.
             endpoint (str): URL of the Labelbox server to connect to.
+            enable_experimental (bool): Indicated whether or not to use experimental features
```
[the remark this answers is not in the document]
msokoloff1: Fixed.

labelbox/client.py
```diff
@@ -208,9 +213,9 @@ def check_errors(keywords, *path):
         if internal_server_error is not None:
             message = internal_server_error.get("message")
 
-            if message.startswith("Syntax Error"):
+            if message.startswith("Syntax Error") or message.startswith(
+                    "Invite(s) cannot be sent"):
```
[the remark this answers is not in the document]
msokoloff1: This is because the backend just throws a 500 and there is no way to handle this without parsing the message (see the comment just above this.

Server side validation is one reason we raise this exception:
https://github.com/Labelbox/labelbox-python/blob/56a366ce102b216a278993908863817282103955/labelbox/exceptions.py#L63-L65

I don't like this pattern either but I don't think fixing this is within the scope of this delivery.

labelbox/schema/organization.py
```diff
@@ -34,3 +44,91 @@ def __init__(self, *args, **kwargs):
...
+            email (str): email address of the user to invite
+            role (Role): Role to assign to the user
+            project_roles (Optional[List[ProjectRoles]]): List of project roles to assign to the User (if they have a project based org role).
+
+        Returns:
+            Invite for the user
+
+        Notes:
+            This function is currently experimental and has a few limitations that will be resolved in future releases
+            1. There is a really unclear foreign key error if you use an unsupported role.
```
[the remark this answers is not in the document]
msokoloff1: Fixed

labelbox/client.py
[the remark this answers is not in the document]
nmaswood: aite, one small nit is that `startswith` takes a tuple I believe
`message.startsiwh(("message1, "message"))`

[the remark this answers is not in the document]
msokoloff1: Fixed.

nmaswood (approved):
@gszpak it all looks good to me, you have any follow up thoughts?

PR #139, opened 2021-04-22
ms/tms mal example

msokoloff1:


labelbox/schema/dataset.py
```diff
@@ -67,7 +67,7 @@ def create_data_row(self, **kwargs):
 
         return self.client._create(DataRow, kwargs)
 
-    def create_data_rows(self, items):
+    def create_data_rows(self, items, tms=False):
```
gszpak, line 70: What does `tms` mean? Could we add it to docstring?

msokoloff1: Tile map service. I will add to the docstring.

msokoloff1: Fixed.

labelbox/schema/dataset.py
```diff
@@ -67,7 +67,7 @@ def create_data_row(self, **kwargs):
 
         return self.client._create(DataRow, kwargs)
 
-    def create_data_rows(self, items):
+    def create_data_rows(self, items, tiled_imagery=False):
```
nmaswood, line 70: Once again not a big fan of continueing to add boolean arguments, could we just add another method that uploads `tiled_imagery` instead of overloading this one?

labelbox/schema/dataset.py
```diff
@@ -135,7 +142,10 @@ def convert_item(item):
             }
 
         # Prepare and upload the desciptor file
-        data = json.dumps([convert_item(item) for item in items])
+        if not tiled_imagery:
```
nmaswood, line 145: Is there no way that we can detect tiled imagery and do the conversion ourselves if necessary? ( rather than the user explicitly telling us)

labelbox/schema/dataset.py
msokoloff1: yeah this is definitely a problematic boolean param because we will want to get rid of it at some point. I'll see if there is a reliable way to auto detect the type.

labelbox/schema/dataset.py
msokoloff1: Looking into it.

msokoloff1: fixed.

labelbox/schema/dataset.py
msokoloff1: Removed.

labelbox/schema/dataset.py
```diff
@@ -135,7 +145,8 @@ def convert_item(item):
             }
 
         # Prepare and upload the desciptor file
-        data = json.dumps([convert_item(item) for item in items])
+        items = [convert_item(item) for item in items]
```
nmaswood, line 148: You can prob change this back

nmaswood (approved):

Issue #76, opened 2020-10-14
LabellingParameterOverrides is a dead end


vpipkt:
You can get a `LabellingParameterOverride` object from the `Project` method but it is not practically useful. The `uid` on this object is the identifier of the LabellingParameterOverride object itself which is fair enough. But there is no way to get the datarow relationship or the uid of that datarow.

So in practice it is not really useful because the API for setting overrides takes tuples of (`DataRow`, priority, numberOfLabels). If the   `LabellingParameterOverride`  has a relationship to its datarow you could write a program like this:

```
new_lpos = []

for lpo in proj.labeling_parameter_overrides():
    datarow = lpo.datarow() # function not defined
    new_lpos = (
        datarow,
        number_of_labels=lpo.number_of_labels + 1,
        priority=lpo.priority-10,
    )

proj. set_labeling_parameter_overrides(new_lpos)
```

Needless to say right now it is actually impossible to do something like this without going through the GraphQL API, or hacking away with a PaginatedCollection.

msokoloff1:
Adding this now `datarow = lpo.datarow()`

msokoloff1:
This has been merged to develop. If you need this functionality you can install from develop or wait until we release the next version (usually within a week).

PR #141, opened 2021-04-22
Ms/cached relationships

msokoloff1:
This PR addresses the issue of expensive queries for 1:1 relationships.

In the issue below a user wants to be able to access a data row for a given lpo. They can export all lpos as a paginated collection `project.labeling_parameter_overrides()` but then have to individually fetch each dataRow related to the lpo. 
https://github.com/Labelbox/labelbox-python/issues/76. This is a problem that keeps popping up. MLSEs often want a ToOne relationship for data that can otherwise be bulk exported (or at least exported via pagination).

To solve this, we can update the queries to fetch `ToOne` relationships when querying for `Field`s. Rough support for this was added in a previous PR. That PR enabled custom queries to pre-populate cached fields.

In this PR we have extended this functionality to leverage the relationship manager to automatically generate queries for cached relationships.

-- 
Also a one line fix for webhook deletion snuck into this pr.


nidoran (approved):

labelbox/orm/model.py
```diff
@@ -239,11 +239,58 @@ class EntityMeta(type):
     of the Entity class object so they can be referenced for example like:
         Entity.Project.
     """
+    mappings: Dict[str, "Entity"] = {}
```
gszpak, line 242: nit: this code is pretty complicated. Could we rename `mappings` so that it tells more of what it's used for?

labelbox/orm/model.py
```diff
@@ -239,11 +239,58 @@ class EntityMeta(type):
...
+    @staticmethod
+    def raise_for_nested_cache(first: str, middle: str, last: str):
+        raise TypeError(
+            "Cannot cache a relationship to an Entity with its own cached relationship(s). "
+            f"`{first}` caches `{middle}` which caches `{last}`")
+
+    def validate_cached_relationships(cls):
+        """
+        Graphql doesn't allow for infinite nesting in queries. 
+        This function checks that cached relationships result in valid queries.
```
gszpak, line 261: Forgive my ignorance (I should have context after reviewing previous PR), but why would caching affect the query in such a way that it would get infinite / invalid?

labelbox/orm/model.py
```diff
@@ -239,11 +239,58 @@ class EntityMeta(type):
...
+            * It does this by making sure that a cached relationship do not
+              reference any entity with its own cached relationships.
+
+        This check is performed by looking to see if this entity caches 
+        any entities that have their own cached fields. If this entity 
+        that we are checking has any cached fields then we also check 
+        all currently defined entities to see if they cache this entity.
+
+        A two way check is necessary because checks are performed as classes are being defined.
+        As opposed to after all objects have been created.
```
gszpak, line 271: I'll continue reviewing tomorrow, but at a first glance: isn't this code equivalent to detecting a cycle in relationship graph? If so, doing a two-way check doesn't seem enough?

labelbox/orm/model.py
msokoloff1: I don't think infinite ToOne relationships work. It was causing problems for me. But this keeps the logic much simpler. If we want to expand this in the future we can.

labelbox/orm/model.py
msokoloff1: This should be sufficient because the relationship that we are checking is only between two Entities and there are only three conditions.

1. An object is defined that doesn't have any cache.
2. A parent is defined and caches a child that isn't defined
    * Later that child is defined and checks for if this parent caches it and it has its own cached children. If so, raise else pass.
3. A child is defined that caches an object.
    * Later a parent is defined and checks if this child caches another object. If so, raise else pass.

tests/test_entity_meta.py
```diff
@@ -0,0 +1,34 @@
...
+        test_entity_b = Relationship.ToOne("TestEntityB", cache=True)
+
+    with pytest.raises(TypeError) as exc_info:
+
+        class TestEntityB(DbObject):
+            another_entity = Relationship.ToOne("AnotherEntity", cache=True)
+
+    assert str(exc_info.value) == \
+        "Cannot cache a relationship to an Entity with its own cached relationship(s)." \
+        " `test_entity_a` caches `test_entity_b` which caches `['another_entity']`"
```
gszpak, line 19: More of a nit, but I wouldn't assert for specific error msg - it gets annoying if you have more tests and want to update the message. Also error msg is not a part of API

gszpak (approved):
LGTM!

tests/test_entity_meta.py
msokoloff1: Ok good thought! I'll update it.

PR #150, opened 2021-05-13
[DIAG-101] bulk export queued data rows

msokoloff1:
* Allow users to fetch all data rows for a given project with a single bulk operation.

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -160,6 +160,35 @@ class Project(DbObject, Updateable, Deletable):
         return PaginatedCollection(self.client, query_str, {id_param: self.uid},
                                    ["project", "labels"], Label)
 
+    def export_queued_data_rows(self, timeout_seconds=120):
+        """ Returns all data rows that are currently enqueued for this project.
+
+        Args:
+            timeout_seconds (float): Max waiting time, in seconds.
+        Returns:
+            URL of the data file with this DataRow information. If the server didn't
+            generate during the `timeout_seconds` period, None is returned.
+        """
+        id_param = "projectId"
+        query_str = """mutation GetQueuedDataRowsExportUrlPyApi($%s: ID!)
+            {exportQueuedDataRows(data:{projectId: $%s }) {downloadUrl createdAt status} }
+        """ % (id_param, id_param)
+        sleep_time = 2
+        while True:
+            res = self.client.execute(query_str, {id_param: self.uid})
+            res = res["exportQueuedDataRows"]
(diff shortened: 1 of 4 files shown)


labelbox/schema/project.py
```diff
@@ -160,6 +160,35 @@ def labels(self, datasets=None, order_by=None):
...
+                return res["downloadUrl"]
+
+            timeout_seconds -= sleep_time
+            if timeout_seconds <= 0:
+                return None
+
+            logger.debug(
+                "Project '%s' queued data row export, waiting for server...",
+                self.uid)
+            time.sleep(sleep_time)
```
gszpak, line 190: nit: it's tempting to use `backoff.on_predicate` here (https://pypi.org/project/backoff/) not to reinvent polling logic, but might be not worth it - I'll leave it to you:)

labelbox/schema/project.py
```diff
@@ -160,6 +160,35 @@ def labels(self, datasets=None, order_by=None):
...
+        id_param = "projectId"
+        query_str = """mutation GetQueuedDataRowsExportUrlPyApi($%s: ID!)
+            {exportQueuedDataRows(data:{projectId: $%s }) {downloadUrl createdAt status} }
+        """ % (id_param, id_param)
+        sleep_time = 2
+        while True:
+            res = self.client.execute(query_str, {id_param: self.uid})
+            res = res["exportQueuedDataRows"]
+            if res["status"] == "COMPLETE":
+                return res["downloadUrl"]
```
gszpak, line 181: Should we raise an exception if status == FAILED?

labelbox/schema/project.py
```diff
@@ -160,6 +160,35 @@ def labels(self, datasets=None, order_by=None):
...
+
+            timeout_seconds -= sleep_time
+            if timeout_seconds <= 0:
+                return None
```
gszpak, line 185: Same here - maybe an exc?

labelbox/schema/project.py
msokoloff1: Good suggestion but I'm not going to add this. The user should be able to adjust the export time since a massive export could potentially take over 2 minutes.

labelbox/schema/project.py
msokoloff1: Yup. Added.

labelbox/schema/project.py
msokoloff1: Added

gszpak (approved):

PR #156, opened 2021-05-17
download exported data rows

msokoloff1:


labelbox/schema/project.py
```diff
@@ -178,7 +181,9 @@ def export_queued_data_rows(self, timeout_seconds=120):
             res = self.client.execute(query_str, {id_param: self.uid})
             res = res["exportQueuedDataRows"]
             if res["status"] == "COMPLETE":
-                return res["downloadUrl"]
+                download_url = res["downloadUrl"]
+                data = requests.get(download_url)
+                return ndjson.loads(data.text)
```
gdj0nes, line 186: probably could just be a single line

msokoloff1: Made it multi-line so the error would be more clear. Added raise_for_status() so that it actually serves that purpose..

nidoran (approved):

PR #157, opened 2021-05-19
NDTextEntity: fix "start >= end" test

ccrndn:
LabelBox convention about text location is that "end" is the index of the *last* character of the substring (that is not the python convention).
I.e.: `substring = string[v["start"] : v["end"] + 1]`

So, `v["start"] == v["end"]` is a valid case corresponding to a 1-character substring.


msokoloff1:
Thanks for the fix!

msokoloff1 (approved):

PR #161, opened 2021-06-03
[DIAG-196] model-run entity

msokoloff1:
https://labelbox.atlassian.net/browse/DIAG-196

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -125,7 +125,6 @@ class Client:
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
-            print(data)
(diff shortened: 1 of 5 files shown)


labelbox/client.py
```diff
@@ -125,6 +125,7 @@ def convert_value(value):
         data = json.dumps({'query': query, 'variables': params}).encode('utf-8')
 
         try:
+            print(data)
```
msokoloff1, line 128: oops. Will remove...

GrantEating, line 128: ✂️

GrantEating (approved):

GrantEating:
Remove that print first, please (unless it is intentional?)

PR #163, opened 2021-06-08
annotation import

msokoloff1:


labelbox/__init__.py
```diff
@@ -2,7 +2,9 @@
 __version__ = "2.5.6"
 
 from labelbox.client import Client
+from labelbox.schema.model import Model
```
awu-labelbox, line 5: This is already on Line 24

labelbox/schema/annotation_import.py
```diff
@@ -0,0 +1,295 @@
...
+            time.sleep(sleep_time_seconds)
+            self.__exponential_backoff_refresh()
+
+    @backoff.on_exception(
+        backoff.expo,
+        (labelbox.exceptions.ApiLimitError, labelbox.exceptions.TimeoutError,
+         labelbox.exceptions.NetworkError),
+        max_tries=10,
+        jitter=None)
+    def __exponential_backoff_refresh(self) -> None:
```
awu-labelbox, line 113: Maybe just call this `__backoff_refresh`

awu-labelbox (approved):
Lgtm, I only saw non-functionality related comments

labelbox/client.py
```diff
@@ -117,19 +126,35 @@ def convert_value(value):
...
-
+        if query is not None:
+            if params is not None:
+                params = {
+                    key: convert_value(value) for key, value in params.items()
+                }
+            data = json.dumps({
+                'query': query,
+                'variables': params
+            }).encode('utf-8')
```
cssetian, line 137: If `params` is `None`, would it set the value of `variables` to `undefined`? Is that OK here?

labelbox/client.py
```diff
@@ -117,19 +126,35 @@ def convert_value(value):
...
+        elif data is None:
+            raise ValueError("Params and data cannot both be none")
```
cssetian, line 139: Just reading through this for my own understanding, should this read `Query and data cannot both be none`? Though I suppose if `params` is defined, `query` should be too right? Or am I misreading?

labelbox/client.py
msokoloff1: Yeah this should be fine. Here is a minimal example you can try that works without setting params.
```
from labelbox import Client
client = Client()
client.execute(""" query { organization {id} }""")
```

labelbox/client.py
msokoloff1: This is wrong. Nice catch!

awu-labelbox (approved):
Lgtm -- just a reminder to remove/stub out local `API_KEY` and `ENDPOINT` in the notebook

labelbox/schema/annotation_import.py
```diff
@@ -0,0 +1,294 @@
...
+        Returns:
+            ndjson as a list of dicts.
+        """
+        response = requests.get(url)
+        response.raise_for_status()
+        return ndjson.loads(response.text)
+
+    @classmethod
+    def _build_import_predictions_query(cls, file_args: str, vars: str):
+        query_str = """mutation testPredictionImportsPyApi($parent_id : ID!, $name: String!, $predictionType : PredictionType!, %s) {
```
GrantEating, line 132: should this be called `testPredictionImportsPyPyi` still? Maybe `ModelErrorAnalysisPredictionImport`?

labelbox/schema/annotation_import.py
```diff
@@ -0,0 +1,294 @@
...
+                                     parent_id=model_run_id,
+                                     name=name,
+                                     path=path)
+
+    @classmethod
+    def create_from_objects(cls, client, model_run_id, name, predictions):
+        return cls._create_from_objects(client, model_run_id, name, predictions)
+
+
+class MALPredictionImport(AnnotationImport):
```
GrantEating, line 277: Im curious why we need this? I dont think were going to be allowing MAL imports via SDK through this mutation for a little while.

labelbox/schema/enums.py
```diff
@@ -4,19 +4,34 @@
 class BulkImportRequestState(Enum):
     """ State of the import job when importing annotations (RUNNING, FAILED, or FINISHED).
 
+    This object is deprecated. Please use PredictionImportState instead.
```
GrantEating, line 7: I think you mean `AnnotationImportState`.

PR #166, opened 2021-06-10
mea metrics

msokoloff1:
* Add code for datarow level iou scores


labelbox/data/metrics/iou.py
```diff
@@ -0,0 +1,290 @@
...
+    ious = [
+        feature_miou(predictions[feature_schema],
+                     labels[feature_schema],
+                     include_subclasses=include_subclasses)
+        for feature_schema in feature_schemas
+    ]
+    ious = [iou for iou in ious if iou is not None]
+    if not ious:
+        return None
+    return 100 * np.mean(ious)
```
msokoloff1, line 272: I wasn't sure where to put this `* 100`. I kind of feel like this should just be a float that is bound between 0 and 1. Any thoughts?

msokoloff1: fyi I reverted this back ( remove the 100) but I still want to discuss.

labelbox/data/metrics/preprocess.py
```diff
@@ -0,0 +1,72 @@
...
+    data = defaultdict(list)
+    for row in rows:
+        data[row.schemaId].append(row)
+    return data
+
+
+@retry.Retry(deadline=10.)
+def url_to_numpy(mask_url: str) -> np.ndarray:
+    """ Downloads an image and converts to a numpy array """
+    return np.array(Image.open(BytesIO(requests.get(mask_url).content)))
```
awu-labelbox, line 72: Are these resources closed automatically?

labelbox/data/metrics/iou.py
```diff
@@ -0,0 +1,290 @@
...
+from labelbox.data.metrics.preprocess import (create_schema_lookup,
+                                              url_to_numpy)
+
+VectorTool = Union[NDPoint, NDRectangle, NDPolyline, NDPolygon]
+ClassificationTool = Union[NDText, NDRadio, NDChecklist]
+
+
+def mask_iou(predictions: List[NDMask], labels: List[NDMask]) -> float:
+    """
+    Creates prediction and label binary mask for all features with the same feature scheama id.
```
awu-labelbox, line 22: Does this method check for same feature schema id's, or is that something assumed from the input?

labelbox/data/metrics/iou.py
```diff
@@ -0,0 +1,290 @@
...
+    classification_iou = [
+        feature_miou(subclass_predictions[feature_schema],
+                     subclass_labels[feature_schema])
+        for feature_schema in feature_schemas
+    ]
+    classification_iou = [x for x in classification_iou if x is not None]
+    return None if not len(classification_iou) else np.mean(classification_iou)
+
+
+def get_vector_pairs(predictions: List[Dict[str, Any]], labels):
```
awu-labelbox, line 107: Maybe prefix this function name with underscore (and other functions in this file that are internal only, like `preprocess_args`)

labelbox/data/metrics/iou.py
```diff
@@ -0,0 +1,290 @@
...
+    """
+    # Get iou score for all pairs of labels and predictions
+    """
+    return [(prediction, label,
+             _polygon_iou(prediction.to_shapely_poly(),
+                          label.to_shapely_poly()))
+            for prediction, label in product(predictions, labels)]
+
+
+def vector_iou(predictions: List[VectorTool], labels: List[VectorTool],
```
awu-labelbox, line 117: Change name to `vector_miou` since it's a mean?

labelbox/data/metrics/iou.py
```diff
@@ -0,0 +1,290 @@
...
+                 ndjsons: List[Dict[str, Any]],
+                 include_classifications=True,
+                 include_subclasses=True) -> float:
+    """
+
+    Args:
+        label_content : one row from the bulk label export - `project.export_labels()`
+        ndjsons: Model predictions in the ndjson format specified here (https://docs.labelbox.com/data-model/en/index-en#annotations)
+        include_classifications: Whether or not to factor top level classifications into the iou score.
+        include_subclassifications: Whether or not to factor in subclassifications into the iou score
```
awu-labelbox, line 254: Are `include_classifications`, `include_subclassifications` independent, or should there be some validation on their relation?

labelbox/data/metrics/iou.py
awu-labelbox: I feel better for some reason with values like this as within [0, 1].
But whatever is typical in this domain would also be a good way to decide.

labelbox/data/metrics/iou.py
```diff
@@ -0,0 +1,290 @@
...
+    all_features = set()
+    for pred, label, agreement in pairs:
+        all_features.update({pred.uuid, label.uuid})
+        if pred.uuid not in solution_features and label.uuid not in solution_features:
+            solution_features.update({pred.uuid, label.uuid})
+            if include_subclasses:
+                classification_iou = subclassification_iou(
+                    pred.classifications, label.classifications)
+                classification_iou = classification_iou if classification_iou is not None else agreement
+                # Weighted average of prediction and agreement.
```
msokoloff1, line 143: I just noticed that this comment is wrong.. will fix

labelbox/data/metrics/iou.py
msokoloff1: They are independent.

labelbox/data/metrics/iou.py
msokoloff1: ✔️

labelbox/data/metrics/iou.py
msokoloff1: It is not checked here. Technically it could accept any ids. because it is just calculating the iou between any set of masks.

labelbox/data/metrics/preprocess.py
msokoloff1: yup!

labelbox/data/metrics/iou.py
msokoloff1: I think 0 to 1 makes the most sense

labelbox/data/metrics/iou.py
```diff
@@ -0,0 +1,286 @@
...
+
+    """
+
+    if len(predictions) != len(labels) != 1:
+        return 0.
+
+    prediction, label = predictions[0], labels[0]
+
+    if type(prediction) != type(label):
+        raise Type
```
awu-labelbox, line 58: This works?

awu-labelbox (approved):
Just one comment, but otherwise lgtm!

labelbox/data/metrics/iou.py
msokoloff1: nope! that is bug haha

msokoloff1:
Please don't delete this branch! There is testing code floating around that relies on it.

PR #167, opened 2021-06-11
twine upload mea

msokoloff1:


msokoloff1:
Modeling this like tensorflow
https://pypi.org/project/tensorflow-gpu/
https://pypi.org/project/tensorflow/

msokoloff1:
My understanding is that the package name in pypi is based off the name of the package name in the wheel. I am fairly certain this is true but please double check my understanding because it will likely cause problems if that is not the case :).

gszpak (approved):

PR #179, opened 2021-06-21
bulk export datasets

msokoloff1:


labelbox/schema/dataset.py
```diff
@@ -227,3 +231,41 @@ def data_row_for_external_id(self, external_id):
                 f"More than one data_row has the provided external_id : `%s`. Use function data_rows_for_external_id to fetch all",
                 external_id)
         return data_rows[0]
+
+    def export_data_rows(self, timeout_seconds=120):
+        """ Returns a generator that produces all data rows that are currently attached to this datarow.
```
awu-labelbox, line 236: "attached to this *dataset*"

labelbox/schema/dataset.py
```diff
@@ -227,3 +231,41 @@ def data_row_for_external_id(self, external_id):
...
+        """ % (id_param, id_param)
+        sleep_time = 2
+        while True:
+            res = self.client.execute(query_str, {id_param: self.uid})
+            res = res["exportDatasetDataRows"]
+            if res["status"] == "COMPLETE":
+                download_url = res["downloadUrl"]
+                response = requests.get(download_url)
+                response.raise_for_status()
+                reader = ndjson.reader(StringIO(response.text))
```
awu-labelbox, line 257: What's the memory usage going to be like for large response sizes?

labelbox/schema/dataset.py
```diff
@@ -227,3 +231,41 @@ def data_row_for_external_id(self, external_id):
...
+
+        Args:
+            timeout_seconds (float): Max waiting time, in seconds.
+        Returns:
+            Generator that yields DataRow objects belonging to this dataset.
+        Raises:
+            LabelboxError: if the export fails or is unable to download within the specified time.
+        """
+        id_param = "datasetId"
+        query_str = """mutation GetQueuedDataRowsExportUrlPyApi($%s: ID!)
```
awu-labelbox, line 246: Does this return only queued (for labeling) data rows?

labelbox/schema/dataset.py
msokoloff1: 500k datarows was only in the 30mb range. So it should be fine.

labelbox/schema/dataset.py
msokoloff1: I named it incorrectly. Will update.

msokoloff1: Just to be clear, this will return all data rows for a the dataset.

awu-labelbox (approved):

Issue #191, opened 2021-06-30
can't instantiate derived class object when using classmethod


vikrant-kamble:
Due to the way the _classmethod_ `from_dict` is defined for class _Classification_ `labelbox.schema.ontology`, it is not possible to use this classmethod in a derived class (that derives from Classification) to obtain an instance of the derived class.

```
    @classmethod
    def from_dict(cls, dictionary: Dict[str, Any]):
        return Classification(                                              #  This needs to be changed from Classification to cls
            class_type=Classification.Type(dictionary["type"]),
            instructions=dictionary["instructions"],
            required=dictionary.get("required", False),
            options=[Option.from_dict(o) for o in dictionary["options"]],
            schema_id=dictionary.get("schemaNodeId", None),
            feature_schema_id=dictionary.get("featureSchemaId", None))
```

Not sure though, if this is by design.

msokoloff1:
Thanks for identifying this. I'll make a PR to fix this.

msokoloff1:
This has been merged and will go out when we do the next release.

PR #197, opened 2021-07-12
support extending ontology objects

msokoloff1:
Fixes the problems outlined here: https://github.com/Labelbox/labelbox-python/issues/191


gszpak (approved):
LGTM!

PR #201, opened 2021-07-21
annotation types

msokoloff1:


labelbox/data/generator.py
```diff
@@ -0,0 +1,84 @@
...
+        finally:
+            self.queue.put(None)
+
+    def __iter__(self):
+        return self
+
+    def __next__(self) -> Any:
+        value = self.queue.get()
+        while value is None:
+            self.completed_threads += 1
```
awu-labelbox, line 78: Would the standard thread pool library (https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor) be better than managing threads manually?

labelbox/data/annotation_types/collection.py
```diff
@@ -0,0 +1,273 @@
...
+            signer: A function that accepts bytes and returns a signed url.
+        Returns:
+            LabelGenerator that signs urls as data is accessed
+        """
+
+        def _add_url_to_data(label: Label):
+            label.add_url_to_data(signer)
+            return label
+
+        self._fns['_add_url_to_data'] = _add_url_to_data
```
awu-labelbox, line 194: nit: change key name by removing leading underscore

labelbox/data/annotation_types/collection.py
```diff
@@ -0,0 +1,273 @@
...
+
+        Args:
+            dataset: labelbox dataset object to add the new data row to
+            signer: A function that accepts bytes and returns a signed url.
+        Returns:
+            LabelGenerator that updates references to the new data rows as data is accessed
+        """
+
+        def _add_to_dataset(label: Label):
+            label.create_data_row(dataset, signer)
```
awu-labelbox, line 213: I was wondering how the generator version looks so much simpler than the collection version?

msokoloff1: It uses the functions provided by the Label class because it operates on each row individually. The Collection class tries to run bulk operations. They are still fairly similar in terms of lines of code.

labelbox/data/generator.py
msokoloff1: I updated this to use the threadpool manager instead of creating threads. But we can't use future.done() to know if there are any items left for the queue. The reason is that even if the futures are done, there might be a baglog of items to pull off the queue. So the logic is still pretty similar.

labelbox/data/annotation_types/collection.py
awu-labelbox: Got it, thanks

labelbox/data/annotation_types/annotation.py
```diff
@@ -0,0 +1,47 @@
...
+
+class BaseAnnotation(FeatureSchema):
+    """ Base annotation class. Shouldn't be directly instantiated
+    """
+
+    extra: Dict[str, Any] = {}
+
+
+class ClassificationAnnotation(BaseAnnotation):
+    """Class represneting classification annotations (annotations that don't have a location) """
```
awu-labelbox, line 17:
```suggestion
    """Class representing classification annotations (annotations that don't have a location) """
```

labelbox/data/annotation_types/data/video.py
```diff
@@ -0,0 +1,159 @@
...
+            return self.url
+        elif self.file_path is not None:
+            with open(self.file_path, 'rb') as file:
+                self.url = signer(file.read())
+        elif self.frames is not None:
+            self.file_path = self.frames_to_video(self.frames)
+            self.url = self.create_url(signer)
+        else:
+            raise ValueError(
+                "One of url, im_bytes, file_path, numpy must not be None.")
```
awu-labelbox, line 120:
```suggestion
                "One of url, file_path, frames must not be None.")
```

awu-labelbox (approved):

PR #202, opened 2021-07-21
lbv1-converter

msokoloff1:
* This code converts between the labelbox annotation objects and labelbox exports. Users should only interact with the `LBV1Converter` class.
* Add tests to make sure this behaves as expected for text, video, and image exports


labelbox/data/serialization/labelbox_v1/converter.py
```diff
@@ -0,0 +1,93 @@
...
+
+class LBV1Converter:
+
+    @staticmethod
+    def deserialize_video(json_data: Iterable[Dict[str, Any]], client):
+        """
+        Converts a labelbox video export into the common labelbox format.
+
+        Args:
+            json_data: An iterable representing the labelbox video export.
```
awu-labelbox, line 25: Add a note about the `client`

labelbox/data/serialization/labelbox_v1/converter.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __init__(self, examples, client):
+        self.client = client
+        super().__init__(examples)
+
+    def _process(self, value):
+        if 'frames' in value['Label']:
+            req = self._request(value)
+            value['Label'] = ndjson.loads(req)
+            return value
```
awu-labelbox, line 83: Raise a custom error message if `frames` isn't found

labelbox/data/serialization/labelbox_v1/converter.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+
+    def __init__(self, examples, client):
+        self.client = client
+        super().__init__(examples)
+
+    def _process(self, value):
+        if 'frames' in value['Label']:
+            req = self._request(value)
+            value['Label'] = ndjson.loads(req)
```
awu-labelbox, line 82: Is it necessary to mutate the original value, won't results be accessed via `__next__()`?

labelbox/data/serialization/labelbox_v1/label.py
```diff
@@ -0,0 +1,236 @@
...
+                                              'feature_id': cls.feature_id,
+                                              'title': cls.title,
+                                              'value': cls.value,
+                                          }) for cls in obj.classifications
+                                  ],
+                                  name=obj.title,
+                                  frame=self.frame_number,
+                                  alternative_name=obj.value,
+                                  schema_id=obj.schema_id,
+                                  extra={
```
awu-labelbox, line 67: I'm wondering in general how to decide what should be a class attribute and what should be in the extra dict

labelbox/data/serialization/labelbox_v1/label.py
```diff
@@ -0,0 +1,236 @@
...
+                return TextData(url=self.row_data, **data_row_info)
+            elif not self._is_url():
+                return TextData(text=self.row_data, **data_row_info)
+            else:
+                # This is going to be urls that do not contain any file extensions
+                # This will only occur on skipped images.
+                # To use this converter on data with this url format
+                #   filter out empty examples from the payload before deserializing.
+                raise TypeError(
+                    "Can't infer data type from row data. Remove empty examples before trying again"
```
awu-labelbox, line 217: Might be helpful to also add something identifying, like `row_data`

labelbox/data/serialization/labelbox_v1/objects.py
```diff
@@ -0,0 +1,245 @@
...
+
+    @classmethod
+    def from_common(cls, mask: Mask,
+                    classifications: List[ClassificationAnnotation],
+                    schema_id: Cuid, title: str,
+                    extra: Dict[str, Any]) -> "LBV1Mask":
+
+        if mask.mask.url is None:
+            raise ValueError(
+                "Mask does not have a url. Use `LabelGenerator.add_url_to_masks`, `LabelCollection.add_url_to_masks`, or `Label.add_url_to_masks`."
```
awu-labelbox, line 136:
```suggestion
                "Mask does not have a url. Use `LabelGenerator.add_url_to_masks`, `LabelList.add_url_to_masks`, or `Label.add_url_to_masks`."
```

labelbox/data/serialization/labelbox_v1/label.py
msokoloff1: Things that aren't necessary for modeling but we want to store for compatibility reasons. E.g. color, instanceURI, feature_id, value, all are extra information that isn't necessarily going to be used. That being said, this should be updated to make those fields optional. If that data doesn't exist it shouldn't break the serializer.

labelbox/data/serialization/labelbox_v1/converter.py
msokoloff1: It actually is fine if frames isn't found. That means that the data row isn't a video data row (which for some reason it is possible to mix data types...). The deserializer will handle any case.

labelbox/data/serialization/labelbox_v1/converter.py
msokoloff1: We are actually returning the mutated payload:
https://github.com/Labelbox/labelbox-python/blob/6a71b2ab9f13536701b14b7472dce83d67e193f8/labelbox/data/serialization/labelbox_v1/converter.py#L82-L83. This generator is only updating part of the data.

labelbox/data/serialization/labelbox_v1/converter.py
msokoloff1: I updated this to return value no matter what.

labelbox/data/serialization/labelbox_v1/label.py
msokoloff1: The extra fields are optional when constructing the lbv1 objects and when converting to annotation types these fields will all just be None.

labelbox/data/serialization/labelbox_v1/converter.py
awu-labelbox: The concern was that the _process API uses return of `None` as a signal 
https://github.com/Labelbox/labelbox-python/blob/develop/labelbox/data/generator.py#L61

awu-labelbox (approved):
Nice work!

PR #205, opened 2021-07-25
annotation examples

msokoloff1:
* Add examples for using annotation types
* Made changes to core labelbox to use annotation types by default on label export.


labelbox/schema/ontology.py
```diff
@@ -4,11 +4,13 @@
 
 from typing import Any, Dict, List, Optional, Union
 
-from labelbox.schema.project import Project
+#from labelbox.schema.project import Project
 from labelbox.orm.db_object import DbObject
 from labelbox.orm.model import Field, Relationship
 from labelbox.exceptions import InconsistentOntologyException
 
+Project = "NONE"
```
awu-labelbox, line 12: Were these file's changes for debug?

labelbox/schema/project.py
```diff
@@ -166,7 +167,14 @@ def export_queued_data_rows(self, timeout_seconds=120):
                 self.uid)
             time.sleep(sleep_time)
 
-    def export_labels(self, timeout_seconds=60):
+    def export_labels(self):
+        json_data = self.export_labels_json()
+        if 'frames' in json_data[0]['Label']:
+            # Assumes all data rows in this project are video.
```
awu-labelbox, line 173: This seems like a stretch of an assumption. Maybe add a docstring that mentions this?

labelbox/schema/ontology.py
msokoloff1: Yeah this needs to be fixed. good catch

labelbox/schema/project.py
msokoloff1: Yeah I think I am going to create a separate function for exporting video.

awu-labelbox (approved):

PR #210, opened 2021-07-27
mea-annotation

msokoloff1:
* Updated MEA metric calculation to use annotation types
* Fixed bug where the lbv1 converter where it was missing a field for schema ids.
* Add retry to remote fetching for data types
* Removed cuid/uuid converter that is unnecessary 
* Updated MEA tests to use the correct format

-- Will update the notebook in another branch


labelbox/data/metrics/iou.py
```diff
@@ -1,281 +1,254 @@
...
+            f"Found `{type(prediction)}` and `{type(ground_truth)}`")
+
+    if isinstance(prediction.value, Text):
+        return text_iou(ground_truth.value, prediction.value)
+    elif isinstance(prediction.value, Radio):
+        return radio_iou(ground_truth.value, prediction.value)
+    elif isinstance(prediction.value, Checklist):
+        return checklist_iou(ground_truth.value, prediction.value)
+    else:
+        raise ValueError(f"Unexpected subclass. {prediction}")
```
awu-labelbox, line 185: So dropdown is not currently supported? (because of MAL, I'm guessing?)

msokoloff1: Yes

labelbox/data/metrics/iou.py
```diff
@@ -1,281 +1,254 @@
...
-
-def classification_miou(predictions: List[ClassificationTool],
-                        labels: List[ClassificationTool]) -> float:
+def get_iou_across_features(
+    ground_truths: List[Union[ObjectAnnotation, ClassificationAnnotation]],
+    predictions: List[Union[ObjectAnnotation, ClassificationAnnotation]]
+) -> Optional[float]:
     """
-    Computes iou for classification features.
+    Groups annotations by schema_id or name (which is available), calculates iou score and returns the mean across all features.
```
awu-labelbox, line 33: Would it be useful to return a mapping of features to per-feature iou score?

awu-labelbox (approved):

labelbox/data/metrics/iou.py
msokoloff1: I think in the future we can create a function for that and basically move most of this logic to that function. For now this make sense since the aggregation is used for both top level and subclasses.

PR #234, opened 2021-08-11
schema id -> feature schema id

msokoloff1:


gdj0nes (approved):

msokoloff1:
Waiting to merge until after adding in MaskData class. This merge will break the notebooks without a new release so I just want to do it all at once.

PR #238, opened 2021-08-11
creating a delete method for bulk import requests

jtsodapop:
I wasn't too sure how to set up a test in the test_bulk_import_request.py file since we have no returned object and I believe we can't see deleted bulk imports (this is what I saw in the API Explorer when testing anyway)


msokoloff1 (approved):
Nice job!!

PR #243, opened 2021-08-12
Toggle verify

gdj0nes:


msokoloff1:
Users can run `export CURL_CA_BUNDLE=""` if they want to turn off ssl verification.

PR #252, opened 2021-08-18
split annotation import

msokoloff1:
* The shared code in `AnnotationImport` was quite messy. I split these functions into `MEAPredictionImport` and `MALPredictionImport`. 
* Fixed the mypy issues that were causing the build to fail in the other branch.
* Added some comments so autodocs can pick that up. 
* Remove ABC because it was conflicting with EntityMeta and would be a nightmare to add.


labelbox/schema/model_run.py
```diff
@@ -50,10 +50,10 @@ def add_predictions(
         kwargs = dict(client=self.client, model_run_id=self.uid, name=name)
         if isinstance(predictions, str) or isinstance(predictions, Path):
             if os.path.exists(predictions):
-                return MEAPredictionImport.create_from_file(path=predictions,
-                                                            **kwargs)
+                return MEAPredictionImport.create_from_file(
+                    path=str(predictions), **kwargs)
             else:
-                return MEAPredictionImport.create_from_url(url=predictions,
+                return MEAPredictionImport.create_from_url(url=str(predictions),
```
GrantEating, line 56: hmmm did this work prior? why do we need this change?

GrantEating (approved):
gave it a pretty close look and dont have much to say! looks better for sure, appreciate the abstraction of these two concepts 👍

labelbox/schema/annotation_import.py
```diff
@@ -266,41 +252,142 @@ def from_name(cls, client: "labelbox.Client", model_run_id: str,
...
+            query_str = cls._get_url_mutation()
+            return cls(
+                client,
+                client.execute(
+                    query_str,
+                    params={
+                        "fileUrl": url,
+                        "projectId": project_id,
+                        'name': name
+                    })["createModelAssistedLabelingPredictionImport"])
```
GrantEating, line 374: so you dont need to pass in a gql query here?  I guess i just havent seen this syntax yet.

labelbox/schema/model_run.py
msokoloff1: It doesn't actually effect anything. It just silences mypy.

labelbox/schema/annotation_import.py
msokoloff1: The first arg is the query: ```query_str```

PR #255, opened 2021-08-18
Jtso/list bir

jtsodapop:
Project.bulk_import_requests()

Created the test under test_bulk_import_request instead of test_project

Also changed some of the code in test_delete() to use the new method


labelbox/schema/project.py
```diff
@@ -582,6 +582,25 @@ def enable_model_assisted_labeling(self, toggle: bool = True) -> bool:
         return res["project"]["showPredictionsToLabelers"][
             "showingPredictionsToLabelers"]
 
+    def bulk_import_requests(self):
+        """ passpasspass
```
msokoloff1, line 586: prob need to update this lol

labelbox/schema/project.py
```diff
@@ -582,6 +582,25 @@ def enable_model_assisted_labeling(self, toggle: bool = True) -> bool:
...
+        """
```
msokoloff1, line 587: Also does this endpoint return the most recent imports first? If so lets add that to the docstring.

labelbox/schema/project.py
jtsodapop: Updated! Also goes by oldest first

msokoloff1 (approved):
Nice!

PR #258, opened 2021-08-20
limit picsum requests

msokoloff1:
* Picsum was rate limiting us which caused our backend to throw errors. 
* For each set of tests that we run there is now only one request to picsum and it is made from the client.


msokoloff1:
Replaced with #259

PR #261, opened 2021-08-25
Change benchmark_id typing to str for LBV1Label

paulnoirel:
I changed the type for benchmark_id to prevent the following error from happening, following a call to `project.label_genetors()`

`"Unexpected exception while filling the queue. ValidationError(model='LBV1Label', errors=[{'loc': ('Benchmark ID',), 'msg': 'value is not a valid float', 'type': 'type_error.float'}])"`

The error was originally mentioned in ticket  https://labelbox.atlassian.net/browse/LS-797.


msokoloff1:
🙌🙌🙌🙌🙌🙌🙌🙌

msokoloff1 (approved):
Thanks for doing this.

PR #267, opened 2021-08-31
bug fix

msokoloff1:
* create data rows not working on amazon linux


msokoloff1:
Tested fix on both amazon linux and ubunut.

GrantEating (approved):

PR #270, opened 2021-09-02
sync create data rows

msokoloff1:


labelbox/schema/dataset.py
```diff
@@ -69,13 +70,111 @@ def create_data_row(self, **kwargs):
...
+        query_str = """mutation AppendRowsToDatasetSyncPyApi($%s: ID!, $%s: String!){
+            appendRowsToDatasetSync(data:{datasetId: $%s, jsonFileUrl: $%s}
+            ){dataset{id}}} """ % (dataset_param, url_param, dataset_param,
+                                   url_param)
+        self.client.execute(query_str, {
+            dataset_param: self.uid,
+            url_param: descriptor_url
+        })
+
     def create_data_rows(self, items):
```
tomislav-peharda, line 117: What in particular makes this fn async while the other one is sync?

msokoloff1: The mutation that it calls. For the sync endpoint, when the request completes it means that the upload was successful. The async endpoint returns a task object and the user can poll for the status.

tomislav-peharda: Thanks.

tomislav-peharda (approved):

PR #274, opened 2021-09-07
coco converter

msokoloff1:


labelbox/data/serialization/coco/instance_dataset.py
```diff
@@ -0,0 +1,183 @@
...
+    segmentation = []
+    if isinstance(annotation.value, Polygon):
+        for point in annotation.value.points:
+            segmentation.extend([point.x, point.y])
+    else:
+        box = annotation.value
+        segmentation.extend([
+            box.start.x, box.start.y, box.end.x, box.start.y, box.end.x,
+            box.end.y, box.start.x, box.end.y
+        ])
```
gszpak, line 53: q: does the orientation matter for coco?

labelbox/data/serialization/coco/instance_dataset.py
```diff
@@ -0,0 +1,183 @@
...
+    return COCOObjectAnnotation(id=annot_idx,
+                                image_id=image_id,
+                                category_id=category_id,
+                                segmentation=[segmentation],
+                                area=shapely.area,
+                                bbox=[xmin, ymin, xmax - xmin, ymax - ymin],
+                                iscrowd=0)
+
+
+def rle_to_common(class_annotations, class_name):
```
gszpak, line 64: What is `common` ? Could you add type hints? Otherwise it's hard to figure out what's going on here

labelbox/data/serialization/coco/instance_dataset.py
```diff
@@ -0,0 +1,183 @@
...
+    mask = rle_decoding(class_annotations.segmentation.counts,
+                        *class_annotations.segmentation.size[::-1])
+    return ObjectAnnotation(name=class_name,
```
gszpak, line 67: Not I was going to leave a comment "I don't see `name` in `ObjectAnnotation` but then realized that `BaseAnnotation` inherits from `FeatureSchema`. Imho it's quite confusing, it should be a composition rather than inheritance.

labelbox/data/serialization/coco/instance_dataset.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @classmethod
+    def from_common(cls, labels: LabelCollection, image_root):
+        all_coco_annotations = []
+        categories = {}
+        images = []
+        futures = []
+        coco_categories = {}
+
+        with ProcessPoolExecutor(max_workers=8) as exc:
```
gszpak, line 128: let's make executor an optional argument of `from_common`

labelbox/data/serialization/coco/panoptic_dataset.py
```diff
@@ -0,0 +1,178 @@
...
+from typing import Dict, Any, List, Union
+
+
+def vector_to_coco_segment_info(canvas: np.ndarray,
+                                annotation: ObjectAnnotation,
+                                annotation_idx: int, image: CocoImage,
+                                category_id: int):
+    shapely = annotation.value.shapely
+    xmin, ymin, xmax, ymax = shapely.bounds
+    canvas = annotation.value.draw(height=image.height,
```
gszpak, line 25: `value` seems to be of type `Geometry` - I don't see it having `draw` method?

labelbox/data/serialization/coco/panoptic_dataset.py
```diff
@@ -0,0 +1,178 @@
...
+    return SegmentInfo(id=annotation_idx,
+                       category_id=category_id,
+                       area=shapely.area,
+                       bbox=[xmin, ymin, xmax - xmin, ymax - ymin]), canvas
+
+
+def mask_to_coco_segment_info(canvas: np.ndarray, annotation,
+                              annotation_idx: int, category_id):
+    color = id_to_rgb(annotation_idx)
+    mask = annotation.value.draw(color=color)
```
gszpak, line 39: same comment about draw?

labelbox/data/serialization/coco/panoptic_dataset.py
gszpak, line 25: `value` seems to be of type `Geometry` - I don't see it having `draw` method?

labelbox/data/serialization/coco/panoptic_dataset.py
```diff
@@ -0,0 +1,178 @@
...
+    categories: List[Categories]
+
+    @classmethod
+    def from_common(cls, labels: LabelCollection, image_root, mask_root,
+                    all_stuff):
+        all_coco_annotations = []
+        coco_categories = {}
+        coco_things = {}
+        images = []
+        with ProcessPoolExecutor(max_workers=8) as exc:
```
gszpak, line 110: Same comment about executor as arg. Not being able to specify concurrency would be pretty annoying to me as user

labelbox/data/serialization/coco/panoptic_dataset.py
msokoloff1: Yeah it has a draw method:
https://github.com/Labelbox/labelbox-python/blob/7432b6615e35130a30c8ee8f549c96a97e4686b0/labelbox/data/annotation_types/geometry/geometry.py#L38

labelbox/data/serialization/coco/annotation.py
```diff
@@ -0,0 +1,55 @@
+from pydantic import BaseModel
+from collections import defaultdict
+from typing import Tuple, List, Union
+import numpy as np
+
+
+def rle_decoding(rle_arr, w, h):
```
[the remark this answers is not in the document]
msokoloff1: I don't want to add pycoco utils as a dependency.

labelbox/data/serialization/coco/panoptic_dataset.py
msokoloff1: What if I just let the number of workers be an arg? My thinking is that most users are going to want the same amount of parallelism - the max that the backend can handle. Also this script is unlikely to run within the context of another executor - users are going to either convert an entire dataset or not. So asking users to pass in the execturor is an extra step that will always be necessary.

labelbox/data/serialization/coco/instance_dataset.py
msokoloff1: See question below for same comment on panoptic dataset.

msokoloff1:
I changed everything to use Pathlib - including the public interfaces. I am a little bit hesitant to keep it that way because we don't support that in other interfaces. Let's chat about this.

labelbox/data/serialization/coco/converter.py
```diff
@@ -0,0 +1,116 @@
...
+        Returns:
+            A dictionary containing labels in the coco object format.
+        """
+        create_path_if_not_exists(image_root, ignore_existing_data)
+        return CocoInstanceDataset.from_common(labels=labels,
+                                               image_root=image_root).dict()
+
+    def serialize_panoptic(labels: LabelCollection,
+                           image_root: str,
+                           mask_root: str,
```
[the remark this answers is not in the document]
msokoloff1: done

labelbox/data/serialization/coco/converter.py
```diff
@@ -0,0 +1,116 @@
...
+    if not os.path.exists(path):
+        os.makedirs(path)
+    elif not ignore_existing_data and os.listdir(path):
+        raise ValueError(
+            f"Directory `{path}`` must be empty. Or set `ignore_existing_data=True`"
+        )
+
+
+def validate_path(path, name):
+    if not os.path.exists(path):
```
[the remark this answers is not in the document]
msokoloff1: done

labelbox/data/serialization/coco/converter.py
```diff
@@ -0,0 +1,116 @@
+from typing import Dict, Any
+import os
+
+from labelbox.data.annotation_types.collection import LabelCollection, LabelGenerator
+from labelbox.data.serialization.coco.instance_dataset import CocoInstanceDataset
+from labelbox.data.serialization.coco.panoptic_dataset import CocoPanopticDataset
+
+
+def create_path_if_not_exists(path: str, ignore_existing_data=False):
```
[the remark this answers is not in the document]
msokoloff1: done

gszpak (approved):

PR #277, opened 2021-09-08
conf matrix metric and confidence

msokoloff1:


labelbox/data/annotation_types/metrics/base.py
```diff
@@ -0,0 +1,35 @@
...
+from pydantic.error_wrappers import ErrorWrapper
+from typing import Dict, Optional, Any, Union
+
+ConfidenceValue = confloat(ge=0, le=1)
+
+MIN_CONFIDENCE_SCORES = 2
+MAX_CONFIDENCE_SCORES = 15
+
+
+class BaseMetric(BaseModel, ABC):
```
tomislav-peharda, line 12: `ABC`?

labelbox/data/annotation_types/metrics/base.py
```diff
@@ -0,0 +1,35 @@
+from abc import ABC
+from pydantic import ValidationError, confloat, BaseModel, validator
+from pydantic.error_wrappers import ErrorWrapper
+from typing import Dict, Optional, Any, Union
+
+ConfidenceValue = confloat(ge=0, le=1)
+
+MIN_CONFIDENCE_SCORES = 2
+MAX_CONFIDENCE_SCORES = 15
```
tomislav-peharda, line 9: how come `[2,15]` and not `[0,1]`?

tomislav-peharda (approved):
I'm giving it a stamp, but would be ideal if we could have someone else look at it too as I don't feel I'm good fit for this PR.

labelbox/data/annotation_types/metrics/base.py
msokoloff1: This is the number of scores. Not the range. We are constraining the range with pydantics confloat type. So there is no point of having confidence if you have less than one value. And if you have more than 15 it is too many details and clutters our db.

labelbox/data/annotation_types/metrics/base.py
msokoloff1: abstract base class. It prevents users from directly instantiating this object.

PR #279, opened 2021-09-10
iam integrations

msokoloff1:


labelbox/schema/iam_integration.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    _DEFAULT = "DEFAULT"
+
+    name = Field.String("name")
+    created_at = Field.DateTime("created_at")
+    updated_at = Field.DateTime("updated_at")
+    provider = Field.String("provider")
+    valid = Field.Boolean("valid")
+    last_valid_at = Field.DateTime("last_valid_at")
+    is_org_default = Field.Boolean("is_org_default")
```
gszpak, line 26: seems like we're missing `settings` field?

gszpak (approved):
Overall LGTM - one thing I'm not a huge fan of is the `create_dataset` flow. The logic to use the default signer (or be able to pass the signer) should be part of `createDataset` mutation and handled by the backend

labelbox/client.py
```diff
@@ -512,14 +514,51 @@ def create_dataset(self, **kwargs):
...
+                    setSignerForDataset(data: { signerId: $signerId}, where: {id: $datasetId}){id}}
+                """, {
+                    'signerId': iam_integration.uid,
+                    'datasetId': dataset.uid
+                })
+            validation_result = self.execute(
+                """mutation validateDatasetPyApi($id: ID!){validateDataset(where: {id : $id}){
+                    valid checks{name, success}}}
+                """, {'id': dataset.uid})
+            if not validation_result['validateDataset']['checks'][0]['success']:
```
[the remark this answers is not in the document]
msokoloff1: checks[0] is the dataset validation and it always is run. Checks[1:] check that each data row is valid and since there are no data rows we only need to check checks[0]. But also looking into this more, validation_result['validateDataset']['valid'] returns true only if all checks are true. So I'll use that instead.

labelbox/schema/iam_integration.py
msokoloff1: ugh - I know. The union thing is a nightmare. But users might want to know which bucket they have access to. I'll try to figure something out.

labelbox/schema/iam_integration.py
```diff
@@ -0,0 +1,31 @@
...
+        created_at (datetime)
+        provider (str)
+        valid (bool)
+        last_valid_at (datetime)
+        is_org_default (boolean)
+
+    """
+
+    def __init__(self, client, data):
+        self.settings = data.pop('settings', {})
```
gszpak, line 20: instead of keeping a dict, could we map it to a dataclass based on `__typename`? (Or to Entity - not sure how we represent "helper" gql types like eg `GcpIamIntegrationSettings`?)

PR #282, opened 2021-09-10
update error message

msokoloff1:


labelbox/client.py
```diff
@@ -554,7 +554,7 @@ def create_dataset(self, iam_integration=IAMIntegration._DEFAULT, **kwargs):
 
             if not validation_result['validateDataset']['valid']:
                 raise labelbox.exceptions.LabelboxError(
-                    f"IAMIntegration {validation_result['validateDataset']['checks']['name']} was not successfully added added to the project."
```
msokoloff1, line 557: this just says signer every time.

msokoloff1: and the typo :(

gdj0nes (approved):

PR #284, opened 2021-09-12
consistent drawing behavior

msokoloff1:
If an array is non-contiguous then opencv drawing functions return opencv matrix objects instead of numpy arrays. Adding np.ascontiguous array prevents this situation from occuring. It also doesn't do anything unless the array is non-contiguous so there isn't any extra overhead for adding this each time.


paulnoirel (approved):
looks fine

PR #285, opened 2021-09-13
confusion matrix metrics

msokoloff1:


labelbox/data/annotation_types/metrics/base.py
```diff
@@ -0,0 +1,35 @@
...
+    subclass_name: Optional[str] = None
+    extra: Dict[str, Any] = {}
+
+    def dict(self, *args, **kwargs):
+        res = super().dict(*args, **kwargs)
+        return {k: v for k, v in res.items() if v is not None}
+
+    @validator('value')
+    def validate_value(cls, value):
+        if isinstance(value, Dict):
```
GrantEating, line 24: should it throw an error if it isnt an instance of a `Dict`?

labelbox/data/annotation_types/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,35 @@
...
+    CONFUSION_MATRIX = "CONFUSION_MATRIX"
+
+
+class ConfusionMatrixMetric(BaseMetric):
+    """ Class representing confusion matrix metrics.
+
+    In the editor, this provides precision, recall, and f-scores.
+    This should be used over multiple scalar metrics so that aggregations are accurate.
+
+    value should be a tuple representing:
```
GrantEating, line 26: nit: capitolize `v` in `value`

labelbox/data/metrics/confusion_matrix/calculation.py
```diff
@@ -0,0 +1,288 @@
...
+                     predictions: List[Union[ObjectAnnotation,
+                                             ClassificationAnnotation]],
+                     include_subclasses: bool,
+                     iou: float) -> ConfusionMatrixMetricValue:
+    """
+    Computes the confusion matrix for an arbitrary set of ground truth and predicted annotations.
+    It first computes the confusion matrix for each metric and then sums across all classes
+
+    Args:
+        ground_truth : Label containing human annotations or annotations known to be correct
```
GrantEating, line 24: i think we avoided using this name in the label import project because it didnt exist anywhere else in the application. Can you double check that we want to bake this into a public-facing method? I think maybe we can just call them `label`?

labelbox/data/metrics/confusion_matrix/calculation.py
```diff
@@ -0,0 +1,288 @@
...
+        prediction: Label representing model predictions
+        include_subclasses (bool): Whether or not to include subclasses in the calculation.
+            If set to True, the iou between two overlapping objects of the same type is 0 if the subclasses are not the same.
```
GrantEating, line 27: Thats cool!

labelbox/data/metrics/confusion_matrix/calculation.py
```diff
@@ -0,0 +1,288 @@
...
+    Returns:
+        confusion matrix as a list: [TP,FP,TN,FN]
+        Returns None if there are no annotations in ground_truth or prediction annotations
+    """
+    if has_no_matching_annotations(ground_truths, predictions):
+        return [0, int(len(predictions) > 0), 0, int(len(ground_truths) > 0)]
+    elif has_no_annotations(ground_truths, predictions):
+        return None
+    elif isinstance(predictions[0].value, Mask):
```
GrantEating, line 65: why do we only check the first value?

labelbox/data/metrics/confusion_matrix/calculation.py
```diff
@@ -0,0 +1,288 @@
...
+
+    prediction, ground_truth = predictions[0], ground_truths[0]
+
+    if type(prediction) != type(ground_truth):
+        raise TypeError(
+            "Classification features must be the same type to compute agreement. "
+            f"Found `{type(prediction)}` and `{type(ground_truth)}`")
+
+    if isinstance(prediction.value, Radio):
+        return radio_confusion_matrix(ground_truth.value, prediction.value)
```
GrantEating, line 109: i dont follow why were returning the confusion matrices for just the first prediction/ground truth values?

labelbox/data/metrics/group.py
```diff
@@ -158,3 +157,19 @@ def _create_feature_lookup(features: List[FeatureSchema],
     for feature in features:
         grouped_features[getattr(feature, key)].append(feature)
     return grouped_features
+
+
+def has_no_matching_annotations(ground_truths: List[ObjectAnnotation],
```
GrantEating, line 162: what is a `matching` annotation? wouldnt a matching prediction and label have the same schemaId? Would it make sense to check for that in this function, or is it not that simple?
Would it make sense to make sure the lengths are the same?

tests/data/metrics/confusion_matrix/test_confusion_matrix_data_row.py
```diff
@@ -0,0 +1,57 @@
...
+        score = confusion_matrix_metric(example.ground_truths,
+                                        example.predictions)
+
+        if len(example.expected) == 0:
+            assert len(score) == 0
+        else:
+            expected = [0, 0, 0, 0]
+            for expected_values in example.expected.values():
+                for idx in range(4):
+                    expected[idx] += expected_values[idx]
```
GrantEating, line 25: isnt this whole block just re-generating a list, `expected`, which is the same list as `example.expected.values()`? 
Do we need this block at all? can we just do 

```
            assert score[0].value == tuple(
                example.expected.values()), f"{example.predictions},{example.ground_truths}"
```
?

tests/data/serialization/ndjson/test_metric.py
```diff
@@ -16,14 +16,29 @@ def test_metric():
...
+
+def test_custom_confusion_matrix_metric():
+    with open('tests/data/assets/ndjson/custom_confusion_matrix_import.json',
+              'r') as file:
+        data = json.load(file)
+
+    label_list = NDJsonConverter.deserialize(data).as_list()
+    reserialized = list(NDJsonConverter.serialize(label_list))
+    assert json.dumps(reserialized,
+                      sort_keys=True) == json.dumps(data, sort_keys=True)
```
GrantEating, line 41: how does this test `test_custom_confusion_matrix_metric` ? it seems like it just deserialize and re-serializes a json file, then asserts that its equal to the file data?

tests/integration/conftest.py
```diff
@@ -214,9 +212,11 @@ def datarow(dataset, image_url):
 @pytest.fixture
 def label_pack(project, rand_gen, image_url):
     client = project.client
-    dataset = client.create_dataset(name=rand_gen(str), projects=project)
-    data_row = dataset.create_data_row(row_data=image_url)
+    dataset = client.create_dataset(name=rand_gen(str))
+    project.datasets.connect(dataset)
+    data_row = dataset.create_data_row(row_data=IMG_URL)
     label = project.create_label(data_row=data_row, label=rand_gen(str))
+    time.sleep(10)
```
GrantEating, line 219: That seems like a lot of time. Is there any way to avoid this? I assume that things were not being created in time? is there a way to "await" (for lack of a pythonic verison of the word) these function calls?

labelbox/data/serialization/ndjson/metric.py
```diff
@@ -1,24 +1,69 @@
...
+        return res
+
+
+class NDConfusionMatrixMetric(BaseNDMetric):
+    metric_value: Union[ConfusionMatrixMetricValue,
+                        ConfusionMatrixMetricConfidenceValue]
+    metric_name: str
+    aggregation: ConfusionMatrixAggregation
+
+    def to_common(self) -> ConfusionMatrixMetric:
```
GrantEating, line 35: what is `common`?

labelbox/data/metrics/confusion_matrix/calculation.py
```diff
@@ -0,0 +1,288 @@
...
+    agree = schema_ids_label & schema_ids_pred
+    all_selected = schema_ids_label | schema_ids_pred
+    disagree = all_selected.difference(agree)
+    fps = len({x for x in disagree if x in schema_ids_pred})
+    fns = len({x for x in disagree if x in schema_ids_label})
+    tps = len(agree)
+    return [tps, fps, 0, fns]
+
+
+def mask_confusion_matrix(ground_truths: List[ObjectAnnotation],
```
GrantEating, line 244: I dont feel like i was able to review these functions at all, fyi

GrantEating:
As badly as a i want to say LGTM, i dont feel confident with my review here. I left some comments, but most arent very useful. 
Happy to stamp, but its more of a `LGTM?` haha. 

This PR is worth being proud of, though. Really shows an accumulation of a lot of knowledge and hard work. I am excited to see people use this, youve really owned custom metrics.

labelbox/data/annotation_types/metrics/base.py
msokoloff1: no because it is valid to not be a dict. Just if it is, then we want to validate the confidence ranges.

labelbox/data/metrics/confusion_matrix/calculation.py
msokoloff1: Great point, we need to come up with some consistent language. The problem here is that Label means "data and annotations" not "human made annotations". There needs to be some distinction between human generated and machine generated labels. I am open to changing the name but I already exposed this concept with the iou metric functions for our first mea release :). Although these are positional args so if we want to change this we probably can. We should probably do it soon though.

labelbox/data/metrics/confusion_matrix/calculation.py
msokoloff1: Because we have already checked that they are all the same type previously. So we just need to quickly check again what is contained in the list.

labelbox/data/metrics/confusion_matrix/calculation.py
msokoloff1: Because at this point we have grouped annotations by class. So there can only be one or zero of a classification in an image. I updated the docstring to explain this.

labelbox/data/metrics/group.py
msokoloff1: All annotations passed to this function have the same schema ids. So this function checks if we even need to run the metric calculation code - if there is no ground truth or no prediction then we just give a default score.

tests/data/metrics/confusion_matrix/test_confusion_matrix_data_row.py
msokoloff1: We are actually accumulating values across all expected_values.
```
expected = [0, 0, 0, 0]
for expected_values in example.expected.values():
    for idx in range(4):
        expected[idx] += expected_values[idx]
assert ... # Each of the values are added to expected before this assert statement is reached.
```

tests/data/serialization/ndjson/test_metric.py
msokoloff1: This tests the serialization and deserialization of the custom metric. Not the actual calculation.

tests/integration/conftest.py
msokoloff1: everything is technically awaited in python. The problem is the async stuff on the backend that can't be awaited.

msokoloff1: Also once label imports are complete, I can throw out this garbage hack to upload labels for tests.

labelbox/data/metrics/confusion_matrix/calculation.py
GrantEating: i honestly have no stake in it, i guess i just wanted to point it out. I think Gareth asked if we could change the name in label imports, so he might be a person to run it by. 
I definitely dont have a problem with it, and honestly, i think ground truth label is a good name, as thats the name in the industry, so i dont know why labelbox wouldnt go that direction.

tests/data/serialization/ndjson/test_metric.py
GrantEating: got it. I might recommend naming it something about the serialization, then but that just a nit!

tests/integration/conftest.py
GrantEating: ah, yeah. dang. 10s is a while to add to a test, but if the processing takes that long, makes sense!

GrantEating (approved):

PR #314, opened 2021-10-15
[DIAG-779] Update SDK model_run.upsert_labels() method

tomislav-peharda:


labelbox/schema/model_run.py
```diff
@@ -16,20 +17,40 @@ class ModelRun(DbObject):
...
 
-        query_str = """mutation upsertModelRunLabelsPyApi($modelRunId: ID!, $labelIds : [ID!]!) {
-          upsertModelRunLabels(where : { id : $modelRunId}, data : {labelIds: $labelIds})}
-          """
-        res = self.client.execute(query_str, {
-            'modelRunId': self.uid,
-            'labelIds': label_ids
-        })
-        # TODO: Return a task
-        return True
```
tomislav-peharda, line 32: this is prev logic. @msokoloff1 how should we approach converting from old approach to new one?

labelbox/schema/model_run.py
```diff
@@ -16,20 +17,40 @@ class ModelRun(DbObject):
...
+        sleep_time = 5
+
+        mutation_name = 'createMEAModelRunLabelRegistrationTask'
+        query_str = """mutation createMEAModelRunLabelRegistrationTaskByApi($modelRunId: ID!, $labelIds : [ID!]!) {
+          %s(where : { id : $modelRunId}, data : {labelIds: $labelIds})}
+          """ (mutation_name)
+
+        while True:
+            res = self.client.execute(query_str, {
```
msokoloff1, line 40: This is going to re-enqueue a new task each time. We should be polling for the status here.

PR #319, opened 2021-10-21
add OperationNotAllowedException to catch limits

msokoloff1:


labelbox/client.py
```diff
@@ -201,6 +201,7 @@ def check_errors(keywords, *path):
             return None
 
         def get_error_status_code(error):
+            print(error)
```
tomislav-peharda, line 204: was this intentional?

tomislav-peharda (approved):

PR #327, opened 2021-11-02
update to all tests that used annotation_submit_fn

jtsodapop:


tests/integration/conftest.py
```diff
@@ -355,10 +326,10 @@ def configured_project_with_label(client, rand_gen, annotation_submit_fn,
             "width": 50
         }
     }]
-    upload_task = MALPredictionImport.create_from_objects(
-        client, project.uid, f'mal-import-{uuid.uuid4()}', predictions)
+    upload_task = LabelImport.create_from_objects(
+        client, project.uid, f'label-import-{uuid.uuid4()}', predictions)
     upload_task.wait_until_done()
-    labels = annotation_submit_fn(project.uid, data_row.uid)
+    labels = [label.uid for label in project.labels()][0]
```
msokoloff1, line 332: why only the zeroth index?

tests/integration/conftest.py
```diff
@@ -355,10 +326,10 @@ def configured_project_with_label(client, rand_gen, annotation_submit_fn,
...
     time.sleep(3)
```
msokoloff1, line 333: You can scrap this sleep timer

tests/integration/conftest.py
jtsodapop: configured_project_with_label used to only return 1 id in the `labels` var, so the test that used it did this:

model_run.upsert_labels([label_id])

i can change the test itself to just take in (label_id) instead of ([label_id]) if that makes more sense, and to keep labels as the whole list

jtsodapop: actually looking at how the asserts are set up, it looks like it does expect it to just be one. let me know if i should change the asserts to account for multiple labels too, or if leaving it as the 0th index is okay

msokoloff1: ohh I was mislead by the variable name. labels is just a single label. Ok then you can just do:
```
label = next(project.labels()).uid
yield [project, label]
```
That way it won't confuse people in the future

msokoloff1 (approved):

PR #331, opened 2021-11-03
parallelize

jtsodapop:


jtsodapop:
image

The following changes to the .yml file will allow all three builds to be run at the same time.

jtsodapop:
looking at the error in the check itself, relates to some of the changes that should be alleviated in DIAG-888

msokoloff1 (approved):

PR #332, opened 2021-11-04
updates to tests after implementation of LabelImport

jtsodapop:
update to tests that were previously marked skipped before pending LabelImport. Is supposed to be the following step after annotation_submit_fn was removed from the files

### DIFF
--- a/tests/integration/test_label.py
+++ b/tests/integration/test_label.py
@@ -1,6 +1,5 @@
 from labelbox.schema.labeling_frontend import LabelingFrontend
 import time
-import json
 
 import pytest
 import requests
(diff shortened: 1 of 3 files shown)


tests/integration/test_label.py
```diff
@@ -30,20 +29,14 @@ def test_labels(label_pack):
...
-    project.create_label(data_row=data_row, label="export_label")
+def test_label_export(client, configured_project_with_label):
+    project, label_id = configured_project_with_label
+
     exported_labels_url = project.export_labels()
     assert exported_labels_url is not None
     exported_labels = requests.get(exported_labels_url)
-    labels = [example['Label'] for example in exported_labels.json()]
-    #assert 'export_label' in labels
+    labels = [example['ID'] for example in exported_labels.json()]
```
jtsodapop, line 38: since we aren't adding in "export_label" specifically in this test anymore, thought it might be more appropriate to test against the ID instead

msokoloff1 (approved):

PR #334, opened 2021-11-08
updates to documentation. also, Line should only be 2 or more Points

jtsodapop:


labelbox/data/annotation_types/geometry/line.py
```diff
@@ -47,3 +49,12 @@ def draw(self,
                              False,
                              color=color,
                              thickness=thickness)
+
+    @validator('points')
+    def is_geom_valid(cls, points):
+        if len(points) < 2:
```
jtsodapop, line 55: saw this was setup similarly for polygon, except it required 3 points. currently can create a "line" with only 1 point

msokoloff1 (approved):

PR #340, opened 2021-11-12
ontology management

msokoloff1:


labelbox/schema/project.py
```diff
@@ -422,6 +422,41 @@ def review_metrics(self, net_score):
...
+                LFO.project:
+                    self,
+                LFO.labeling_frontend:
+                    labeling_frontend,
+                LFO.customization_options:
+                    json.dumps({
+                        "tools": [],
+                        "classifications": []
+                    })
+            })
```
msokoloff1, line 449: There is no harm in adding this but it doesn't seem necessary. @nickroberts404 can I snip this?

tomislav-peharda (approved):

gdj0nes (changes requested):
Hold for naming feedback from @manuaero

labelbox/schema/project.py
```diff
@@ -422,6 +422,41 @@ def review_metrics(self, net_score):
...
+
+        query_str = """mutation ConnectOntologyPyApi($projectId: ID!, $ontologyId: ID!){
+            project(where: {id: $projectId}) {connectOntology(ontologyId: $ontologyId) {id}}}"""
+        self.client.execute(query_str, {
+            'ontologyId': ontology.uid,
+            'projectId': self.uid
+        })
+        timestamp = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+        self.update(setup_complete=timestamp)
```
gdj0nes, line 458: pretty weird this is on the client side

examples/project_configuration/project_setup.ipynb
```diff
@@ -94,86 +58,71 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "id": "limiting-bleeding",
    "metadata": {},
    "source": [
-    "### Setup and attach dataset\n",
+    "### Create Project and Setup\n",
     "* Setting up a project will add an ontology and will enable labeling to begin\n",
```
nickroberts404, line 112: "Setting up an editor" maybe?

labelbox/client.py
```diff
@@ -709,3 +707,189 @@ def get_data_row_ids_for_external_ids(
...
+            name (str): Name of the ontology
+            feature_schema_ids (List[str]): List of feature schema ids corresponding to
+                top level tools and classifications to include in the ontology
+        Returns:
+            The created Ontology
+        """
+        tools, classifications = [], []
+        for feature_schema_id in feature_schema_ids:
+            feature_schema = self.get_feature_schema(feature_schema_id)
+            tool = feature_schema.normalized['tool']
```
nickroberts404, line 807: I'm not good at python so i can't tell if this handles it, but `tool` will only be defined on a tool kind schema, for a classification the field is `type` (i know, its dumb)

msokoloff1: Nice catch!

labelbox/schema/project.py
nickroberts404: Hmmm i'd keep it for right now, I want to do an actual thorough cleansing of LFO's from Labelbox and that would be a better time to take this out, there might be some weird gotchas somehwhere or lesser known features that would break unless we really look around

nickroberts404 (approved):
lgtm, only real concern might be the `normalized[tool]` thing but I can't tell if i'm reading it wrong!

examples/project_configuration/project_setup.ipynb
msokoloff1: updated!

labelbox/client.py
msokoloff1: Fixed

gdj0nes (approved):

labelbox/client.py
```diff
@@ -17,17 +17,12 @@
...
-from labelbox.schema.dataset import Dataset
-from labelbox.schema.data_row import DataRow
-from labelbox.schema.model import Model
-from labelbox.schema.user import User
-from labelbox.schema.organization import Organization
 from labelbox.schema.data_row_metadata import DataRowMetadataOntology
-from labelbox.schema.labeling_frontend import LabelingFrontend
 from labelbox.schema.iam_integration import IAMIntegration
 from labelbox.schema import role
+from labelbox.schema.ontology import Tool, Classification
```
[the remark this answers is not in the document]
msokoloff1: No because these aren't entities.

labelbox/client.py
nickroberts404: thanks!

PR #345, opened 2021-11-17
video classification keyframe

msokoloff1:
- Keyframes are present in classification answers so we are supporting that
- There is a new optional field for media type in the labelbox v1 converter. This will be used properly in the future but users can set the media type manually now if it can't be inferred.


labelbox/data/serialization/labelbox_v1/label.py
```diff
@@ -174,44 +171,49 @@ def from_common(cls, label: Label):
...
-            # Use information from the row_data format if possible.
             if self._row_contains((".jpg", ".png", ".jpeg")) and self._is_url():
-                return ImageData(url=self.row_data, **data_row_info)
-            elif self._row_contains(
-                (".txt", ".text", ".html")) and self._is_url():
-                return TextData(url=self.row_data, **data_row_info)
-            elif not self._is_url():
-                return TextData(text=self.row_data, **data_row_info)
+                return 'image'
+            elif (self._row_contains((".txt", ".text", ".html")) and
```
jtsodapop, line 205: can this be an or statement with line 198? `if self._has_text_annotations():`

labelbox/data/serialization/labelbox_v1/label.py
```diff
@@ -174,44 +171,49 @@ def from_common(cls, label: Label):
...
-            # If it has object annotations and none are text annotations then it must be an image
-            if self._is_url():
-                return ImageData(url=self.row_data, **data_row_info)
-            else:
-                return ImageData(text=self.row_data, **data_row_info)
+            return 'image'
         else:
-            # no annotations to infer data type from.
-            # Use information from the row_data format if possible.
             if self._row_contains((".jpg", ".png", ".jpeg")) and self._is_url():
```
jtsodapop, line 203: can this be an or statement with line 200? `elif self._has_object_annotations():`

jtsodapop (approved):

labelbox/data/serialization/labelbox_v1/label.py
msokoloff1: We would also need to check if there are not object annotations. This also would make the condition on 200 harder to read.

labelbox/data/serialization/labelbox_v1/label.py
msokoloff1: We could but I think this is more readable

labelbox/data/serialization/labelbox_v1/classification.py
```diff
@@ -68,25 +65,14 @@ class LBV1Dropdown(LBV1Feature):
...
     def to_common(self) -> Dropdown:
-        return Dropdown(answer=[
-            ClassificationAnswer(feature_schema_id=answer.schema_id,
-                                 name=answer.title,
-                                 extra={
-                                     'feature_id': answer.feature_id,
-                                     'value': answer.value
-                                 }) for answer in self.answer
-        ])
+        return Dropdown(answer=[answer.to_common() for answer in self.answer])
```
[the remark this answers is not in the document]
msokoloff1: I think it is a list because that is how we export it

PR #355, opened 2021-12-02
Geo annot

jtsodapop:


labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+from .raster import RasterData
+"""TODO: consider how to swap lat,lng to lng,lt when version = 2...
+should the bounds validator be inside the TiledImageData class then, 
+since we need to check on Version?
+"""
+
+VALID_LAT_RANGE = range(-90, 90)
+VALID_LNG_RANGE = range(-180, 180)
+TMS_TILE_SIZE = 256
+MAX_TILES = 300  #TODO: thinking of how to choose an appropriate max tiles number. 18 seems too small, but over 1000 likely seems too large
```
jtsodapop, line 26: @msokoloff1 is it okay to arbitrarily set this for now and see if we need to adjust it later on?

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+import tensorflow as tf
+from retry import retry
+from pyproj import Transformer
+from pydantic import BaseModel, validator, conlist
+from pydantic.class_validators import root_validator
+
+from ..geometry import Point
+from .base_data import BaseData
+from .raster import RasterData
+"""TODO: consider how to swap lat,lng to lng,lt when version = 2...
```
jtsodapop, line 18: @msokoloff1 not sure how to approach this one...originally left validator inside the TiledBounds class since it made the most sense to me, but if a user inputs lng,lt instead and has version=2 in the old import methodology, it would be okay because it flipped

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+        if first_bound == second_bound:
+            raise AssertionError(f"Bounds cannot be equal, contains {bounds}")
+        return bounds
+
+    #bounds are assumed to be in EPSG 4326 as that is what leaflet assumes
+    @root_validator
+    def validate_bounds_lat_lng(cls, values):
+        epsg = values.get('epsg')
+        bounds = values.get('bounds')
+        #TODO: look into messaging that we only support 4326 right now. raise exception, not implemented
```
jtsodapop, line 76: @jtsodapop this is included in TiledBounds class, should consider if we should include it elsewhere

labelbox/data/annotation_types/data/tiled_image.py
msokoloff1, line 26: I would make this a param of `as_raster` with a reasonable default. Probably much smaller than 300. Maybe something like 32.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+from .base_data import BaseData
+from .raster import RasterData
+"""TODO: consider how to swap lat,lng to lng,lt when version = 2...
+should the bounds validator be inside the TiledImageData class then, 
+since we need to check on Version?
+"""
+
+VALID_LAT_RANGE = range(-90, 90)
+VALID_LNG_RANGE = range(-180, 180)
+TMS_TILE_SIZE = 256
```
msokoloff1, line 25: maybe call this DEFAULT_TMS_TILE_SIZE.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+class EPSG(Enum):
+    """ Provides the EPSG for tiled image assets that are currently supported.
+    
+    SIMPLEPIXEL is Simple that can be used to obtain the pixel space coordinates
+
+    >>> epsg = EPSG()
+    """
+    SIMPLEPIXEL = 1
```
msokoloff1, line 40: Aren't we also going to have simple and simple pixel. Simple is zoom 0 and is an actual projection. Simple pixel is more like 'no projection'.

msokoloff1: Maybe we can change this to `EPSG.None` to indicate that the coordinates do not have an epsg.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+    EPSG4326 = 4326
+    EPSG3857 = 3857
+    EPSG3395 = 3395
```
msokoloff1, line 43: I would remove `EPSG3395` it will cause more trouble than it is worth to have hanging around. We also can always add it back.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+        #TODO: look into messaging that we only support 4326 right now. raise exception, not implemented
+
+        if epsg != EPSG.SIMPLEPIXEL:
+            for bound in bounds:
+                lat, lng = bound.y, bound.x
+                if int(lng) not in VALID_LNG_RANGE or int(
+                        lat) not in VALID_LAT_RANGE:
+                    raise ValueError(f"Invalid lat/lng bounds. Found {bounds}. "
+                                     "lat must be in {VALID_LAT_RANGE}. "
+                                     "lng must be in {VALID_LNG_RANGE}.")
```
msokoloff1, line 85: you need an `f` for each line for the interpolation to work

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+        )
+    """
+    url: str
+    name: Optional[str] = "default"
+
+    @validator('url')
+    def validate_url(cls, url):
+        xyz_format = "/{z}/{x}/{y}"
+        if xyz_format not in url:
+            raise AssertionError(f"{url} needs to contain {xyz_format}")
```
msokoloff1, line 106: This is a ValueError since the value of the string is wrong.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+        alternative_layers: List[TileLayer]     
+
+    >>> tiled_image_data = TiledImageData(tile_layer=TileLayer,
+                                  tile_bounds=TiledBounds,
+                                  zoom_levels=[1, 12])
+    """
+    tile_layer: TileLayer
+    tile_bounds: TiledBounds
+    alternative_layers: List[TileLayer] = None
+    zoom_levels: conlist(item_type=int, min_items=2, max_items=2)
```
msokoloff1, line 132: I know I reccomended conlist but maybe use `Tuple[int,int]` since it is easier to read and should be immutable. Also maybe add a validator to make sure that the lower bound is less than or equal to the upper bound

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+    max_native_zoom: int = None
```
msokoloff1, line 133: `Optional[int] = None`

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+    tile_size: Optional[int] = TMS_TILE_SIZE
+    version: int = 2
+
+    def _as_raster(self, zoom=0) -> RasterData:
```
msokoloff1, line 137: I would make this public and change the name to `.as_raster_data`

msokoloff1: also add the max_tile size as a param here.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+            xstart, ystart, xend, yend = self._get_simple_image_params(zoom)
+
+        # Currently our editor doesn't support anything other than 3857.
+        # Since the user provided projection is ignored by the editor
+        #    we will ignore it here and assume that the projection is 3857.
+        else:
+            if self.tile_bounds.epsg != EPSG.EPSG3857:
+                logger.info(
+                    f"User provided EPSG is being ignored {self.tile_bounds.epsg}."
+                )
```
msokoloff1, line 153: I wouldn't include this here. The exporter will do this messy conversion and the expectation is that when this class is instantiated we will be getting the right type. If you want to limit the scope you can just say:
```
if epsg == simple:
    ...
elif epsg == 3857
   ...
else:
   raise ValueError("unsupported epsg... ")

File: labelbox/data/annotation_types/data/tiled_image.py [hunk 13]
```diff
@@ -0,0 +1,364 @@
...
+            xstart, ystart, xend, yend = self._get_3857_image_params(zoom)
+
+        total_n_tiles = (yend - ystart + 1) * (xend - xstart + 1)
+        if total_n_tiles > MAX_TILES:
+            logger.info(
+                f"Too many tiles requested. Total tiles attempted {total_n_tiles}."
+            )
+            return None
```
msokoloff1, line 161: Maybe move this into a separate function and raise an exception instead of logging the tile count.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+        x_tile_start: int,
+        y_tile_start: int,
+        x_tile_end: int,
+        y_tile_end: int,
+        zoom: int,
+    ) -> np.ndarray:
+        """Fetches the tiles and combines them into a single image
+        """
+        tiles = {}
+        with ThreadPoolExecutor(max_workers=TILE_DOWNLOAD_CONCURRENCY) as exc:
```
msokoloff1, line 248: make the executor a param of the class with the ThreadPoolExecutor as the default. That way a user can specify if they want concurrency or not. If they are already running this inside of another thread they might not want us to use a ThreadPoolExecutor.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+            for x in range(x_tile_start, x_tile_end + 1):
+                for y in range(y_tile_start, y_tile_end + 1):
+                    tiles[(x, y)] = exc.submit(self._fetch_tile, x, y, zoom)
+
+        rows = []
+        for y in range(y_tile_start, y_tile_end + 1):
+            rows.append(
+                np.hstack([
+                    tiles[(x, y)].result()
```
msokoloff1, line 257: Use np.pad based on your findings that the tiles might not be guarenteed to be TILE_SIZE. You can pad with zeros but maybe make that a param of the class too.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+    @retry(delay=1, tries=6, backoff=2, max_delay=16)
+    def _fetch_tile(self, x: int, y: int, z: int) -> np.ndarray:
+        """
+        Fetches the image and returns an np array. If the image cannot be fetched, 
+        a padding of expected tile size is instead added.
+        """
+        try:
+            data = requests.get(self.tile_layer.url.format(x=x, y=y, z=z))
+            data.raise_for_status()
+            decoded = tf.image.decode_image(data.content, channels=3).numpy()
```
msokoloff1, line 272: Don't use tensorflow. It is a huge dependency. We already have plenty. You can used any image library open cv or PIL and read from BytesIO.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,364 @@
...
+            if decoded.shape[:2] != (self.tile_size, self.tile_size):
+                logger.warning(
+                    f"Unexpected tile size {decoded.shape}. Results aren't guarenteed to be correct."
+                )
+        except:
+            logger.warning(
+                f"Unable to successfully find tile. for z,x,y: {z},{x},{y} "
+                "Padding is being added as a result.")
+            decoded = np.zeros(shape=(self.tile_size, self.tile_size, 3),
```
msokoloff1, line 281: This will prevent the retry of the request from occurring since there will never be an exception. If you want this functionality, it needs to go where we call `.result()`

labelbox/data/annotation_types/data/tiled_image.py
msokoloff1: The cool part about annotation types is we can just set a specific format that we expect. It is the converters problem to deal with mapping different formats into the annotation type spec.

labelbox/data/annotation_types/data/tiled_image.py
jtsodapop: Could you explain why this needs to be where the `.result()` is sitting? I understand it will prevent the retry of the request but think it's appropriate to be under the `_fetch_tile` method

jtsodapop: Could you explain why this needs to be where the `.result()` is sitting? I understand it will prevent the retry of the request but think it's appropriate to be under the `_fetch_tile` method

labelbox/data/annotation_types/data/tiled_image.py
jtsodapop: I thought we decided to only keep SIMPLEPIXEL and not both Simple, and SimplePixel 

Wouldn't Simple at zoom 0 also be the same as simple pixel if there is no projection? 

Notes:

===========================
- Use cases we need to support
4326 to SIMPLEPIXEL
SIMPLESIMPLEPIXEL to 4326
4326 to 3857
===========================

labelbox/data/annotation_types/data/tiled_image.py
jtsodapop: Hmm, if it is included in where we call `.result()` would it be a try/except on the result, and if there is no exception, then add the padding there?

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,482 @@
...
+
+import requests
+import numpy as np
+
+from retry import retry
+from PIL import Image
+from pyproj import Transformer
+from pygeotile.point import Point as PygeoPoint
+from pydantic import BaseModel, validator
+from pydantic.class_validators import root_validator
```
msokoloff1, line 16: Make sure to add any dependencies to the setup.py 
https://github.com/Labelbox/labelbox-python/blob/83aebf1124c1111156997697f7cf6ed217fefdd2/setup.py#L35

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,482 @@
...
+            xstart, ystart, xend, yend = self._get_simple_image_params(zoom)
+
+        # Currently our editor doesn't support anything other than 3857.
+        # Since the user provided projection is ignored by the editor
+        #    we will ignore it here and assume that the projection is 3857.
+        elif self.tile_bounds.epsg == EPSG.EPSG3857:
+            xstart, ystart, xend, yend = self._get_3857_image_params(zoom)
+        else:
+            raise ValueError(
+                f"Unsupported epsg found...{self.tile_bounds.epsg}")
```
msokoloff1, line 157: I would get rid of this comment. We aren't ignoring the epsg / making an assumption. The exception kind of gives the info that you need.

msokoloff1: Also get rid of the extra dots in the error message.

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,482 @@
...
+        #no multithreading
+        else:
+            for x in range(x_tile_start, x_tile_end + 1):
+                for y in range(y_tile_start, y_tile_end + 1):
+                    try:
+                        tiles[(x, y)] = self._fetch_tile(x, y, zoom)
+                    except:
+                        tiles[(x, y)] = np.zeros(shape=(self.tile_size,
+                                                        self.tile_size, 3),
+                                                 dtype=np.uint8)
```
msokoloff1, line 275: You can consolidate all of this
```
       if multithread:
           tiles = {}
            with ThreadPoolExecutor(
                    max_workers=TILE_DOWNLOAD_CONCURRENCY) as exc:
                for x in range(x_tile_start, x_tile_end + 1):
                    for y in range(y_tile_start, y_tile_end + 1):
                        tiles[(x, y)] = exc.submit(self._fetch_tile, x, y, zoom)

        rows = []
        for y in range(y_tile_start, y_tile_end + 1):
            row = []
            for x in range(x_tile_start, x_tile_end + 1):
                try:
                    if multithread:
                        row.append(tiles[(x, y)].result())
                    else:
                        row.append( self._fetch_tile(x, y, zoom))
                except:
                    row.append(
                        np.zeros(shape=(self.tile_size, self.tile_size, 3), dtype=np.uint8))
            rows.append(np.hstack(row))

File: labelbox/data/annotation_types/data/tiled_image.py [hunk 21]
```diff
@@ -0,0 +1,482 @@
...
+    def _fetch_tile(self, x: int, y: int, z: int) -> np.ndarray:
+        """
+        Fetches the image and returns an np array.
+        """
+        data = requests.get(self.tile_layer.url.format(x=x, y=y, z=z))
+        data.raise_for_status()
+        decoded = np.array(Image.open(BytesIO(data.content)))[..., :3]
+        if decoded.shape[:2] != (self.tile_size, self.tile_size):
+            logger.warning(
+                f"Unexpected tile size {decoded.shape}. Results aren't guarenteed to be correct."
```
msokoloff1, line 297: Maybe just `f"Unexpected tile size {decoded.shape}.`

tests/data/annotation_types/test_tiled_image.py
```diff
@@ -27,3 +29,20 @@ def test_tiled_bounds_same(epsg):
...
+#     tile_layer = TileLayer(url=url, name="slippy map tile")
+#     tile_bounds = TiledBounds(epsg=EPSG.EPSG4326, bounds=bounds_points)
+#     tiled_image_data = TiledImageData(tile_layer=tile_layer,
+#                                       tile_bounds=tile_bounds,
+#                                       zoom_levels=zoom_levels,
+#                                       version=2)
+#     assert isinstance(tiled_image_data, TiledImageData)
+#     assert tiled_image_data.tile_bounds == bounds_points
+#     assert tiled_image_data.tile_layer.url == url
+#     assert tiled_image_data.zoom_levels == zoom_levels
```
msokoloff1, line 48: What's this?

msokoloff1:
Looks good! Just a few nits. Are there any test cases? Where are the ones you showed me this A.M. for converting to and from different projections?

tests/data/annotation_types/test_tiled_image.py
jtsodapop: oops, this should not be commented out

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,460 @@
...
+        y_range = np.max(bounds[:, 1]) - np.min(bounds[:, 1])
+        return (x_range, y_range)
+
+    def _min_max_x_y(self, bounds: np.ndarray):
+        """returns the min x, max x, min y, max y of a numpy array
+        """
+        return np.min(bounds[:, 0]), np.max(bounds[:, 0]), np.min(
+            bounds[:, 1]), np.max(bounds[:, 1])
+
+    def geo_and_geo(self, src_epsg: EPSG, tgt_epsg: EPSG) -> None:
```
msokoloff1, line 365: This should be a classmethod constructor. Same with geo_and_pixel. Then you can make the transform_function a required field. And not have to raise raise Exception("No transformation has been set.").

msokoloff1: Maybe create 3 functions.

1. def create_geo_to_geo_transformer() -> EPSGTransformer (this is just geo_and_geo renamed and it will return the object instead of None. 
2. def create_geo_to_pixel_transformer() -> EPSGTransformer
3. def create_pixel_to_geo_transformer() -> EPSGTransformer

2 and 3 will call the same geo_and_pixel function under the hood but this way it is more clear what the function is doing. Also it gives us more flexibility in the future

labelbox/data/annotation_types/data/tiled_image.py
```diff
@@ -0,0 +1,460 @@
...
+
+                return (x * (local_x_range) / (global_x_range),
+                        y * (local_y_range) / (global_y_range))
+
+            self.transform_function = transform
+
+    def __call__(self, point: Point):
+        if self.transform_function is not None:
+            res = self.transform_function(point.x, point.y)
+            return Point(x=res[0], y=res[1])
```
msokoloff1, line 458: Only a point for now?

msokoloff1, line 458: This should accept any vector geometry (non-mask) and then project each point. and return the updated geometry right? Or was that going to happen in the geometry?

tests/data/annotation_types/test_tiled_image.py
```diff
@@ -27,3 +27,25 @@ def test_tiled_bounds_same(epsg):
...
+                                      version=2)
+    assert isinstance(tiled_image_data, TiledImageData)
+    assert tiled_image_data.tile_bounds.bounds == bounds_points
+    assert tiled_image_data.tile_layer.url == url
+    assert tiled_image_data.zoom_levels == zoom_levels
+
+
+#TODO: create a test from 4326->SIMPLE->3857->4326
+def test_epsg_projections():
+    pass
```
msokoloff1, line 51: You had a bunch of great tests. Why not throw them in here?

msokoloff1 (approved):
This is awesome!

PR #358, opened 2021-12-07
add yapf pre commit

jtsodapop:
looking to add a pre commit hook to add yapf when making pr's


jtsodapop:
It looks like if I add in **any** additional args, the precommit will fail. I've also added style as a .style.yapf file instead, which also did not produce any results.

Tests:
* modified an existing file for poor formatting
* ran yapf as normal, looks like it is adhering to the correct format (unsure and would need to test some more)
* added command line args. usual are the following: -i --verbose --recursive --parallel --style "google" 
* pre-commit would fail

of the most importance are --recursive  --style "google" 

would appreciate any insights here! also shamelessly asked a similar question on SO here:
https://stackoverflow.com/questions/70278167/pre-commit-hook-yapf-returns-different-results-than-running-yapf-in-the-command

jtsodapop:
Based on my SO post, it sounds like my argument had a typo. However, when adding it in the way the user posted, I could not get it to work still. Would appreciate any follow up help on this.

msokoloff1 (approved):

PR #369, opened 2021-12-16
initial work into converters

jtsodapop:


jtsodapop:
image

Looks to pass all test cases

labelbox/data/serialization/labelbox_v1/objects.py
```diff
@@ -42,6 +46,52 @@ def validate_subclasses(cls, value, field):
...
+
+
+class LBV1TIRectangle(LBV1ObjectBase):
+    object_type: Literal['rectangle'] = Field(..., alias='type')
+    geometry: _Coordinates
+
+    def to_common(self) -> Rectangle:
+        coord_list = self.geometry.coordinates[0]
+        start = coord_list[0]
+        end = coord_list[2]
```
msokoloff1, line 90: These positions aren't guaranteed since the bounding box is represented as a polygon. Maybe we should find the min/max x/y values to be used. That way this can't silently break

labelbox/data/serialization/labelbox_v1/objects.py
```diff
@@ -42,6 +46,52 @@ def validate_subclasses(cls, value, field):
         return value
 
 
+class _Coordinates(BaseModel):
+    """Union of the possible coordinate lists in tiled imagery exports"""
+    coordinates: Union[List[float], List[List[float]], List[List[List[float]]]]
```
msokoloff1, line 51: Instead of creating a class that is the union of all other potential coordinate types, I would just type each geometry field individually.

tests/data/serialization/labelbox_v1/test_tiled_image.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    with open(file_path, 'r') as f:
+        payload = json.load(f)
+
+        collection = LBV1Converter.deserialize(payload)
+        collection_as_list = collection.as_list()
+
+        assert len(collection_as_list) == 2
+        assert len(collection_as_list[0].annotations) == 6
+        assert len(collection_as_list[1].annotations) == 8
```
msokoloff1, line 22: You might want to check a few more details than this.

msokoloff1 (approved):

PR #372, opened 2021-12-22
added validator to ensure uid has another value to it

jtsodapop:


labelbox/data/serialization/ndjson/label.py
```diff
@@ -56,7 +56,11 @@ def _generate_annotations(
                     raise TypeError(
                         f"Unsupported annotation. {type(annotation)}")
 
-            data = self._infer_media_type(annotations)(uid=data_row_id)
+            media_type = self._infer_media_type(annotations)
+            if media_type == ImageData:
+                data = media_type(uid=data_row_id, url="")
```
msokoloff1, line 61: Setting a dummy value isn't a good idea. Somewhere else in the code we might see that the url is set and expect it to be valid.

jtsodapop: Should this be set to a specific url in that case to ensure it is valid? Or what is a better alternative?

PR #378, opened 2022-01-07
add in import errors where they were blank and correct a few nb's

jtsodapop:


jtsodapop:
Motivation of this ticket was based on AL-711. Some confusion by users was why their MAL was not successfully showing (likely due to errors). 

Adding in upload.errors where it was missing, and small corrections to existing MAL notebooks

msokoloff1 (approved):

PR #382, opened 2022-01-10
SE-1295: Update upsert_review_queue so quota_factor can accept 1.0 and 0.0

paulnoirel:
Related to ticket https://labelbox.atlassian.net/browse/SE-1295.

Allow SDK to set quota of 0% and 100%. Currently, these values are not accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 # Version 3.11.1 (2022-01-10)
 ## Fix
 * Make `TypedArray` class compatible with `numpy` versions `>= 1.22.0`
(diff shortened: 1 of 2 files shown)


jtsodapop (approved):
Looked at the SE ticket, we should also try to have this reflected accurately with the mutation. Specifically, we shouldn't be allowing things like 400% when it should be maxed at 100%. Can you reach out to the appropriate team on that?

msokoloff1 (approved):
LGTM. I have to make a PR that will fix the builds. It will be ready in about 2 hours.

PR #383, opened 2022-01-10
numpy bug fix

msokoloff1:


msokoloff1:
Taking a different approach

PR #388, opened 2022-01-11
add back html instructions

msokoloff1:


cssetian (approved):
Should we add an extra assertion to `test_attach_instructions`? LGTM otherwise

PR #395, opened 2022-01-14
addition of from shapely function for annotation types

jtsodapop:


labelbox/data/annotation_types/geometry/rectangle.py
```diff
@@ -30,6 +32,28 @@ def geometry(self) -> geojson.geometry.Geometry:
...
+                f"Expected Shapely Polygon. Got {shapely_obj.geom_type}")
+
+        #we only consider 0th index because we only allow for filled polygons
+        obj_coords = np.array(shapely_obj.__geo_interface__['coordinates'][0])
+
+        min_x, max_x = np.min(obj_coords[:, 0]), np.max(obj_coords[:, 0])
+        min_y, max_y = np.min(obj_coords[:, 1]), np.max(obj_coords[:, 1])
+
+        start = [min_x, min_y]
+        end = [max_x, max_y]
```
msokoloff1, line 52: you can just use `minx, miny, maxx, maxy = shapely_obj.bounds`

labelbox/data/annotation_types/geometry/polygon.py
```diff
@@ -30,6 +32,17 @@ def geometry(self) -> geojson.Polygon:
...
         return geojson.Polygon([[(point.x, point.y) for point in self.points]])
 
+    @classmethod
+    def from_shapely(cls, shapely_obj: SPolygon) -> Polygon:
+        """Transforms a shapely object."""
+        #we only consider 0th index because we only allow for filled polygons
+        obj_coords = shapely_obj.__geo_interface__['coordinates'][0]
+        if not isinstance(shapely_obj, SPolygon):
+            raise ValueError(
+                f"Expected Shapely Polygon. Got {shapely_obj.geom_type}")
```
msokoloff1, line 42: I would check the type before getting the coordinates. Also throw a TypeError

labelbox/data/annotation_types/geometry/rectangle.py
```diff
@@ -30,6 +32,28 @@ def geometry(self) -> geojson.geometry.Geometry:
...
         ]])
 
+    @classmethod
+    def from_shapely(cls, shapely_obj: SPolygon) -> Rectangle:
+        """Transforms a shapely object.
+        
+        If the provided shape is a non-rectangular polygon, a rectangle will be
+        returned based on the min and max x,y values."""
+        if not isinstance(shapely_obj, SPolygon):
+            raise ValueError(
```
msokoloff1, line 42: TypeError

labelbox/data/annotation_types/geometry/point.py
```diff
@@ -24,6 +26,16 @@ class Point(Geometry):
     def geometry(self) -> geojson.Point:
         return geojson.Point((self.x, self.y))
 
+    @classmethod
+    def from_shapely(cls, shapely_obj: SPoint) -> Point:
+        """Transforms a shapely object."""
+        if not isinstance(shapely_obj, SPoint):
+            raise ValueError(
```
msokoloff1, line 33: TypeError

labelbox/data/annotation_types/geometry/line.py
```diff
@@ -25,6 +27,17 @@ def geometry(self) -> geojson.MultiLineString:
         return geojson.MultiLineString(
             [[[point.x, point.y] for point in self.points]])
 
+    @classmethod
+    def from_shapely(cls, shapely_obj: SLineString) -> Line:
+        """Transforms a shapely object."""
+        if not isinstance(shapely_obj, SLineString):
+            raise ValueError(
```
msokoloff1, line 34: TyperError

labelbox/data/annotation_types/geometry/rectangle.py
```diff
@@ -30,6 +32,28 @@ def geometry(self) -> geojson.geometry.Geometry:
             [self.start.x, self.start.y],
         ]])
 
+    @classmethod
+    def from_shapely(cls, shapely_obj: SPolygon) -> Rectangle:
```
msokoloff1, line 36: Do these type hints work? I thought you need to put "Rectangle" in quotes because technically Rectangle isn't defined until you are outside the scope of these function definitions (Same with all the other types)

jtsodapop: They work if you include `from __future__ import annotations` at the top of the file

jtsodapop: https://www.python.org/dev/peps/pep-0563/#enabling-the-future-behavior-in-python-3-7

jtsodapop: Ahh thinking about this more, this is only for Python 3.7 and we are still doing Python 3.6-3.8.

I will go with the quote method for now but once we go to Python 3.7-3.9 I think we can do `from __future__ import annotations` if we want to

labelbox/data/annotation_types/geometry/rectangle.py
```diff
@@ -1,8 +1,9 @@
-from typing import Optional, Union, Tuple
+from typing import Optional, Type, Union, Tuple
```
msokoloff1, line 1: I don't think you want to import `Type`

msokoloff1 (approved):
LGTM

PR #403, opened 2022-01-22
Script to list and delete Vertex AI objects

Juanbuhler:
Ref: https://labelbox.atlassian.net/browse/AL-1338

Original intent was to mitigate the risk of high gcloud bills because of leftover endpoints or long running jobs.

This script can list and delete those plus datasets and models. Can be a quick way to check the status of things from the command line.

I also added a print to direct the user to their GCP console, just so people without gcloud experience don't assume this is the only way to do things.

### DIFF
--- a/examples/integrations/gcp/utils/resource_manager.py
+++ b/examples/integrations/gcp/utils/resource_manager.py
@@ -1,0 +1,109 @@
+import os
+import argparse
+import sys
+import operator
+from google.cloud import aiplatform
+from prettytable import PrettyTable
+
+
+def list_objects(object_types, attrs):
+
+    # Use 'id' instead of 'name' since Google uses 'id' in the interface
+    # but 'name' in the attributes
+    display_attrs = ['id' if x == 'name' else x for x in attrs]
+    for t in object_types:
+        objs = getattr(aiplatform, t).list()
+        if objs:
+            table = PrettyTable(display_attrs, horizontal_char='—',
+                                vertical_char='│', junction_char=' ')
+            table.title = t
+            table.align = "l"
+            for obj in objs:
+                row = []
+                for attr in attrs:
+                    # using attrgetter in order to get state.name for jobs
+                    row.append(operator.attrgetter(attr)(obj))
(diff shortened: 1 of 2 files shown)


examples/integrations/gcp/utils/resource_manager.py
```diff
@@ -0,0 +1,109 @@
...
+    job_types = ['AutoMLForecastingTrainingJob', 'AutoMLImageTrainingJob',
+                 'AutoMLTabularTrainingJob', 'AutoMLTextTrainingJob',
+                 'AutoMLVideoTrainingJob', 'BatchPredictionJob',
+                 'CustomContainerTrainingJob', 'CustomJob',
+                 'CustomPythonPackageTrainingJob', 'CustomTrainingJob',
+                 'HyperparameterTuningJob', 'PipelineJob']
+    job_attrs = ['name', 'display_name', 'state.name', 'has_failed']
+
+    endpoint_types = ['Endpoint']
+    endpoint_attrs = ['name', 'display_name']
```
msokoloff1, line 89: Is there no api that just allows users to see what resources are being used / incurring costs without specifying the type of resource? Even if we just provide like a `list_additional_resources` function that maybe just makes sure users don't expect 0 costs but they have something running.

Juanbuhler: Apparently there isn't. I went through a little bit of a stackoverflow rabbit hole just now, ended up here:
https://issuetracker.google.com/issues/124508767?pli=1
It's a requested feature.

For our case, I think we need to assume the used has access to the console and will monitor costs from there. Also--as far as I can tell, the only objects my script lists that can incur costs are running jobs, and endpoints. Should I add some copy saying that, for extra clarity?

msokoloff1: Ok that is fine then. I think that this is a great start. If we feel like there are any gaps in this as we develop additional features we can always revisit it.

msokoloff1 (approved):

PR #404, opened 2022-01-25
bbox training

msokoloff1:
* Code for training using vertex ai
* Updated ETL so that training actually works

### DIFF
--- a/examples/integrations/gcp/src/coordinator/training/images/bounding_box_training_placeholder.py
+++ b/examples/integrations/gcp/src/coordinator/training/images/bounding_box_training_placeholder.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict
 import logging
 
-from job import Job, JobStatus
+from job import Job, JobStatus, JobState
 from google.cloud import aiplatform
 from uuid import uuid4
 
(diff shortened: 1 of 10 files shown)


examples/integrations/gcp/src/coordinator/training/images/bounding_box_training_placeholder.py
```diff
@@ -14,8 +17,28 @@ def parse_args(self, json_data: Dict[str, Any]) -> str:
...
+            bounding_box)
+        job = aiplatform.AutoMLImageTrainingJob(
+            display_name="matt-test-train-automl",
+            prediction_type="object_detection",
+            model_type="MOBILE_TF_LOW_LATENCY_1")
+        model = job.run(
+            dataset=dataset,
+            # Every 1,000 is 1 node-hour. This is close to the minimum (20k).
+            # Increase / parameterize this before deploying
+            budget_milli_node_hours=1000,
```
Juanbuhler, line 34: the interface doesn't let you budget less than 8 node hours, is this possible via the API?

examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
```diff
@@ -19,47 +25,87 @@
...
+    # Vertex will not work unless the input data is a gcs_uri
+    gcs_key = f"training/images/{data_row_uid}.jpg"
+    blob = bucket.blob(gcs_key)
+    blob.upload_from_file(image_bytes, content_type="image/jpg")
+    return f"gs://{bucket.name}/{blob.name}"
+
+
+def process_label(label: Label, bucket: Bucket) -> str:
+    bounding_box_annotations = []
+    # TODO: Only download if the label has annotations
```
Juanbuhler, line 54: Also, are you downloading even if the image was in gs:// already? Won't that make training with images that were already in a bucket unnecessarily slow/expensive?
I'm not sure how common is for Labelbox projects to have their image in Google buckets so this might not be super relevant.

Juanbuhler (commented):
Left a couple of comment/questions, but nothing that should interfere with the execution of the code...

examples/integrations/gcp/src/coordinator/training/images/bounding_box_training_placeholder.py
msokoloff1: This is the min before the api complains (I set this at 500 and it threw an excpetion). Although it ran longer than the budget allowed. So I am not really sure how this is enforced.

examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
msokoloff1: The data can't be in gs:// format because labelbox will only expose a signed url to customers. I agree that we don't want to keep processing the same data but I was going to follow up and maybe make accommodating backend changes.

This runs in a few seconds on my dataset of 250 images so the run time isn't a problem.

Juanbuhler: Oh I see, that makes sense.

Juanbuhler (approved):
lgtm

PR #405, opened 2022-01-26
removal of rasterio

jtsodapop:


jtsodapop:
image

Comparison between the old method and the updated opencv method. similarity is in decimals

Similarity is less than 99% in some cases due to high irregularity, spikiness and number of vertices modified:

    vertices = generate_polygon(center=(250, 250),
                                avg_radius=100,
                                irregularity=0.55,
                                spikiness=0.7,
                                num_vertices=20)

Adapted from here: 
https://stackoverflow.com/questions/8997099/algorithm-to-generate-random-2d-polygon

msokoloff1 (approved):

PR #406, opened 2022-01-27
ner training and eval

msokoloff1:
* Add etl logic for processing ner data
* Add training code for entity modeling using vertex
* Update the coordinator to use new etl and training code
* Add seed script so anyone can test


Juanbuhler (approved):
Looks good to me. Haven't run the code though, but I did read it and follow as carefully as I could.

PR #409, opened 2022-01-31
initial addition to include filters for export_labels

jtsodapop:
Initial phase is done, but there looks to be an issue with the filters themselves. Speaking with Utkarsh more on this before requesting a review

### DIFF
--- a/labelbox/schema/organization.py
+++ b/labelbox/schema/organization.py
@@ -4,6 +4,7 @@ from labelbox.exceptions import LabelboxError
 from labelbox import utils
 from labelbox.orm.db_object import DbObject, query, Entity
 from labelbox.orm.model import Field, Relationship
+from labelbox.schema.invite import InviteLimit
 
 if TYPE_CHECKING:
     from labelbox import Role, User, ProjectRole, Invite, InviteLimit, IAMIntegration
@@ -94,7 +95,7 @@ class Organization(DbObject):
             raise LabelboxError(f"Unable to send invite for email {email}")
         return Entity.Invite(self.client, invite_response)
 
-    def invite_limit(self) -> "InviteLimit":
+    def invite_limit(self) -> InviteLimit:
         """ Retrieve invite limits for the org
         This already accounts for users currently in the org
         Meaining that  `used = users + invites, remaining = limit - (users + invites)`
@@ -108,7 +109,7 @@ class Organization(DbObject):
             """query InvitesLimitPyApi($%s: ID!) {
             invitesLimit(where: {id: $%s}) { used limit remaining }
         }""" % (org_id_param, org_id_param), {org_id_param: self.uid})
-        return Entity.InviteLimit(
+        return InviteLimit(
             **{utils.snake_case(k): v for k, v in res['invitesLimit'].items()})
 
     def remove_user(self, user: "User") -> None:
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -248,7 +248,10 @@ class Project(DbObject, Updateable, Deletable):
         Will only generate a new URL at a max frequency of 30 min.
 
         Args:
+            download (bool): Returns the url if False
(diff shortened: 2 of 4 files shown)


jtsodapop:
This is failing on the following error:

```       with pytest.raises(InvalidQueryError):
>           assert task.created_by() == client.get_user()
E           Failed: DID NOT RAISE <class '***.exceptions.InvalidQueryError'>
```

on this test:
`tests/integration/test_data_rows.py::test_data_row_bulk_creation`

This is causing other PR's to fail and also when I run locally. Unsure for RCA for why this keeps failing here...

labelbox/schema/organization.py
```diff
@@ -4,6 +4,7 @@
 from labelbox import utils
 from labelbox.orm.db_object import DbObject, query, Entity
 from labelbox.orm.model import Field, Relationship
+from labelbox.schema.invite import InviteLimit
 
 if TYPE_CHECKING:
     from labelbox import Role, User, ProjectRole, Invite, InviteLimit, IAMIntegration
```
msokoloff1, line 10: Why not Entity.class name? Do you need to merge the type checking pr in here?

labelbox/schema/project.py
```diff
@@ -250,26 +254,71 @@ def label_generator(self, timeout_seconds=600) -> LabelGenerator:
...
+                f"""{c}: {dictionary.get(c)}""" for c in dictionary
+                if dictionary.get(c)
+            ])
+
+        def _validate_datetime(string_date: str) -> bool:
+            """helper function validate that datetime is as follows: YYYY-MM-DD for the export"""
+            if string_date:
+                try:
+                    datetime.strptime(string_date, "%Y-%m-%d")
+                except:
```
msokoloff1, line 296: I would catch a specific exception here

labelbox/schema/project.py
```diff
@@ -250,26 +254,71 @@ def label_generator(self, timeout_seconds=600) -> LabelGenerator:
...
+            quotations while key will not. This can be toggled with `value_with_quotes`"""
+            if value_with_quotes:
+                return ",".join([
+                    f"""{c}: "{dictionary.get(c)}\"""" for c in dictionary
+                    if dictionary.get(c)
+                ])
+            return ",".join([
+                f"""{c}: {dictionary.get(c)}""" for c in dictionary
+                if dictionary.get(c)
+            ])
```
msokoloff1, line 289: You can probably shrink this. 
```
quote = "\"" if  value_with_quotes else ""
return ",".join([
    f"""{c}: "{dictionary.get(c)}{quote}""" for c in dictionary
                    if dictionary.get(c)
])
```

labelbox/schema/organization.py
msokoloff1: yeah InviteLimit is not an Entity but everything else is. So just remove all the other imports here

msokoloff1: I guess a role isn't either.

msokoloff1: Just make sure all Entities are always type checked with Entity.Name. instead of if TYPE_CHECKING.

labelbox/schema/project.py
```diff
@@ -250,26 +254,69 @@ def label_generator(self, timeout_seconds=600) -> LabelGenerator:
...
+            
+            The string will be formatted as {key}: 'value' for each key. Value will be inclusive of
+            quotations while key will not. This can be toggled with `value_with_quotes`"""
+
+            quote = "\"" if value_with_quotes else ""
+            return ",".join([
+                f"""{c}: {quote}{dictionary.get(c)}{quote}"""
+                for c in dictionary
+                if dictionary.get(c)
+            ])
```
msokoloff1, line 287: Did you test this?

msokoloff1 (approved):
LGTM can you just make a ticket to add tests for this later? I think it is kind of a pain to test and we just want to ship for now.

PR #412, opened 2022-02-03
adds to raise exception to main caller thread from prefetch generator

jtsodapop:


jtsodapop:
Hi @msokoloff1  updated the code to work for non-multithread. Code was changed under impression that errors will be noticed upon reaching into the generator's next item

labelbox/data/generator.py
```diff
@@ -33,7 +33,15 @@ class PrefetchGenerator:
...
-    def __init__(self, data: Iterable[Any], prefetch_limit=20, num_executors=4):
+    #maybe change num exec to just 1, and if 1, make sync
+    #instead of self.get qeue in next, itll return just self._data.next
+    #kwarg on export for multithread, and all other things that use prefetch
+
+    def __init__(self,
+                 data: Iterable[Any],
+                 prefetch_limit=20,
+                 num_executors=4,
+                 multithread: bool = False):
```
msokoloff1, line 44: Instead of a multithread param, lets just say when num_executors == 1 we don't use threading.

labelbox/data/serialization/labelbox_v1/converter.py
```diff
@@ -82,7 +82,7 @@ class LBV1VideoIterator(PrefetchGenerator):
     Generator that fetches video annotations in the background to be faster.
     """
 
-    def __init__(self, examples, client):
+    def __init__(self, examples, client, multithread: bool = False):
```
msokoloff1, line 85: Let's just set the concurrency to 1 in the base class and scrap this

msokoloff1 (changes requested):
Looks good. Just one small change requested

labelbox/data/generator.py
```diff
@@ -64,19 +77,22 @@ def fill_queue(self):
...
     def __iter__(self):
         return self
 
     def __next__(self) -> Any:
-        if self.done:
+        if self.done or self.queue.empty():
             raise StopIteration
         value = self.queue.get()
+        if isinstance(value, ValueError):
+            raise value
```
msokoloff1, line 91: this solution fixes both multithread and non-multithread issue right. So this covers both tickets?

labelbox/data/generator.py
```diff
@@ -64,19 +77,22 @@ def fill_queue(self):
                     raise ValueError("Unexpected None")
                 self.queue.put(value)
         except Exception as e:
-            logger.warning("Unexpected exception while filling the queue. %r",
-                           e)
-        finally:
-            self.queue.put(None)
```
msokoloff1, line 70: you still need this finally where you write a None to the queue to indicate that the job is complete.

labelbox/data/generator.py
jtsodapop: I believe so!

labelbox/data/generator.py
```diff
@@ -43,15 +43,20 @@ def __init__(self, data: Iterable[Any], prefetch_limit=20, num_executors=4):
         self._data = ThreadSafeGen(self._data)
         self.completed_threads = 0
         # Can only iterate over once it the queue.get hangs forever.
+        self.multithread = False if num_executors == 1 else True
```
msokoloff1, line 46: self.multithread = num_executors > 1

labelbox/data/generator.py
```diff
@@ -43,15 +43,20 @@ def __init__(self, data: Iterable[Any], prefetch_limit=20, num_executors=4):
...
-        self.num_executors = num_executors
-        self.threads = [
-            threading.Thread(target=self.fill_queue)
-            for _ in range(num_executors)
-        ]
-        for thread in self.threads:
-            thread.daemon = True
-            thread.start()
+
+        if self.multithread:
```
msokoloff1, line 49: move `self._data = ThreadSafeGen(self._data)` here. We don't need the lock if we aren't using threads. It is one less thing to go wrong.

jtsodapop:
Latet pytest update released on 2-3-2022 broke compatibility with pytest-cases. Some file changes are to accommodate some deprecations along with using an older version of pytest to continue testing until the issue is resolved.

msokoloff1 (approved):

PR #430, opened 2022-02-09
get_resource_tags and create_project_resource_tags

manibox:
This PR adds two calls to our SDK,
- get_resource_tags which returns a list of resource tags for the organization
- create_project_resource_tags(tag_ids = []): to create/associate resource tags to projects

TDD: https://docs.google.com/document/d/1r60QC3Ch_zu3UPEEMVF9R0la4OQlelmQngfhIPMIPEo/edit#heading=h.ql1qz79ruc9v
Jira
https://labelbox.atlassian.net/jira/software/c/projects/ORCH/boards/52?modal=detail&selectedIssue=ORCH-835

https://user-images.githubusercontent.com/74563614/153306005-1aab13fc-42db-459e-b5f0-e8e136e6a0b7.mov

### DIFF
--- a/labelbox/schema/organization.py
+++ b/labelbox/schema/organization.py
@@ -1,3 +1,4 @@
+import json
 from typing import TYPE_CHECKING, List, Optional
 
 from labelbox.exceptions import LabelboxError
@@ -125,6 +126,24 @@ class Organization(DbObject):
             """mutation DeleteMemberPyApi($%s: ID!) {
             updateUser(where: {id: $%s}, data: {deleted: true}) { id deleted }
         }""" % (user_id_param, user_id_param), {user_id_param: user.uid})
+
+    def get_resource_tags(self) -> List["ResourceTag"]:
+        """
+        Returns all resource tags for an organization
+        """
+        res = self.client.execute(
+            """query {
+                organization {
+                    resourceTag {
+                        id,
+                        text,
+                        color
+                    }
+                }
+            }""")
+
+        # print(json.dumps(res['organization']['resourceTag'], indent=2, sort_keys=True))
+        return res['organization']['resourceTag']
 
     def get_iam_integrations(self) -> List["IAMIntegration"]:
         """
--- a/labelbox/schema/organization.py
+++ b/labelbox/schema/organization.py
@@ -44,7 +44,7 @@ class Organization(DbObject):
     users = Relationship.ToMany("User", False)
     projects = Relationship.ToMany("Project", True)
     webhooks = Relationship.ToMany("Webhook", False)
-    resourceTags = Relationship.ToMany("resource_tags", False)
+    resource_tags = Relationship.ToMany("ResourceTags", False)
 
     def invite_user(
             self,
--- a/labelbox/schema/organization.py
+++ b/labelbox/schema/organization.py
@@ -44,7 +44,7 @@ class Organization(DbObject):
     users = Relationship.ToMany("User", False)
     projects = Relationship.ToMany("Project", True)
     webhooks = Relationship.ToMany("Webhook", False)
-    resourceTags = Relationship.ToMany("ResourceTag", False)
+    resourceTags = Relationship.ToMany("resource_tags", False)
 
     def invite_user(
             self,
@@ -161,7 +161,10 @@ class Organization(DbObject):
         query_str = """query GetOrganizationResourceTagsPyApi{organization{resourceTag{%s}}}""" % (
             query.results_query_part(ResourceTag))
 
-        return self.client.execute(query_str)['organization']['resourceTag']
+        return [
+            ResourceTag(self.client, tag) for tag in self.client.execute(
+                query_str)['organization']['resourceTag']
+        ]
 
     def get_iam_integrations(self) -> List["IAMIntegration"]:
         """
--- a/labelbox/schema/organization.py
+++ b/labelbox/schema/organization.py
@@ -142,7 +142,6 @@ class Organization(DbObject):
                 }
             }""")
 
-        # print(json.dumps(res['organization']['resourceTag'], indent=2, sort_keys=True))
         return res['organization']['resourceTag']
 
     def get_iam_integrations(self) -> List["IAMIntegration"]:
--- a/labelbox/schema/organization.py
+++ b/labelbox/schema/organization.py
@@ -161,10 +161,7 @@ class Organization(DbObject):
         query_str = """query GetOrganizationResourceTagsPyApi{organization{resourceTag{%s}}}""" % (
             query.results_query_part(ResourceTag))
 
-        return [
-            ResourceTag(self.client, resourceTag)
-            for resourceTag in res['createResourceTag']
-        ]
+        return self.client.execute(query_str)['organization']['resourceTag']
 
     def get_iam_integrations(self) -> List["IAMIntegration"]:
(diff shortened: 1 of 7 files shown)


labelbox/schema/organization.py
```diff
@@ -126,6 +127,24 @@ def remove_user(self, user: "User") -> None:
...
+                organization {
+                    resourceTag {
+                        id,
+                        text,
+                        color
+                    }
+                }
+            }""")
+
+        # print(json.dumps(res['organization']['resourceTag'], indent=2, sort_keys=True))
```
msokoloff1: ✂️

labelbox/schema/project.py
```diff
@@ -121,6 +121,35 @@ def members(self) -> PaginatedCollection:
                                    {id_param: str(self.uid)},
                                    ["project", "members"], ProjectMember)
 
+    def create_project_resource_tags(self, resourceTagIds) -> List["ProjectResourceTag"]:
```
msokoloff1, line 124: make resourceTagIds snakecase please :)

labelbox/schema/project_resource_tag.py
```diff
@@ -0,0 +1,15 @@
...
+    """ Project resource tag to associate ProjectResourceTag to Project.
+
+    Attributes:
+        resourceTagId (str)
+        projectId (str)
+
+        resource_tag (Relationship): `ToOne` relationship to ResourceTag
+    """
+
+    resourceTagId = Field.ID("resourceTagId")
```
msokoloff1, line 14: This should be snake case. It should automatically convert to camelcase.  Same with projectId below

labelbox/schema/resource_tag.py
```diff
@@ -0,0 +1,15 @@
+from labelbox.orm.db_object import DbObject
+from labelbox.orm.model import Field, Relationship
+
+class ResourceTag(DbObject):
```
msokoloff1, line 4: Are these objects deleteable?

msokoloff1:
You need to install the pre-commit hooks to format the files. Also you mind adding a couple of tests? Overall looks good!

labelbox/schema/project.py
msokoloff1: also add the type hint for that param too List[str]

labelbox/schema/resource_tag.py
manibox: not through the SDK for now. do I still need to add those

labelbox/schema/project.py
```diff
@@ -121,6 +121,35 @@ def members(self) -> PaginatedCollection:
...
+
+        Args:
+            resource_tag_ids
+        Returns:
+            a list of ResourceTag ids that was created.
+        """
+        project_id_param = "projectId"
+        tag_ids_param = "resourceTagIds"
+
+        query_str = """mutation AttatchProjectResourceTags($%s: ID!, $%s: [String!]) {
```
msokoloff1, line 135: queries and mutations in the SDK should end with PyApi. So this should be `mutation AttatchProjectResourceTagsPyApi(...`

labelbox/schema/organization.py
```diff
@@ -126,6 +127,23 @@ def remove_user(self, user: "User") -> None:
             updateUser(where: {id: $%s}, data: {deleted: true}) { id deleted }
         }""" % (user_id_param, user_id_param), {user_id_param: user.uid})
 
+    def get_resource_tags(self) -> List["ResourceTag"]:
+        """
+        Returns all resource tags for an organization
+        """
+        res = self.client.execute(
+            """query {
```
msokoloff1, line 135: You should probably name this query. Something descriptive like `getOrganizationResourceTagsPyApi` could work.

cssetian:
> Also you mind adding a couple of tests?
+1 let's have end to end tests here for each schema + change that assert both success cases and expected failure cases

labelbox/schema/organization.py
```diff
@@ -42,6 +44,7 @@ def __init__(self, *args, **kwargs):
     users = Relationship.ToMany("User", False)
     projects = Relationship.ToMany("Project", True)
     webhooks = Relationship.ToMany("Webhook", False)
+    resourceTags = Relationship.ToMany("ResourceTag", False)
```
msokoloff1, line 47: Should be camel case `resource_tags`.

labelbox/schema/organization.py
```diff
@@ -126,6 +129,40 @@ def remove_user(self, user: "User") -> None:
...
+        res = self.client.execute(query_str, params)
+        return ResourceTag(self.client, res['createResourceTag'])
+
+    def get_resource_tags(self) -> List[ResourceTag]:
+        """
+        Returns all resource tags for an organization
+        """
+        query_str = """query GetOrganizationResourceTagsPyApi{organization{resourceTag{%s}}}""" % (
+            query.results_query_part(ResourceTag))
+
```
msokoloff1, line 163: Is this not a paginated query? What if the number of resource tags gets really large?

labelbox/schema/organization.py
```diff
@@ -126,6 +129,40 @@ def remove_user(self, user: "User") -> None:
...
+        return self.client.execute(query_str)['organization']['resourceTag']
```
msokoloff1, line 164: This should return an iterable of ResourceTag objects instead of json. Ideally this would be a paginated collection. At the very lest it should do the following:
```return [ResourceTag(self.client, tag) for tag in self.client.execute(query_str)['organization']['resourceTag']]```

File: labelbox/schema/project.py [hunk 10]
```diff
@@ -121,6 +122,30 @@ def members(self) -> PaginatedCollection:
...
+            project(where:{id:$%s}){updateProjectResourceTags(input:{%s:$%s}){%s}}}""" % (
+            project_id_param, tag_ids_param, project_id_param, tag_ids_param,
+            tag_ids_param, query.results_query_part(ResourceTag))
+
+        res = self.client.execute(query_str, {
+            project_id_param: self.uid,
+            tag_ids_param: resource_tag_ids
+        })
+
+        return res["project"]["updateProjectResourceTags"]
```
msokoloff1, line 147: This also should return ResourceTag objects instead of the json

labelbox/schema/organization.py
manibox: no its not paginated for v1. we thought about it and didn't think the tag list can get too long initially at least to reduce scope

manibox:
@jtsodapop could you take a look at this

labelbox/schema/organization.py
```diff
@@ -42,6 +44,7 @@ def __init__(self, *args, **kwargs):
     users = Relationship.ToMany("User", False)
     projects = Relationship.ToMany("Project", True)
     webhooks = Relationship.ToMany("Webhook", False)
+    resourceTags = Relationship.ToMany("resource_tags", False)
```
jtsodapop, line 47: I think what @msokoloff1  was referring to was it should be resource_tags = ... since we use snake_case in the sdk for vars

labelbox/schema/organization.py
cssetian: enforcing pagination after releasing it without any will require a breaking change to any consumers of this functionality won't it? Isn't leveraging pagination fairly boilerplate?

msokoloff1 (approved):

PR #436, opened 2022-02-11
jtso/AL-677

jtsodapop:
…nput their API key


gdj0nes:
Lets remove the endpoint too. that can just be highlighted somewhere else

jtsodapop:
objectives:
standardize pip install
remove Google secret storage from all notebooks -> client = Client()
remove getpass

gdj0nes (approved):

PR #443, opened 2022-02-16
deploy model

msokoloff1:
* Add code for deploying vertex models to endpoints


Juanbuhler (approved):
not sure who calls the xxxDeployment methods, otherwise looks good.

will there be a deploy endpoint in coordinator or something like that?

PR #444, opened 2022-02-16
sample inference

msokoloff1:
* Call deployed models with these scripts


Juanbuhler (approved):
oh, I think I see, nvm my question in the other review.

lgtm!

PR #445, opened 2022-02-16
[AL-1555] Updating Test Cases

jtsodapop:
Update to test cases


msokoloff1:
Not necessarily something we have to do now. But I wonder if we can create resources that are scoped to the session if we aren't modifying anything. That might speed up the process because each test won't have to rerun the same resource creation steps. 

Maybe we can follow up in a subsequent "performance" pr.

tests/integration/conftest.py
```diff
@@ -299,6 +259,11 @@ def configured_project(project, client, rand_gen, image_url):
...
 @pytest.fixture
 def configured_project_with_label(client, rand_gen, image_url):
+    """Project with a connected dataset, having one datarow
+    Project contains an ontology with 1 bbox tool
+    Additionally includes a create_label method for any needed extra labels
+    One label is already created and yielded when using fixture
+    """
     project = client.create_project(name=rand_gen(str))
     dataset = client.create_dataset(name=rand_gen(str), projects=project)
     data_row = dataset.create_data_row(row_data=image_url)
```
msokoloff1, line 269: We should use the fixtures instead of defining these here. That way the resources get cleaned up.

msokoloff1:
Can you please make a follow up ticket to seed the right data into the orgs to uncomment out these DA lines? https://github.com/Labelbox/labelbox-python/blob/53031fa48c98b2503f0327f47204cdbc99398a62/tests/integration/test_delegated_access.py#L5-L20

tests/unit/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+                                                 len(ascii_letters) - 1)]
+                           for _ in range(16))
+
+        if field_type is datetime:
+            return datetime.now()
+
+        raise Exception("Can't random generate for field type '%r'" %
+                        field_type)
+
+    return gen
```
msokoloff1, line 25: Since this is used by both unit and integration tests, you can just put a conftest.py under tests/ that contain the shared functions.

msokoloff1 (approved):

PR #450, opened 2022-02-21
jtso/AL-1278 - update supported Python versions for 3.7-3.9

jtsodapop:
removal of 3.6 and addition of 3.9 to sdk


msokoloff1:
This looks good but let's close this until we can get the messaging out.

PR #451, opened 2022-02-22
reference architecture deployment

msokoloff1:
* Simplify pipeline to always run remote
* Update config to support remote deployment
* Read secrets from google cloud instead of env vars
* Add basic deployment scripts

Still requires manual config to set secrets and the rest of the env. A subsequent pr will be created to make deployment / env config as easy as running a single script.


examples/integrations/gcp/src/coordinator/pipelines/images/bounding_box.py
```diff
@@ -5,48 +5,42 @@
...
 
-from pipelines.types import Pipeline, CustomJob, JobStatus, JobState, Job
+from pipelines.types import Pipeline, JobStatus, JobState, Job
 
 logger = logging.getLogger("uvicorn")
 
 
-class BoundingBoxETL(CustomJob):
+class BoundingBoxETL(Job):
+    container_name = "gcr.io/sandbox-5500/training-repo/bounding_box_etl"
```
Juanbuhler, line 14: what's the plan longer term for these container URIs? will they be passed as arguments? right now a user would have to edit these both in the docker-compose file and in all of the etl classes, right?

Juanbuhler (commented):
Could you add at least a mention of how to use deploy.sh in the README?

examples/integrations/gcp/src/coordinator/pipelines/images/bounding_box.py
msokoloff1: oops. yeah the project name should come from an env var. I just updated it. Thanks!

msokoloff1:
> Could you add at least a mention of how to use deploy.sh in the README?

I made some updates to the readme. Instead of explaining all of the steps that need to be taken to prep for deployment via clicking the UI, I am going to follow up in a new PR to #1 automate the deployment and #2 explain how to use the script.

Juanbuhler:
Looking good!

PR #459, opened 2022-02-28
add datarow to etl

msokoloff1:
Write data row ids out to the training file. 

According to the vertex docs:

`dataItemResourceLabels` - Can contain any number of key-value string pairs. 

https://cloud.google.com/vertex-ai/docs/datasets/prepare-image#single-label-classification


Juanbuhler (approved):
LGTM this is great, should make it really easy to put results back for diagnotics

PR #462, opened 2022-03-01
[AL-1623] - ModelRun Export Annotations

jtsodapop:
Looking to generate annotations similar to project.export_labels()

Currently we cannot do project.label_generator() which would return Label objects as it may struggle to infer media type.

### DIFF
--- a/labelbox/schema/model_run.py
+++ b/labelbox/schema/model_run.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Dict, Iterable, Union
+from typing import TYPE_CHECKING, Dict, Iterable, Union, List, Optional, Any
 from pathlib import Path
 import os
 import time
@@ -182,7 +182,11 @@ class ModelRun(DbObject):
         })
 
     @experimental
-    def export_annotations(self, download: bool = False) -> LabelGenerator:
+    def export_labels(
+        self,
+        download: bool = False,
+        timeout_seconds: int = 600
+    ) -> Optional[Union[str, List[Dict[Any, Any]]]]:
         """
         Experimental. To use, make sure client has enable_experimental=True.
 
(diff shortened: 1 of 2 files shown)


labelbox/schema/model_run.py
```diff
@@ -175,6 +181,60 @@ def delete_model_run_data_rows(self, data_row_ids):
             data_row_ids_param: data_row_ids
         })
 
+    @experimental
+    def export_annotations(self, download: bool = False) -> LabelGenerator:
```
msokoloff1, line 185: Let's call this export_labels to be consistent with the project export signature

labelbox/schema/model_run.py
```diff
@@ -175,6 +181,60 @@ def delete_model_run_data_rows(self, data_row_ids):
...
+                query_str, {'modelRunId': self.model_id},
+                experimental=True)['exportModelRunAnnotations']['downloadUrl']
+
+        if not download:
+            return url
+        else:
+            response = requests.get(url)
+            response.raise_for_status()
+            contents = ndjson.loads(response.content)
+            return contents
```
msokoloff1, line 227: I would just copy this as closely as possible:
https://github.com/Labelbox/labelbox-python/blob/ac1a03739ddcfe480416bf0ef169c1addfcf79a4/labelbox/schema/project.py#L342-L364

labelbox/schema/model_run.py
msokoloff1: and change the return type to `Optional[Union[str, List[Dict[Any, Any]]]]`

labelbox/schema/model_run.py
```diff
@@ -175,6 +181,60 @@ def delete_model_run_data_rows(self, data_row_ids):
...
+        """
+        # Need to discuss how to properly get the asset data type from export
+        # Until then, we can't do much with contents as the LBV1Converter may 
+        # not infer correct media type and error out
+        # for now, return the contents as either url or ndjson content
+        # """
```
msokoloff1, line 233: don't worry about this for now. We will follow the pattern established in project and support both model_run.export_labels() and model_run.label_generator() here

msokoloff1: but you don't need to add model_run.label_generator() yet

PR #463, opened 2022-03-02
[AL-1283] Merge Label Generator and Video Label Generator

jtsodapop:
This will need to follow standard deprecation procedure once approved.

Calling project.label_generator() will now determine the type of labels and provide either video, or image/text. Mixed data types will still need to use project.export_labels()

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -221,6 +221,40 @@ class Project(DbObject, Updateable, Deletable):
                 self.uid)
             time.sleep(sleep_time)
 
+    def video_label_generator(self, timeout_seconds=600, **kwargs):
+        """
+        Download video annotations
+
+        Returns:
+            LabelGenerator for accessing labels for each video
+        """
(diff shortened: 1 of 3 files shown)


labelbox/schema/project.py
```diff
@@ -272,11 +245,14 @@ def label_generator(self, timeout_seconds=600, **kwargs):
         is_video = [
             'frames' in row['Label'] for row in json_data if row['Label']
         ]
-        if len(is_video) and any(is_video):
+
+        if len(is_video) and not all(is_video) and any(is_video):
```
msokoloff1, line 249: This condition is off.

msokoloff1 (changes requested):
looks good. just one thing to prevent breaking changes

jtsodapop:
Reaching out to Alex on setting up a deprecation doc, will update this and the AL ticket shortly after

msokoloff1 (approved):
LGTM

PR #467, opened 2022-03-02
[AL-1688] Inclusion of warning messages

jtsodapop:
Warnings for when an ontology includes Dropdown, or when a user tries to create a Dropdown annotation type.

This includes:
* creation of a Dropdown class object
* pulling an OntologyBuilder object from an existing project or ontology
* creation of a Dropdown annotation type

### DIFF
--- a/labelbox/schema/ontology.py
+++ b/labelbox/schema/ontology.py
@@ -83,6 +83,11 @@ class Option:
 @dataclass
 class Classification:
     """
+
+    Deprecation Notice: Dropdown classification is deprecated and will be
+        removed in a future release. Dropdown will also
+        no longer be able to be created in the Editor on 3/31/2022.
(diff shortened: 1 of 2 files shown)


labelbox/schema/ontology.py
```diff
@@ -127,6 +128,12 @@ class Type(Enum):
     schema_id: Optional[str] = None
     feature_schema_id: Optional[str] = None
 
+    def __post_init__(self):
+        if self.class_type == Classification.Type.DROPDOWN:
+            warnings.warn(
+            "Dropdown classification is deprecated and will be "
+            "removed in a future release")
```
msokoloff1, line 135: I think we want to also warn users that dropdown tools cannot be created in the editor after 3/31.

PR #479, opened 2022-03-04
Zeke demo

ezekielemerson:


jtsodapop (changes requested):
no files here. also, is this supposed to be inside labelbox-python repo or mlse repo?

ezekielemerson:
I was intending to be inside the labelbox-python repo. Does that seem incorrect?

PR #480, opened 2022-03-05
prototype batch predictions for image classifier

Juanbuhler:
I ended up having to hard code more than the model run uid. I needed the ontology and didn't know how to get that without the project, and not sure how to get the model run object without the model id either.

The etl file name might be another issue, not sure we are tracking/saving that somewhere.

### DIFF
--- a/examples/integrations/gcp/custom_example/batch_predictions/batch_prediction.py
+++ b/examples/integrations/gcp/custom_example/batch_predictions/batch_prediction.py
@@ -1,4 +1,3 @@
-
 import os
 import ndjson
 import uuid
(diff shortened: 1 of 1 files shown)


examples/integrations/gcp/custom_example/batch_predictions/batch_prediction.py
```diff
@@ -0,0 +1,116 @@
...
+        annotations=[
+            ClassificationAnnotation(
+                value=Radio(answer=ClassificationAnswer(
+                    name=l['prediction']['displayNames'][0])),
+                name=ontology.classifications[0].instructions
+            )
+        ]
+    ))
+
+signer = lambda _bytes: client.upload_data(content=_bytes, sign=True)
```
msokoloff1, line 105: This is unused

msokoloff1 (approved):
One small thing. Other than that, this looks good!

examples/integrations/gcp/custom_example/batch_predictions/batch_prediction.py
Juanbuhler: copy&paste mistake :)

PR #486, opened 2022-03-07
[AL-1723] NER Confusion Matrix

jtsodapop:
https://labelbox.atlassian.net/browse/AL-1723

Addition of confusion matrix logic for NER/ TextEntity annotations

### DIFF
--- a/labelbox/data/metrics/iou/calculation.py
+++ b/labelbox/data/metrics/iou/calculation.py
@@ -4,6 +4,8 @@ from itertools import product
 from shapely.geometry import Polygon
 import numpy as np
 
+from labelbox.data.annotation_types.ner import TextEntity
+
 from ..group import get_feature_pairs, get_identifying_key, has_no_annotations, has_no_matching_annotations
 from ...annotation_types import (ObjectAnnotation, ClassificationAnnotation,
                                  Mask, Geometry, Point, Line, Checklist, Text,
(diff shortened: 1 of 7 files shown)


labelbox/data/metrics/iou/calculation.py
```diff
@@ -269,3 +271,26 @@ def _ensure_valid_poly(poly):
...
+def _ner_iou(ner1: TextEntity, ner2: TextEntity):
+    """Computes iou between two text entity annotations"""
+    intersection_start, intersection_end = max(ner1.start, ner2.start), min(
+        ner1.end, ner2.end)
+    union_start, union_end = min(ner1.start,
+                                 ner2.start), max(ner1.end, ner2.end)
+    #edge case of only one character in text
+    if union_start == union_end:
+        return 1
+    return (intersection_end - intersection_start) / (union_end - union_start)
```
msokoloff1, line 296: `intersection_end - intersection_start` can actually be negative here if the words do not intersect.

You need something like:
```
if intersection_start > intersection_end:
    return 0
```

labelbox/data/metrics/iou/calculation.py
```diff
@@ -269,3 +271,26 @@ def _ensure_valid_poly(poly):
...
+        pairs.append((ground_truth, prediction, score))
+    return pairs
+
+
+def _ner_iou(ner1: TextEntity, ner2: TextEntity):
+    """Computes iou between two text entity annotations"""
+    intersection_start, intersection_end = max(ner1.start, ner2.start), min(
+        ner1.end, ner2.end)
+    union_start, union_end = min(ner1.start,
+                                 ner2.start), max(ner1.end, ner2.end)
```
msokoloff1, line 292: This also assumes that there is overlap. Otherwise the characters in-between the labels are considered part of the union. This doesn't matter as long as you return 0 if there is no overlap.

msokoloff1 (changes requested):
Found a small bug. Also the tests are lacking a bit. I would add some iou tests if you are feeling ambitious but this is simple enough where it might not matter.

labelbox/data/metrics/iou/calculation.py
jtsodapop: Thanks for catching! Will add

msokoloff1 (approved):

PR #487, opened 2022-03-08
image partition

msokoloff1:
* Partition the dataset based on the split specified in the export

### DIFF
--- a/examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
+++ b/examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
@@ -107,7 +107,6 @@ def process_label(label: Label, partition: Optional[Partition],
             gcs_uri,
         'boundingBoxAnnotations':
             bounding_box_annotations[:VERTEX_MAX_EXAMPLES_PER_IMAGE],
-        # TODO: Replace with the split from the export in the future.
         'dataItemResourceLabels': {
             "aiplatform.googleapis.com/ml_use": partition_mapping[partition],
             "dataRowId": label.data.uid
--- a/examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
+++ b/examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
@@ -5,7 +5,7 @@ import json
 import argparse
 import time
 import logging
-from typing import Tuple
+from typing import Tuple, Union, Literal, Optional
 from io import BytesIO
 import os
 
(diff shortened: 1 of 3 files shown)


examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
```diff
@@ -61,7 +69,12 @@ def upload_to_gcs(image_bytes: BytesIO, data_row_uid: str, h: int, w: int,
     return f"gs://{bucket.name}/{blob.name}"
 
 
-def process_label(label: Label, bucket: Bucket) -> str:
+def process_label(label: Label, partition: Optional[Partition],
+                  bucket: Bucket) -> str:
```
Juanbuhler, line 73: nit: I'd prefer to set the bucket type as "storage.Bucket" instead of importing Bucket and using it directly, because it is more descriptive. Also it's the way the text etl does it...

examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
```diff
@@ -61,7 +69,12 @@ def upload_to_gcs(image_bytes: BytesIO, data_row_uid: str, h: int, w: int,
...
+    if partition is None:
+        logger.warning("No partition assigned. Skipping.")
+        return
+
```
Juanbuhler, line 77: I figured something like this was missing previously.
But this will never be the case, right? There's always going to be a split, even if a default one, correct?

examples/integrations/gcp/src/jobs/etl/images/classification/classification_etl.py
```diff
@@ -66,15 +79,19 @@ def process_single_classification_label(label: Label, bucket: Bucket) -> str:
         'classificationAnnotation': classification,
         # TODO: Replace with the split from the export in the future.
```
Juanbuhler, line 80: nit: we don't need that comment anymore :)

examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
msokoloff1: yeah definitely

examples/integrations/gcp/src/jobs/etl/images/bounding_box/bounding_box_etl.py
msokoloff1: Unfortunately it can be empty.. If they launch the job from the UI it won't be through

Juanbuhler (approved):
Couple of tiny nits but nothing that changes behavior.
LGTM

PR #491, opened 2022-03-10
[Fix] Utils not addressing nums

jtsodapop:
Related to:
https://labelbox.slack.com/archives/C02SQ2VF6KG/p1646872771810379

The current _convert() method ignores numbers and only accounts for letters. Users passing in string are stating that they expect numbers to be retained.

The proposed change will allow both numbers and letters to be retained.

### DIFF
--- a/labelbox/utils.py
+++ b/labelbox/utils.py
@@ -2,7 +2,7 @@ import re
 
 
 def _convert(s, sep, title):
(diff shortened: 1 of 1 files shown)


jtsodapop:
Closing for time being. There ought to be some bigger discussion on if we want to support this as well as special characters, etc.

PR #496, opened 2022-03-11
Skip Tag Test

jtsodapop:
Mark tag test until one of 2 following conditions are met:
* We have an easy way to delete all tags through the sdk for an organization
* We have unlimited tag permissions for all testing organizations

Reason:
* Tests will consistently fail due to the fact that orgs are limited to a default of 3 tags. 
* We do not want to manually delete tags every time we want to push to sdk

### DIFF
--- a/tests/integration/test_project.py
+++ b/tests/integration/test_project.py
@@ -1,5 +1,4 @@
 import json
-import re
 
 import pytest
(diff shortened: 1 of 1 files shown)


Juanbuhler (approved):
lgtm

PR #501, opened 2022-03-15
[AL-1611] Store Mask Values

jtsodapop:
Allows for caching mask values. This way, when we call on the mask, it is stored on the disk rather than in memory, so it does not take up all the memory space for large projects/number of labels of masks

### DIFF
--- a/labelbox/data/annotation_types/data/raster.py
+++ b/labelbox/data/annotation_types/data/raster.py
@@ -11,7 +11,7 @@ from pydantic import BaseModel
 from pydantic import root_validator
 
 from .base_data import BaseData
-from ..types import TypedArray
+from ..types import TypedArray, TempFile
 
 
 class RasterData(BaseModel, ABC):
@@ -21,7 +21,7 @@ class RasterData(BaseModel, ABC):
     file_path: Optional[str] = None
     url: Optional[str] = None
     arr: Optional[TypedArray[Literal['uint8']]] = None
-    cache_data: Optional[bytes] = None
+    cache_data: Optional[TempFile] = None
 
     @classmethod
     def from_2D_arr(cls, arr: Union[TypedArray[Literal['uint8']],
--- a/labelbox/data/annotation_types/data/raster.py
+++ b/labelbox/data/annotation_types/data/raster.py
@@ -13,7 +13,7 @@ from joblib import Memory
 from .base_data import BaseData
 from ..types import TypedArray
 
-memory = Memory("/tmp")
+memory = Memory(location="/tmp", verbose=0)
 
 
 #restructured outside of class to use memory cache
--- a/labelbox/data/annotation_types/data/raster.py
+++ b/labelbox/data/annotation_types/data/raster.py
@@ -1,3 +1,4 @@
+import tempfile
 from abc import ABC
 from io import BytesIO
 from typing import Callable, Optional, Union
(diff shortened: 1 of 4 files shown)


labelbox/data/annotation_types/data/raster.py
```diff
@@ -8,10 +8,44 @@
...
 
 from .base_data import BaseData
 from ..types import TypedArray
 
+memory = Memory(location="/tmp", verbose=0)
+
+
+#restructured outside of class to use memory cache
+@memory.cache
+def bytes_to_np(image_bytes: bytes) -> np.ndarray:
```
msokoloff1, line 21: these keys are huge. It will prob be expensive to create a hash. We know the result is unique for each object so we don't need to hash on the bytes.

msokoloff1: Also the cache needs to be optional

labelbox/data/annotation_types/data/raster.py
```diff
@@ -53,37 +87,11 @@ def from_2D_arr(cls, arr: Union[TypedArray[Literal['uint8']],
...
-        if arr.dtype != np.uint8:
-            raise TypeError(f"image data type must be uint8. Found {arr.dtype}")
-
-        im_bytes = BytesIO()
-        Image.fromarray(arr).save(im_bytes, format="PNG")
-        return im_bytes.getvalue()
+        self._validate_array(arr)
+        return np_to_bytes(arr)
 
     @property
```
msokoloff1, line 96: This is outside of the changes in the PR so it won't let me make the comment in the right spot but we are still caching the results in memory in addition to disk.
```
arr = self.bytes_to_np(im_bytes)
```

labelbox/data/annotation_types/data/raster.py
```diff
@@ -8,10 +8,44 @@
 from google.api_core import retry
 from pydantic import BaseModel
 from pydantic import root_validator
+from joblib import Memory
 
 from .base_data import BaseData
 from ..types import TypedArray
 
+memory = Memory(location="/tmp", verbose=0)
```
msokoloff1, line 16: This prob won't work on windows.

msokoloff1: Also, I am thinking maybe joblib isn't what we want. I was hoping this would handle a lot of complexity for us. But I think it is easy enough to update the caching logic that already exists to save to disk if users want. Sorry :(

msokoloff1: and 1 less dependency is nice

labelbox/data/annotation_types/data/raster.py
```diff
@@ -20,6 +21,7 @@ class RasterData(BaseModel, ABC):
     file_path: Optional[str] = None
     url: Optional[str] = None
     arr: Optional[TypedArray[Literal['uint8']]] = None
+    cache_data: Optional[TempFile] = None
```
jtsodapop, line 24: Note: this is here so that a reference is persisted when creating the temp file. This is so that garbage collection doesn't immediately remove the file and relevant file path

msokoloff1: This isn't necessary. If you maintain a reference to this object and set `self.cache_data = ...` then this won't get gc-ed.

msokoloff1:
So when this object is deleted is the temporary file also deleted? We need to be careful not to fill up the disk space on users machines. I also think that this should be an option but not the default behavior. Can you add a way to toggle this cachine behavior on and off?

jtsodapop:
After some discussion, marking this as complete and closing PR, but we can always revisit this. 

Reasons are:

No customer has asked this except for one, who did not understand the issue was unrelated to how this is being cached

This is potentially a breaking change and can affect clients' hardware space

If this is requested more in the future, we can always look this up again

PR #516, opened 2022-03-28
[AL-1886] Inclusion of Scope param to Classification

jtsodapop:


labelbox/schema/ontology.py
```diff
@@ -151,21 +156,31 @@ def from_dict(cls, dictionary: Dict[str, Any]) -> Dict[str, Any]:
...
             "options": [o.asdict() for o in self.options],
-            "schemaNodeId": self.schema_id,
-            "featureSchemaId": self.feature_schema_id
+            "schemaNodeId":
+                self.schema_id,
+            "featureSchemaId":
+                self.feature_schema_id,
+            "scope":
+                self.scope.value
+                if self.scope is not None else self.Scope.GLOBAL.value
```
msokoloff1, line 183: So we always want to include global scope by default? This won't cause any problems?

jtsodapop: Yeah, when talking with Amanda about this, for new ontologies it should be okay to default to global, since global is inclusive of image/videos while index is specific to videos

I am unsure how to update the code to check for this though:
```We also shouldn't allow users to update the scope if classifications annotations exist for that classification.```

msokoloff1 (approved):

PR #527, opened 2022-03-31
[AL-1859] Update SDK Tests for On-Prem

jtsodapop:
Update to test cases and availability to run on-prem given your local environment has the following var:

`LABELBOX_TEST_API_KEY_ONPREM` - take the api key from the onprem instance

`LABELBOX_TEST_ONPREM_HOSTNAME` example - https://app.replicated-bb2f865.labelbox.dev


jtsodapop:
It is difficult to make this consistently pass all test cases. The test results are very flaky for on-prem instance.

Example test case:
client.upload_data() method will intermittently fail and we will get a NoneType error (likely due to the result not being available yet)

Running several times will allow the test case to pass, but it is not guaranteed with X number of times.

tests/integration/annotation_import/conftest.py
```diff
@@ -223,11 +223,12 @@ def entity_inference(prediction_id_mapping):
 @pytest.fixture
 def segmentation_inference(prediction_id_mapping):
     segmentation = prediction_id_mapping['superpixel'].copy()
-    segmentation.update(
-        {'mask': {
-            'instanceURI': "sampleuri",
+    segmentation.update({
+        'mask': {
+            'instanceURI': "https://picsum.photos/200/300",
```
msokoloff1, line 228: This potentially could result in a failed import. If we run the tests enough picsum starts limiting us. Also if there are no [0,0,0] pixels then the import is unsuccessful.

msokoloff1 (approved):
One nit, otherwise looks good.

PR #528, opened 2022-03-31
revised text and video export notebooks

ezekielemerson:
Significantly revised the text and video export notebooks to demonstrate how to parse through the exports in both cases, which was especially needed for video example, and how to convert annotation types to ndjson for text. I also removed the very project-specific example at the end of the original text notebook, which had no reusability. Additionally, the video export notebook also came with a project ID, but the visualization task at the end is well-written and applies to all projects, so I kept it.


jtsodapop (changes requested):
When trying to look at the notebooks with reviewnb, it says that these are not valid notebooks. can you confirm these are good to go?

jtsodapop:
Closing this out in favor of opening up another PR so as to avoid conflicts

PR #529, opened 2022-03-31
revised video and text export notebooks

ezekielemerson:
Significantly revised the text and video export notebooks to demonstrate how to parse through the exports in both cases, which was especially needed for video example, and how to convert annotation types to ndjson for text. I also removed the very project-specific example at the end of the original text notebook, which had no reusability. Additionally, the video export notebook also came with a project ID, but the visualization task at the end is well-written and applies to all projects, so I kept it.

### DIFF
--- a/examples/label_export/text.ipynb
+++ b/examples/label_export/text.ipynb
@@ -1,343 +1,324 @@
 {
-    "cells": [
-        {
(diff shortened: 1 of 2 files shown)


gdj0nes (approved):
LGTM

PR #539, opened 2022-04-07
easier deployment

msokoloff1:
* Includes all aspects of deployment now
    - Enable gcp resources for first time access
    - Create gcs bucket if it doesn't exist and if provided validate the bucket meets the vertex requirements
    - Creates static ip and checks if the instance responds to requests
    - Teardown potentially dangling resources if the deployment fails 
    - Prints important information about the deployment configuration so that users don't accidentally configure the deployment improperly. 
* New deployment scripts
    - deploying new changes to the existing instance without changing the public ip address
    - update secrets for the instance
* Complete teardown logic
    - teardown script deletes all resources now
* Support parallel deployments with different ecr images to enable experimentation or multiple labelbox organizations within a single gcp project.
* Rename WEBHOOK_SECRET env var to SERVICE_SECRET because we are no longer calling the integration a webhook.


Juanbuhler (approved):
I love having stuff like that configure deployment script, makes everything tons easier!

Automatic service enablement is great too.

LGTM!

PR #544, opened 2022-04-13
[AL-1946] Subclass classification

jtsodapop:


jtsodapop:
Note: The new test is likely to break when we update nested classifications to work with MAL/LI. The test case will just need to be updated to include the nested classifications under `video_serialized_bbox_label`

jtsodapop:
Closing this out for the following reasons:
* This code is built on top of broken existing code. Updated backend code changes may require a format change.
* We are exploring updating changing the format for video export data. Because of this, this code may also have to change. Closing this out and will revisit in the future after we make the necessary updated changes.

PR #548, opened 2022-04-19
[AL-2173] Update to on-prem test case input

jtsodapop:
update to onprem test case input from instance id to hostname

After talking to Joanna/Joe, sounds like the hostname will be highly variable. At the same time, they mentioned that they did not usually have any problems with `hostname` appended with /api/_gql.


msokoloff1 (approved):

hydak:
merging this with an override due to a Snyk issue

PR #549, opened 2022-04-20
[AL-2193] Deprecation message for Python 3.6

jtsodapop:
We will be no longer supporting 3.6 starting 6/1/22. We will support 3.7-3.9

### DIFF
--- a/labelbox/__init__.py
+++ b/labelbox/__init__.py
@@ -1,6 +1,13 @@
 name = "labelbox"
 __version__ = "3.19.1"
 
+import sys
+import warnings
+
+if sys.version_info < (3, 7):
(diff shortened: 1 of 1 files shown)


msokoloff1 (approved):

msokoloff1:
We should probably add official support for 3.9 now. Basically just add one more test case.

jtsodapop:
> We should probably add official support for 3.9 now. Basically just add one more set of tests

This older PR could be reopened now and I think captures 3.7-3.9
https://github.com/Labelbox/labelbox-python/pull/450/files

hydak:
merging this with an override due to a Snyk issue

PR #550, opened 2022-04-21
[AL-2196] Test Update for Test batch

jtsodapop:
We did not have logging for it before, but this was updated - deleting a dataset twice technically tries to delete a nonexistent dataset. Because of this, the SDK tests now fail from test_batch.py. This update removes the double delete problem


msokoloff1 (approved):

hydak:
merging without snyk review because that functionality is broken

PR #554, opened 2022-04-27
[X-0] Fix M1 test issue

farkob:
Running tests locally on Mac M1's was failing with ```OSError: Could not find lib geos_c or load any of its variants ['libgeos_c.so.1', 'libgeos_c.so'].```

Adding `libgeos-dev` to the install list in Dockerfile resolves the issue.


jburnitz:
can confirm, fixes this on M1 processor

jtsodapop (approved):

PR #555, opened 2022-04-27
AL-2183 - Add MediaType support to Project

paulnoirel:
This is the implementation of https://labelbox.atlassian.net/browse/AL-2183 to create, retrieve and update media types for projects.

### Terminology

We use the terminology `MediaType` and `media_type` to match the GraphQL type and make it easier to use compared to `allowedMediaType` found in the `Project` table.

### Implementation of the MediaType enumeration

The names in the `MediaType` enumeration use title case. This matches the convention of the existing `QueueMode` enumeration in the `Project` class. With that said, the enumeration also supports upper case names for compatibility with the GraphQL type.
To support the case when `media_type` is not set for the project, if `MediaType` is called for `None` then it returns `MediaType.Unknown`.

In order to match the UI, a few aliases were added compared to the GraphQL `MediaType` type:

|Type GraphQL| Alias|
|-|-|
| PDF | Document |
| TMS_SIMPLE | Simple_Tile|
|TMS_GEO | Geospatial_Tile |

### Case of MediaType.Unknown

Although the member `MediaType.UNKOWN` exists in GraphQL, it is not a supported value. Since using it would lead to a HTTP 500 (internal server error) error, this value is filtered for both `client.create_project()` and `Project.update()`.

### Side modifcation

The file `labelbox/orm/db_object.py` is slightly modified to support the case where data_type is not set (aka value is None).


tests/integration/test_project.py
```diff
@@ -191,3 +201,25 @@ def test_queue_mode(configured_project: Project):
...
     assert configured_project.queue_mode() == configured_project.QueueMode.Batch
+
+
+def test_media_type(client, configured_project: Project, rand_gen):
+    # Existing project
+    assert configured_project.media_type is None or isinstance(
+        configured_project.media_type, Project.MediaType)
+
+    # No media_type
+    project = client.create_project(name=rand_gen(str))
```
jtsodapop, line 212: do we need to create a new project to test this? or can we just use the configured project for this?

tests/integration/test_project.py
```diff
@@ -191,3 +201,25 @@ def test_queue_mode(configured_project: Project):
...
+    assert project.media_type == Project.MediaType.Unknown
+    project.update(media_type=Project.MediaType.Image)
+    assert project.media_type == Project.MediaType.Image
+    project.delete()
+
+    for media_type in Project.MediaType:
```
jtsodapop, line 218: is this different than the assert on media type of None vs. Unknown?

maybe we should have this as a separate test?

jtsodapop (approved):
lgtm, minor nits

labelbox/schema/project.py
```diff
@@ -63,6 +63,37 @@ class Project(DbObject, Updateable, Deletable):
         benchmarks (Relationship): `ToMany` relationship to Benchmark
         ontology (Relationship): `ToOne` relationship to Ontology
     """
+
+    class MediaType(Enum):
```
msokoloff1, line 67: I don't think these media types are going to be unique to a project. In the future if we were typing a dataset it would be weird to say `client.create_dataset(..., media_type = Project.MediaType.Image)` Let's maybe move this to the data_row.py as a top level enum (top level meaning not nested under a DataRow).

labelbox/client.py
```diff
@@ -611,6 +611,20 @@ def create_project(self, **kwargs) -> Project:
...
+            if isinstance(media_type, Project.MediaType
+                         ) and media_type is not Project.MediaType.Unknown:
+                kwargs["media_type"] = media_type.value
+            else:
+                media_types = [
+                    item for item in Project.MediaType.__members__
+                    if item != "Unknown"
+                ]
+                raise TypeError(
+                    f"{media_type} is not a supported type. Please use any of {media_types} from the {type(media_type).__name__} enumeration."
```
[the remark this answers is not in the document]
paulnoirel: As discussed, the message was changed to:
```f"{media_type} is not a valid media type. Use any of MediaType.get_accepted_members()} from MediaType. Example: MediaType.Image."```

File: tests/integration/test_project.py (hunk 1 above)
Comment (paulnoirel) [reply tests/integration/test_project.py:212]: Test modified as suggested.

File: tests/integration/test_project.py (hunk 2 above)
Comment (paulnoirel) [reply tests/integration/test_project.py]: Test amended.

File: labelbox/schema/media_type.py [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+            if member.name == name.upper():
+                return member
+
+    @classmethod
+    def is_accepted(cls, value):
+        return isinstance(value,
+                          cls) and value not in [cls.Unknown, cls.Unsupported]
+
+    @classmethod
+    def get_accepted_members(cls):
```
msokoloff1, line 42: Maybe rename this to is_supported and get_supported_members

labelbox/schema/project.py
paulnoirel: The `MediaType` enumeration was moved to its own file (media_type.py) so it could be conveniently used from different classes.

labelbox/schema/media_type.py
paulnoirel: Methods' names refactored as suggested.

msokoloff1 (approved):

PR #556, opened 2022-04-27
[AL-2296] Remove Scope key when turning Nested Classification to dict

jtsodapop:
update to ontology classifications.asdict to not include scope if it is a subclass

### DIFF
--- a/labelbox/schema/ontology.py
+++ b/labelbox/schema/ontology.py
@@ -69,7 +69,7 @@ class Option:
             "featureSchemaId": self.feature_schema_id,
             "label": self.label,
             "value": self.value,
-            "options": [o.asdict() for o in self.options]
+            "options": [o.asdict(is_subclass=True) for o in self.options]
         }
 
     def add_option(self, option: 'Classification') -> None:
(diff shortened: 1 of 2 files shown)


AmandaJBell (commented):
LGTM but matt should probably okay too, since this isn't my arena

labelbox/schema/ontology.py
```diff
@@ -159,29 +159,25 @@ def from_dict(cls, dictionary: Dict[str, Any]) -> Dict[str, Any]:
...
-            "scope":
-                self.scope.value
-                if self.scope is not None else self.Scope.GLOBAL.value
+            "schemaNodeId": self.schema_id,
+            "featureSchemaId": self.feature_schema_id
         }
+        if is_subclass:
+            return classification
+        classification[
+            "scope"] = self.scope.value if self.scope is not None else self.Scope.GLOBAL.value
```
AmandaJBell, line 179: Instead of "not None" could you just say `if self.scope`? (I'm not a python expert so not 100% sure)

jtsodapop: I think so! It may be a minor nit though but I can change if it is more clear

msokoloff1 (approved):

PR #557, opened 2022-04-28
[AL-1707] Update to Delegated Access Test Cases

jtsodapop:


.github/workflows/python-package.yml
```diff
@@ -17,12 +17,15 @@ jobs:
           - python-version: 3.6
             prod-key: LABELBOX_API_KEY
             staging-key: STAGING_LABELBOX_API_KEY
+            da-test-key: DA_GCP_LABELBOX_API_KEY
```
jtsodapop, line 20: Wasn't entirely sure if there was a better way than to add the key into every matrix slot. Added this in this way since it seemed to work like this

tests/integration/test_delegated_access.py
```diff
@@ -1,35 +1,56 @@
...
+
 
-@pytest.mark.skip("Can only be tested in specific organizations.")
-def test_default_integration(client):
-    # This tests assumes the following:
-    # 1. gcp delegated access is configured to work with utkarsh-da-test-bucket
-    # 2. the integration name is gcp test
-    # 3. This integration is the default
+@pytest.mark.skipif(os.environ.get("DA_GCP_LABELBOX_API_KEY") is None,
+                    reason="DA_GCP_LABELBOX_API_KEY not found")
```
msokoloff1, line 10: Love this. When I saw the changes to the makefile I was thinking we need something like this.

msokoloff1 (approved):
Looks good!

.github/workflows/python-package.yml
msokoloff1: https://github.com/Labelbox/labelbox-python/blob/5679390fa0ffa1b7b7c7f6311de43866178d6574/.github/workflows/python-package.yml#L90

jtsodapop: I think I tried doing this and it was tossing me an error - when I looked up the error, it told me I was not specifying the variable in my yaml file and couldn't find it. I'll merge this for the time being but can revisit this for sure

PR #563, opened 2022-05-03
[AL-2218] Adding optional metadata_fields to create_data_row()

kkim-labelbox:
Changes:
- Adding metadata_fields as an optional parameter to create_data_row()
- Setting metadata_fields as a new Field in DataRow as a new field type, List
- Adjusting graphql query parameters to adjust for metadataFields
- Caching metadata ontology

Heads up:
- Thinking of a major version upgrade from 3.20.x to 3.21.0 for the overall change. I will make this version upgrade once all parts of the TDD are implemented

Tests:
- Added a case for setting metadata_fields for data row integration test

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -90,6 +90,7 @@ class Client:
             'Authorization': 'Bearer %s' % api_key,
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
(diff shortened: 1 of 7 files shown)


kkim-labelbox:
Messed up the branch name. Will create a new PR

PR #565, opened 2022-05-05
[AL-2218] Adding optional metadata_fields to create_data_row()

kkim-labelbox:
https://labelbox.atlassian.net/browse/AL-2218
Changes:
- Adding metadata_fields as an optional parameter to create_data_row()
- Setting metadata_fields as a new Field in DataRow as a new field type, List
- Adjusting graphql query parameters to adjust for metadataFields
- Caching metadata ontology

Heads up:
- Thinking of a major version upgrade from 3.20.x to 3.21.0 for the overall change. I will make this version upgrade once all parts of the TDD are implemented

Tests:
- Added a case for setting metadata_fields for data row integration test

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -37,7 +37,7 @@ from labelbox.schema.media_type import MediaType
 logger = logging.getLogger(__name__)
 
 _LABELBOX_API_KEY = "LABELBOX_API_KEY"
-
+_DATAROW_METADATA_CREATE_ERROR = "Failed to add metadata on DataRow ID"
 
 class Client:
     """ A Labelbox client.
@@ -270,6 +270,8 @@ class Client:
 
             if get_error_status_code(internal_server_error) == 400:
                 raise labelbox.exceptions.InvalidQueryError(message)
+            elif _DATAROW_METADATA_CREATE_ERROR in message:
+                raise labelbox.exceptions.ResourceCreationError(message)
             else:
                 raise labelbox.exceptions.InternalServerError(message)
 
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -90,6 +90,7 @@ class Client:
             'Authorization': 'Bearer %s' % api_key,
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
+        self.data_row_metadata_ontology = None
 
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
@@ -648,7 +649,9 @@ class Client:
             DataRowMetadataOntology: The ontology for Data Row Metadata for an organization
 
         """
-        return DataRowMetadataOntology(self)
+        if self.data_row_metadata_ontology is None:
+            self.data_row_metadata_ontology = DataRowMetadataOntology(self)
+        return self.data_row_metadata_ontology
 
     def get_model(self, model_id) -> Model:
         """ Gets a single Model with the given ID.
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -39,6 +39,7 @@ logger = logging.getLogger(__name__)
 _LABELBOX_API_KEY = "LABELBOX_API_KEY"
 _DATAROW_METADATA_CREATE_ERROR = "Failed to add metadata on DataRow ID"
 
+
 class Client:
     """ A Labelbox client.
 
@@ -90,7 +91,7 @@ class Client:
             'Authorization': 'Bearer %s' % api_key,
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
-        self.data_row_metadata_ontology = None
+        self._data_row_metadata_ontology = None
 
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError))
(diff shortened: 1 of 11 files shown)


kkim-labelbox:
SDK build failing because the API changes need to go in first

labelbox/orm/model.py
```diff
@@ -91,8 +102,12 @@ def Enum(enum_cls: type, *args):
     def Json(*args):
         return Field(Field.Type.Json, *args)
 
+    @staticmethod
+    def List(list_cls: type, *args):
```
jtsodapop, line 106: noticing you modeled `List` similar to `Enum`

jw, but what is the reason why `Enum` and `List` (your new static method) are separate from the `Field.Type` enums?

kkim-labelbox: The List takes in a data type, whereas Field.Types can't do that. The new data type defines what kind of data is stored in the List, so it's mostly done to provide clarity to the reader. Currently, we're not doing much with that data type, but it'll be good to have this structure for transparency, and we may come back to do something with it in the future

labelbox/client.py
```diff
@@ -648,7 +649,9 @@ def get_data_row_metadata_ontology(self) -> DataRowMetadataOntology:
             DataRowMetadataOntology: The ontology for Data Row Metadata for an organization
 
         """
-        return DataRowMetadataOntology(self)
+        if self.data_row_metadata_ontology is None:
```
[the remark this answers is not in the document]
kkim-labelbox: Good point! Currently, the way customers would be using this object is, they'll call
```mdo = client.get_data_row_metadata_ontology()
mdo.fields```
and if they happen to change the metadata after, mdo.fields wouldn't return the most up-to-date metadata, because the `fields` is only set when `client.get_data_row_metadata_ontology()` is first called, via `DataRowMetadataOntology`'s `init`.

They would be able to update the `fields` only if they call `client.get_data_row_metadata_ontology()` again.

In order to support fetching of the latest metadata, I can make a change to make `refresh_ontology` public, so the customers have a way to refresh the metadata list. What do you think?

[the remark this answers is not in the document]
Comment (kkim-labelbox) [reply labelbox/client.py:652]: Good point! Currently, the way customers would be using this object is, they'll call
```
mdo = client.get_data_row_metadata_ontology()
mdo.fields
```
and if they happen to change the metadata after, mdo.fields wouldn't return the most up-to-date metadata, because the `fields` is only set when `client.get_data_row_metadata_ontology()` is first called, via `DataRowMetadataOntology`'s `init`.

They would be able to update the `fields` only if they call `client.get_data_row_metadata_ontology()` again.

In order to support fetching of the latest metadata, I can make a change to make `refresh_ontology` public, so the customers have a way to refresh the metadata list. What do you think?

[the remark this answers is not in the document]
Comment (jtsodapop) [reply labelbox/client.py:652]: I like having `refresh_ontology` as public!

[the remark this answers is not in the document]
Comment (kkim-labelbox) [reply labelbox/client.py:652]: Done!

File: labelbox/client.py [hunk 3]
```diff
@@ -648,7 +651,9 @@ def get_data_row_metadata_ontology(self) -> DataRowMetadataOntology:
...
```
msokoloff1, line 654: The client doesn't cache any other resources. It only exposes get and create functions. I think at the least, we should make the data_row_metadata_ontology attribute private by prepending an underscore so that users don't directly access the attribute.

Although I kind of feel like the DataRowMetadataOntology class should be a singleton. The client already does too much here. I'll leave that up to you if you want to change it.

labelbox/orm/query.py
```diff
@@ -41,7 +41,10 @@ def results_query_part(entity):
         entity (type): The entity which needs fetching.
     """
     # Query for fields
-    fields = [field.graphql_name for field in entity.fields()]
+    fields = [
+        field.graphql_name if field.graphql_name != "customMetadata" else
+        "customMetadata { value schemaId }" for field in entity.fields()
+    ]
```
msokoloff1, line 47: This isn't really maintainable. If we are making custom queries for classes, i'd like to move that logic to the DBObject class that requires the custom query. 

I'm think that a terrible alternative way to support this would be to override the field.graphql_name for the customMetadata field to be the "customMetadata { value schemaId }".

Instead of ^^ maybe we can add a new optional attribute to the field called like field.subquery or something. Then in the field definition we can set that as part of the class. In the list comprehension above we will use the subquery if it is set. E.g.

```
class DataRow:
    custom_metadata = Field.List(DataRowMetadataField, "custom_metadata", subquery ="customMetadata { value schemaId }" )
```
and 
```
 fields = [
        field.subquery if field.subquery is not None else field.graphql_name
    ]
```

labelbox/orm/model.py
```diff
@@ -54,6 +54,17 @@ def __init__(self, enum_cls: type):
...
+    class ListType:
+
+        def __init__(self, list_cls: type):
+            self.list_cls = list_cls
+
+        @property
+        def name(self):
+            if self.list_cls.__name__ == "DataRowMetadataField":
+                return "[DataRowCustomMetadataUpsertInput!]"
+            return self.list_cls.__name__
```
msokoloff1, line 66: Could this also just be a param of the ListType? This could be the param_name that is set in the field declaration in the DataRow class?

labelbox/orm/query.py
kkim-labelbox: Love this!

labelbox/orm/model.py
kkim-labelbox: Done!

msokoloff1 (approved):

PR #567, opened 2022-05-11
3.21.0

msokoloff1:


labelbox/schema/project.py
```diff
@@ -94,6 +97,15 @@ def update(self, **kwargs):
         if mode:
             self._update_queue_mode(mode)
 
+        media_type = kwargs.get("media_type")
```
jtsodapop, line 100: jw for the future, will we be mandating media type for project creation in the future? i see it as optional right now

jtsodapop (approved):

PR #568, opened 2022-05-11
[AL-2326] Metadata and Media Attributes for common BaseData object

jtsodapop:
* addition of media attributes and metadata for a Data  object

The addition of media attributes and metadata in the export now allow us to fetch these. this PR is an attempt to add the fields under the BaseData class which can then be looked at via common Labelbox format Data objects under a Label 

class `BaseData` now contains `media_attributes` and `metadata` fields.

* addition of media attributes and metadata for all previous test cases to make sure they match the new expected export format


jtsodapop:
Note: Need to update the test cases for inclusion of these two new fields.

kkim-labelbox (approved):
LGTM!

PR #569, opened 2022-05-12
[AL-2384] Increase create_batch timeout

farkob:
[JIRA ticket](https://labelbox.atlassian.net/browse/AL-2384)

Increases `Project.create_batch()` timeout, since for batch sizes > 10,000 thirty seconds usually isn't enough

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -617,7 +617,9 @@ class Project(DbObject, Updateable, Deletable):
             }
         }
 
-        res = self.client.execute(query_str, params,
+        res = self.client.execute(query_str,
+                                  params,
+                                  timeout=180.0,
                                   experimental=True)["project"][method]
 
         res['size'] = len(dr_ids)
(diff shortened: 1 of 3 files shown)


labelbox/schema/project.py
```diff
@@ -617,7 +617,9 @@ def create_batch(self, name: str, data_rows: List[str], priority: int = 5):
             }
         }
 
-        res = self.client.execute(query_str, params,
+        res = self.client.execute(query_str,
+                                  params,
+                                  timeout=180.0,
```
jtsodapop, line 622: Since this is change is only to accommodate `project.create_batch`, can we instead update the timeout for the specific method?

https://github.com/Labelbox/labelbox-python/blob/47063d7f1905f7df7f3696b3ff32ebfc6bc1ae19/labelbox/schema/project.py#L621

`res = self.client.execute(query_str, params,
                                  experimental=True, timeout=180)["project"][method]`

jtsodapop: Oh, I was misreading the PR. Sorry about that, lgtm!

jtsodapop (approved):

PR #574, opened 2022-05-17
[AL-2327] Update to logic for fetching metadata from DataRow exports

jtsodapop:
update to files to be able to handle pulling in metadata from datarows

previously, we had to accommodate metadata by filling it in with an empty list. once the following PR is approved, this should be checked and validated again, and should be able to correctly fetch metadata from areas where we export datarows vs. just the label

https://github.com/Labelbox/intelligence/pull/8565

### DIFF
--- a/labelbox/schema/dataset.py
+++ b/labelbox/schema/dataset.py
@@ -410,9 +410,8 @@ class Dataset(DbObject, Updateable, Deletable):
                 response.raise_for_status()
                 reader = ndjson.reader(StringIO(response.text))
                 # TODO: Update result to parse customMetadata when resolver returns
-                return (Entity.DataRow(self.client, {
-                    **result, 'customMetadata': []
-                }) for result in reader)
+                return (Entity.DataRow(self.client, {**result})
+                        for result in reader)
             elif res["status"] == "FAILED":
                 raise LabelboxError("Data row export failed.")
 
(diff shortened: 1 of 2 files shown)


kkim-labelbox (approved):
Nice!

labelbox/schema/dataset.py
```diff
@@ -410,9 +410,8 @@ def export_data_rows(self, timeout_seconds=120) -> Generator:
                 response.raise_for_status()
                 reader = ndjson.reader(StringIO(response.text))
                 # TODO: Update result to parse customMetadata when resolver returns
-                return (Entity.DataRow(self.client, {
-                    **result, 'customMetadata': []
-                }) for result in reader)
+                return (Entity.DataRow(self.client, {**result})
```
msokoloff1, line 413: This can just be 
```Entity.DataRow(self.client, result)```

`{**result}` is exactly the same as `result` it just makes a copy.

jtsodapop: Thanks, just updated

msokoloff1 (changes requested):
This is going to go behind a feature flag - something like bulk-export-metata which will effect labels and data row exports. (Based on PR feedback incase these queries don't perform well) so we want to check if data row metadata exists on creation of the entity object and default to an empty list.

jtsodapop:
Closing this out as there are some issues with merging the most recent dev. Reopening with new PR

PR #575, opened 2022-05-18
retry on ConnectTimeout errors

msokoloff1:
* We now retry the request when a ConnectionError is encountered
* Added tests to make sure that the ConnectionError is being retried
* Added a timeout arg to upload_file so that it doesn't hang indefinitely
* Modified the timeout for execute so that the connection timeout is short and the read timeout is long. This way retries on connection issues will happen quickly instead of waiting 30 seconds. 
------ 
* Error handling should be simplified. Also upload_file should also use execute but it is too much of a risk to change without a lot of testing. We should follow up in a separate PR.


msokoloff1:
waiting on some additional tests

PR #585, opened 2022-05-25
[AL-2290] Add result field to Task and get_result()

kkim-labelbox:
https://labelbox.atlassian.net/browse/AL-2290

### DIFF
--- a/labelbox/schema/task.py
+++ b/labelbox/schema/task.py
@@ -1,5 +1,8 @@
 import logging
+import requests
 import time
+import validators
+from validators import ValidationFailure
 from typing import TYPE_CHECKING, Optional
 
 from labelbox.exceptions import ResourceNotFoundError
@@ -31,6 +34,7 @@ class Task(DbObject):
     name = Field.String("name")
     status = Field.String("status")
     completion_percentage = Field.Float("completion_percentage")
+    result = Field.String("result")
     _user: Optional["User"] = None
 
     # Relationships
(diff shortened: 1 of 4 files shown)


labelbox/schema/task.py
```diff
@@ -31,6 +34,7 @@ class Task(DbObject):
     name = Field.String("name")
     status = Field.String("status")
     completion_percentage = Field.Float("completion_percentage")
+    result = Field.String("result")
```
msokoloff1, line 37: maybe call this result_url

kkim-labelbox: There's already a resolver called `result` off Task, so I think it's best to keep the naming consistent

labelbox/schema/task.py
```diff
@@ -65,3 +67,12 @@ def wait_till_done(self, timeout_seconds=300) -> None:
...
             self.refresh()
+
+    def get_result(self) -> str:
+        """ Downloads the result file from Task
+        """
+        if self.result:
+            response = requests.get(self.result)
+            response.raise_for_status()
+            return response.text
+        return ""
```
msokoloff1, line 78: it might be good to return None vs an empty string. That way an empty response is clearly different from a result that doesn't exist.

labelbox/schema/task.py
```diff
@@ -65,3 +67,13 @@ def wait_till_done(self, timeout_seconds=300) -> None:
...
             self.refresh()
+
+    def errors(self):
+        """ Downloads the result file from Task
+        """
+        if self.status == "FAILED" and self.result:
+            response = requests.get(self.result)
+            response.raise_for_status()
+            data = response.json()
+            return data['error']
```
msokoloff1, line 78: data.get('error')

PR #586, opened 2022-05-25
[AL-0] Limit create_data_rows() to 30k if datarows contain metadata

kkim-labelbox:


labelbox/schema/dataset.py
```diff
@@ -342,6 +342,16 @@ def convert_item(item):
                 f"Must pass an iterable to create_data_rows. Found {type(items)}"
             )
 
+        # TODO: If any datarows contain metadata, we're limiting max # of datarows
+        # until we address performance issues with datarow create with metadata
+        max_datarow_with_metadata = 30000
+        if (len(items) > max_datarow_with_metadata):
+            for row in items:
+                if row.get("metadata_fields"):
```
msokoloff1, line 350: if 'metadata_fields' in row is prob faster.

labelbox/schema/dataset.py
```diff
@@ -342,6 +342,16 @@ def convert_item(item):
                 f"Must pass an iterable to create_data_rows. Found {type(items)}"
             )
 
+        # TODO: If any datarows contain metadata, we're limiting max # of datarows
+        # until we address performance issues with datarow create with metadata
+        max_datarow_with_metadata = 30000
```
msokoloff1, line 347: nit: I would prefer to write as 30_000 to make it easier to read.

msokoloff1 (approved):

PR #593, opened 2022-06-03
improve result fetching experience

msokoloff1:
* Fetch the result file for the user
* Try to match AnnotationImport.statuses and .errors as close as possible.


labelbox/schema/task.py
```diff
@@ -68,12 +74,44 @@ def wait_till_done(self, timeout_seconds=300) -> None:
...
         """ Downloads the result file from Task
         """
-        if self.status == "FAILED" and self.result:
-            response = requests.get(self.result)
-            response.raise_for_status()
-            data = response.json()
-            return data.get('error')
+        if self.status == "FAILED":
+            result = self._fetch_remote_json()
+            return result['error']
```
kkim-labelbox, line 83: result.get('error')?

msokoloff1: I was thinking that this should never be None since we are specifically checking that the action is Json import.

labelbox/schema/task.py
```diff
@@ -68,12 +74,44 @@ def wait_till_done(self, timeout_seconds=300) -> None:
...
+        """
+        if self.name != 'JSON Import':
+            raise ValueError(
+                "Task result is only supported for `JSON Import` tasks."
+                " Download task.result_url manually to access the result for other tasks."
+            )
+        self.wait_till_done(timeout_seconds=600)
+        if self.status == "IN_PROGRESS":
+            raise ValueError(
+                "Job status still in `IN_PROGRESS`. The result is not available. Increase timeout or contact support."
```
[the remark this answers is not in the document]
msokoloff1: They can `wait_till_done` before calling this with a larger timeout. I'll update the message to make this clear. Also the cache does not save the result if an error is thrown.

jtsodapop (approved):

kkim-labelbox (approved):

PR #602, opened 2022-06-14
[AL-2619] Add 'metadata' field representing data row metadata as DataRowMetadat…

kkim-labelbox:
…aFields

- Add new field 'metadata' for data_row, which represents the serialized metadata fields off of a data row
- 'metadata' uses customMetadata resolver to retrieve a flattened view of metadata, and uses mdo.parse_metadata_fields to convert the dict --> DataRowMetadataField format
- Added test cases for them

### DIFF
--- a/labelbox/orm/db_object.py
+++ b/labelbox/orm/db_object.py
@@ -70,15 +70,15 @@ class DbObject(Entity):
                         "field %s", value, field)
             elif isinstance(field.field_type, Field.EnumType):
                 value = field.field_type.enum_cls(value)
-            elif isinstance(field.field_type,
-                            Field.ListType) and field.name == "metadata":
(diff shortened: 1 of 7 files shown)


labelbox/orm/db_object.py
```diff
@@ -70,6 +70,15 @@ def _set_field_values(self, field_values):
                         "field %s", value, field)
             elif isinstance(field.field_type, Field.EnumType):
                 value = field.field_type.enum_cls(value)
+            elif isinstance(field.field_type,
+                            Field.ListType) and field.name == "metadata":
```
msokoloff1, line 74: I would check the list type of the field instead of the field name.

msokoloff1:
Just one small nit. Otherwise looks good. I hate having to support metadataFields but I don't think that there is a better option.

PR #608, opened 2022-06-23
Fixing the notebook with the historical data available

paultancre:
Will make sure we are matching (the other basic tuto) on another pull : https://github.com/Labelbox/labelbox-python/blob/develop/examples/annotation_import/basics.ipynb


paulnoirel (approved):
LGTM

PR #609, opened 2022-06-23
Add example to list data rows from a batch

paulnoirel:
Add a simple example to show how to retrieve a list of data rows from a batch:

image


paulnoirel:
Need to republish due to SDK errors following numpy 1.23.0 release

PR #617, opened 2022-06-29
assign data row split

msokoloff1:
allows users to assign data rows to splits in diagnostics

### DIFF
--- a/labelbox/schema/model_run.py
+++ b/labelbox/schema/model_run.py
@@ -1,3 +1,4 @@
+# type: ignore
 from typing import TYPE_CHECKING, Dict, Iterable, Union, List, Optional, Any
 from pathlib import Path
 import os
--- a/labelbox/schema/model_run.py
+++ b/labelbox/schema/model_run.py
@@ -92,6 +92,7 @@ class ModelRun(DbObject):
 
     def _wait_until_done(self, status_fn, timeout_seconds=60, sleep_time=5):
         # Do not use this function outside of the scope of upsert_data_rows or upsert_labels. It could change.
+        original_timeout = timeout_seconds
         while True:
             res = status_fn()
             if res['status'] == 'COMPLETE':
@@ -102,7 +103,7 @@ class ModelRun(DbObject):
             timeout_seconds -= sleep_time
             if timeout_seconds <= 0:
                 raise TimeoutError(
-                    f"Unable to complete import within {timeout_seconds} seconds."
+                    f"Unable to complete import within {original_timeout} seconds."
                 )
 
             time.sleep(sleep_time)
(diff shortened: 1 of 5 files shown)


labelbox/schema/model_run.py
```diff
@@ -180,6 +181,39 @@ def delete_model_run_data_rows(self, data_row_ids):
             data_row_ids_param: data_row_ids
         })
 
+    @experimental
+    def assign_data_rows_to_split(self,
+                                  data_row_ids,
+                                  split,
+                                  timeout_seconds=60):
+        valid_splits = ["TRAINING", "TEST", "VALIDATION"]
```
jtsodapop, line 189: **Thought:** Would it make sense to have this as an enum of Dataset or ModelRun? Unlikely, but if we add more to the splits or use this elsewhere

jtsodapop (approved):
LGTM, left one comment but not really necessary

labelbox/schema/model_run.py
msokoloff1: It prob should be an enum. My thought was that 1) we aren't consistent with our usage of enums so this isn't inconsistent with our existing lack of a pattern and 2) if we require an enum it is another thing for users to import. I do want to chat about this with you and Kevin today if you have some time. Just more generally what we want to do.

jtsodapop: sgtm!

labelbox/schema/model_run.py
```diff
@@ -180,22 +195,62 @@ def delete_model_run_data_rows(self, data_row_ids):
...
+        if split_value == DataSplit.UNASSIGNED.value:
+            raise ValueError(
+                f"Cannot assign split value of `{DataSplit.UNASSIGNED.value}`.")
+
+        valid_splits = filter(lambda name: name != DataSplit.UNASSIGNED.value,
+                              DataSplit._member_names_)
+
+        if split_value not in valid_splits:
+            raise ValueError(
+                f"split must be one of : `{valid_splits}`. Found : `{split}`")
```
kkim-labelbox, line 215: Capitalize `Split`?

kkim-labelbox (commented):
A couple small nits. Otherwise, LGTM!

labelbox/schema/model_run.py
msokoloff1: It is a variable name. I'll put tickets around it maybe to highlight that.

labelbox/schema/model_run.py
```diff
@@ -180,22 +195,62 @@ def delete_model_run_data_rows(self, data_row_ids):
             data_row_ids_param: data_row_ids
         })
 
+    @experimental
+    def assign_data_rows_to_split(self,
+                                  data_row_ids: List[str],
+                                  split: Union[DataSplit, str],
+                                  timeout_seconds=60):
```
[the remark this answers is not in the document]
msokoloff1: will update

jtsodapop (approved):

kkim-labelbox (approved):

PR #627, opened 2022-07-13
[AL-2954] Add video point and polyline support in annotation serialization

kkim-labelbox:
- Added video point and video polyline support for serialization/deserialization, along with test cases
- fixed deserialization to pass in uuid for video annotations
- fixed infer_media_type()


jtsodapop (approved):
LGTM

PR #635, opened 2022-07-19
[AL-2896] Video annotation serialization/deserialization using segment_index

kkim-labelbox:
### Background
Currently, the only way to group video segments in SDK's `VideoObjectAnnotations` is to have an instance for every consecutive frame. This PR introduces a concept of segment_index, a new attribute in Video annotation classes, that can group annotation instances together into a video segment, without having to be consecutive frames.

### How does `segment_index` work?
- `segment_index` must be present in every `VideoObjectAnnotation` or `VideoClassificationAnnotation`. If only some annotations contain `segment_index`, error is raised
- If none of the annotations contain `segment_index`, segments are evaluated using consecutive frames, same as the original logic
- Every annotation belonging to the same `segment_index` will be grouped within the same video segment

### Heads up
- Both serialization and deserialization are updated. User can deserialize NDJson with multiple keyframes in a segment into VideoObjectAnnotations, which will automatically assign `segment_index`. User can also serialize VideoObjectAnnotations with `segment_index`, which will properly evaluate frames within the same segment. I updated the test cases to test this
- `VideoClassificationAnnotations` is not affected by this change much, because classification answer change using `VideoClassificationAnnotations` seems to not be supported. Will tackle this in another ticket


labelbox/data/serialization/ndjson/label.py
```diff
@@ -88,6 +88,39 @@ def _get_consecutive_frames(
...
+        sorted_frame_segment_indices = sorted([
+            (annotation.frame, annotation.segment_index)
+            for annotation in annotation_group
+            if annotation.segment_index is not None
+        ])
+        if len(sorted_frame_segment_indices) == 0:
+            # Group segment by consecutive frames, since `segment_index` is not present
+            return cls._get_consecutive_frames(
+                sorted([annotation.frame for annotation in annotation_group]))
+        elif len(sorted_frame_segment_indices) == len(annotation_group):
```
jtsodapop, line 105: it's been a while since i have reviewed `annotation_group`, could you explain why this check is `== len(annotation_group)` ? Is it because we expect the number of segments to be the number of items in the annotation group?

labelbox/data/serialization/ndjson/label.py
```diff
@@ -88,6 +88,39 @@ def _get_consecutive_frames(
...
+            # Group segment by segment_index
+            last_segment_id = 0
+            segment_groups = defaultdict(list)
+            for frame, segment_index in sorted_frame_segment_indices:
+                if segment_index < last_segment_id:
+                    raise ValueError(
```
jtsodapop, line 111: **Question** In what instances do we expect this error to occur, since we are fetching indices via enumerate and then sorting? Should we ever expect this to be a problem?

labelbox/data/serialization/ndjson/objects.py
```diff
@@ -235,11 +240,12 @@ class NDSegments(NDBaseObject):
 
     def to_common(self, name: str, feature_schema_id: Cuid):
         result = []
-        for segment in self.segments:
+        for idx, segment in enumerate(self.segments):
```
jtsodapop, line 243: **Comment** This is a really neat idea :)

labelbox/data/serialization/ndjson/label.py
```diff
@@ -88,6 +88,39 @@ def _get_consecutive_frames(
...
+                        f"`segment_index` must be in ascending order. Please investigate video annotation at frame, '{frame}'"
+                    )
+                segment_groups[segment_index].append(frame)
+                last_segment_id = segment_index
+            frame_ranges = []
+            for group in segment_groups.values():
+                frame_ranges.append((group[0], group[-1]))
+            return frame_ranges
+        else:
```
jtsodapop, line 120: **Question** When would we expect only partial indices to occur?

labelbox/data/serialization/ndjson/label.py
kkim-labelbox: Yeah! So this is just checking that every annotation inside `annotation_group` contains a `segment_index`.
```
segment_index must be present in every VideoObjectAnnotation or VideoClassificationAnnotation. If only some annotations contain segment_index, error is raised
```

labelbox/data/serialization/ndjson/label.py
kkim-labelbox: This can be an issue if users construct their own Video annotations, and accidentally assign segment_indices in a non-ascending order

labelbox/data/serialization/ndjson/label.py
kkim-labelbox: If users are providing their own Video annotation objects!

PR #642, opened 2022-07-21
[AL-3013] Deprecation warning for assign_feature_schema_ids

jtsodapop:


labelbox/data/annotation_types/label.py
```diff
@@ -138,6 +139,9 @@ def assign_feature_schema_ids(
 
         Note: You can now import annotations using names directly without having to lookup schema_ids
         """
+        warnings.warn("This method is deprecated and will be "
```
kkim-labelbox, line 142: Looks like there's one more `assign_feature_schema_ids` in this file under `LabelGenerator`

jtsodapop: Thanks for catching!

kkim-labelbox (approved):

PR #644, opened 2022-07-21
[AL-2897] Batch management UI parity

farkob:
- Replaces `Batch.remove_queued_data_rows()` method endpoint with newer version.
- Adds `Batch.delete()` method
- Adds `Batch.delete_labels(labels_as_template=False)` method


labelbox/schema/batch.py
```diff
@@ -121,3 +122,49 @@ def export_data_rows(self, timeout_seconds=120) -> Generator:
...
+        self.client.execute("""mutation DeleteBatchPyApi($%s: ID!, $%s: ID!) {
+            project(where: {id: $%s}) { deleteBatch(batchId: $%s) { deletedBatchId } }
+        }""" % (project_id_param, batch_id_param, project_id_param,
+                batch_id_param), {
+                    project_id_param: self.project_id,
+                    batch_id_param: self.uid
+                },
+                            experimental=True)
+
+    def delete_labels(self, labels_as_template=False) -> None:
```
cssetian, line 146: maybe better to call the flag set_labels_as_template, otherwise could be confusing what operation is being performed with them

tests/integration/test_batch.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
 from labelbox import Dataset, Project
+import labelbox
```
jtsodapop, line 4: **suggestion** can be omitted

tests/integration/test_batch.py
```diff
@@ -50,6 +51,16 @@ def test_archive_batch(configured_project: Project, small_dataset: Dataset):
     assert len(exported_data_rows) == 0
 
 
+def test_delete(configured_project: Project, small_dataset: Dataset):
+    data_rows = [dr.uid for dr in list(small_dataset.export_data_rows())]
+    configured_project.update(queue_mode=Project.QueueMode.Batch)
+    batch = configured_project.create_batch("batch to delete", data_rows)
+    batch.delete()
+
+    with pytest.raises(Exception):
```
jtsodapop, line 60: I think we can omit this section of the test

tests/integration/test_batch.py
```diff
@@ -80,3 +89,29 @@ def test_export_data_rows(configured_project: Project, dataset: Dataset):
 
     assert len(result) == n_data_rows
     assert set(data_rows) == set(exported_data_rows)
+
+
+@pytest.mark.skip(
+    reason="Test cannot be used effectively with MAL/LabelImport. \
```
kkim-labelbox, line 95: Can you investigate the reason behind this skip, and make sure that we have this fix lined up as a jira ticket?

kkim-labelbox (approved):
LGTM! Just one comment on the skipped tests

tests/integration/test_batch.py
jtsodapop: This is likely interaction between `Batch` vs what we expect from `Dataset` since they work so differently. We're seeing a similar issue when trying to even create labels in a batch project vs a dataset project. I'll create a follow up ticket for this as a future SDK incidental

jtsodapop: Created polish ticket here. will be pushing forward with these changes
https://labelbox.atlassian.net/browse/AL-3261

PR #647, opened 2022-07-26
[AL-2996] Fix metadata uploads with local file datarow upload

kkim-labelbox:
- Fixed metadata not uploading when data row is a local file
- Added test cases


labelbox/schema/dataset.py
```diff
@@ -260,11 +260,10 @@ def upload_if_necessary(item):
             row_data = item['row_data']
             if os.path.exists(row_data):
                 item_url = self.client.upload_file(item['row_data'])
```
jtsodapop, line 262: **suggestion** can we change this `item_url = self.client.upload_file(item['row_data'])` to this?

`item_url = self.client.upload_file(row_data)`

labelbox/schema/dataset.py
```diff
@@ -260,11 +260,10 @@ def upload_if_necessary(item):
...
-                item = {
-                    "row_data": item_url,
-                    "external_id": item.get('external_id', item['row_data']),
-                    "attachments": item.get('attachments', [])
-                }
+                item['row_data'] = item_url
+                if 'external_id' not in item:
+                    # Default `external_id` to local file name
+                    item['external_id'] = row_data
```
msokoloff1, line 266: How is this different?

kkim-labelbox: it uses the original `item`, and alters it, rather than constructing a brand new `item`

jtsodapop, line 266: **question** should we include back attachments?

labelbox/schema/dataset.py
```diff
@@ -260,11 +260,10 @@ def upload_if_necessary(item):
             row_data = item['row_data']
             if os.path.exists(row_data):
                 item_url = self.client.upload_file(item['row_data'])
-                item = {
-                    "row_data": item_url,
-                    "external_id": item.get('external_id', item['row_data']),
-                    "attachments": item.get('attachments', [])
-                }
+                item['row_data'] = item_url
+                if 'external_id' not in item:
```
jtsodapop, line 264: **question** im a little confused on how this resolved the issue. from the previous code, wouldn't it also do the same?

`"external_id": item.get('external_id', item['row_data']),`

labelbox/schema/dataset.py
kkim-labelbox: If attachments are included in the original `item`, they will remain

labelbox/schema/dataset.py
jtsodapop: resolving convo as this was asked in another comment

labelbox/schema/dataset.py
jtsodapop: resolving as i now understand attachments will already be part of existing item

jtsodapop (approved):

labelbox/schema/dataset.py
msokoloff1: What is the need for this?

msokoloff1: oh sorry, I should have read the description.

msokoloff1 (approved):

labelbox/schema/dataset.py
kkim-labelbox: No worries, it's a subtle change haha

PR #648, opened 2022-07-26
[AL-0] Prepping 3.25.2

kkim-labelbox:


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog
 
+# Version 3.25.2 (2022-07-26)
+## Updated
+* Notebooks are up-to-date with examples of importing annotations without `schema_id`
```
jtsodapop, line 5: **question**  jw, is this something we want to mention in changelogs for updates or something we leave more as hidden?

kkim-labelbox: I see some past updates including notebook updates, so I think it should be fine. @msokoloff1 what do you think?

msokoloff1: I think it is good to keep a record of any changes that are made. Maybe we can specifically call this out in a documentation section so that users know this change doesn't impact the functionality.

msokoloff1 (approved):

CHANGELOG.md
kkim-labelbox: Added documentation section!

PR #651, opened 2022-07-27
Fix timeout logic

corynezinstitchfix:
The current logic subtracts `sleep_time` every iteration regardless of whether the request actually took that long.  This change ensures that `timeout_seconds` seconds pass before returning `None`


msokoloff1 (approved):
Thank you!

PR #656, opened 2022-08-01
Add model run config on create [AL-3060]

mmarinwilcke:
Add initial metadata config on the model run when creating it


tomislav-peharda:
Hey, I don't have plenty of context on here, so would rather have someone else give a stamp / review

labelbox/schema/model_run.py
```diff
@@ -32,6 +32,7 @@ class ModelRun(DbObject):
     created_at = Field.DateTime("created_at")
     created_by_id = Field.String("created_by_id", "createdBy")
     model_id = Field.String("model_id")
+    training_metadata = Field.Json("training_metadata")
```
kkim-labelbox, line 35: FYI, adding this field here means that graphql query will now expect `trainingMetadata` from the resolver response. This may break until the backend is updated to include this field

labelbox/schema/model.py
```diff
@@ -18,23 +18,26 @@ class Model(DbObject):
     name = Field.String("name")
     model_runs = Relationship.ToMany("ModelRun", False)
 
-    def create_model_run(self, name) -> "ModelRun":
+    def create_model_run(self, name, config) -> "ModelRun":
```
kkim-labelbox, line 21: Adding a mandatory config here will be a breaking change for existing customer scripts that call this function. Better to set a default value of None here, and handle that accordingly
```suggestion
    def create_model_run(self, name, config=None) -> "ModelRun":
```

kkim-labelbox:
Looks like the build is failing because your code is stale. It'll pass after you pull the latest changes

labelbox/schema/model_run.py
mmarinwilcke: Yeah, the resolver has already been updated with this optional field!

jtsodapop (approved):
LGTM

labelbox/schema/model.py
```diff
@@ -18,23 +18,26 @@ class Model(DbObject):
...
-            createModelRun(data: {name: $%s, modelId: $%s}) {%s}}""" % (
-            name_param, model_id_param, name_param, model_id_param,
-            query.results_query_part(ModelRun))
+        query_str = """mutation CreateModelRunPyApi($%s: String!, $%s: Json, $%s: ID!) {
+            createModelRun(data: {name: $%s, trainingMetadata: $%s, modelId: $%s}) {%s}}""" % (
+            name_param, config_param, model_id_param, name_param, config_param,
+            model_id_param, query.results_query_part(ModelRun))
         res = self.client.execute(query_str, {
             name_param: name,
+            config_param: config,
```
kkim-labelbox, line 40: Just curious, does this work okay when the param is None?

mmarinwilcke: Hm, let me test that

kkim-labelbox: While you're at it, could you add a test case for `config` being empty as well?

mmarinwilcke: yeah, thats what I meant! haha

mmarinwilcke: All good

PR #657, opened 2022-08-04
[AL-0] Picsum Alternative

jtsodapop:
8/3/2022 - Using IMG_URL=picsum... is giving an error `invalid magic bytes for file`

This is circumvented by using alternative websites. This PR is to use an alternative lorem image generator.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -5,6 +5,7 @@ import time
 import uuid
 from enum import Enum
 from types import SimpleNamespace
+import random
 
 import pytest
 import requests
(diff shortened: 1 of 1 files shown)


jtsodapop:
These tests will still fail on test_benchmark due to some other test issues but should pass `test_create_batch` which is where they were originally failing that brought up this problem (the first test that would use `tests/integration` which would cause `conftest.py` to generate a `dataset` fixture)

tests/integration/conftest.py
```diff
@@ -317,6 +319,7 @@ def create_label():
     project.create_label = create_label
     project.create_label()
     label = next(project.labels())
+    print(f"this is a label: \n{label}")
```
kkim-labelbox, line 322: Remove this

kkim-labelbox (commented):
LGTM! i'll approve once you remove the print statement

PR #658, opened 2022-08-05
[AL-0] Picsum Alternative - hydak

hydak:
Duplicating the PR https://github.com/Labelbox/labelbox-python/pull/657 but without the print statement since JTSO is out.


andre-labelbox (approved):
LGTM

PR #660, opened 2022-08-06
[CCV-1501] Add two new attachment types - raw_text and text_file

lgluszek:
This change is allowing us to use Python SDK to set new attachment types for text - RAW_TEXT and TEXT_FILE.

More documentation about this change is available [here](https://github.com/Labelbox/intelligence/pull/9837)

### DIFF
--- a/labelbox/schema/asset_attachment.py
+++ b/labelbox/schema/asset_attachment.py
@@ -9,7 +9,7 @@ class AssetAttachment(DbObject):
     """ Asset attachment provides extra context about an asset while labeling.
 
     Attributes:
-        attachment_type (str): IMAGE, VIDEO, TEXT, IMAGE_OVERLAY, or HTML
+        attachment_type (str): IMAGE, VIDEO, TEXT, IMAGE_OVERLAY, HTML, RAW_TEXT, or TEXT_FILE
         attachment_value (str): URL to an external file or a string of text
     """
 
@@ -19,6 +19,8 @@ class AssetAttachment(DbObject):
         TEXT = "TEXT"
         IMAGE_OVERLAY = "IMAGE_OVERLAY"
         HTML = "HTML"
+        RAW_TEXT = "RAW_TEXT"
+        TEXT_FILE = "TEXT_FILE"
 
     for topic in AttachmentType:
         vars()[topic.name] = topic.value
--- a/labelbox/schema/asset_attachment.py
+++ b/labelbox/schema/asset_attachment.py
@@ -9,7 +9,7 @@ class AssetAttachment(DbObject):
     """ Asset attachment provides extra context about an asset while labeling.
 
     Attributes:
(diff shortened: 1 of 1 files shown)


labelbox/schema/asset_attachment.py
```diff
@@ -19,6 +19,8 @@ class AttachmentType(Enum):
         TEXT = "TEXT"
```
msokoloff1, line 19: Should we add a deprecation warning or at least a comment here?

lgluszek: I'll create a new ticket to remove this attachment type and in the description I'll specify the exact place where it needs to be removed.

When it comes to warning, maybe we can do it in a follow-up ticket?

lgluszek: added a warning https://github.com/Labelbox/labelbox-python/pull/660/commits/db2a6227a5600420637c98f603c7706f28781988

labelbox/schema/asset_attachment.py
```diff
@@ -19,6 +19,8 @@ class AttachmentType(Enum):
...
         IMAGE_OVERLAY = "IMAGE_OVERLAY"
         HTML = "HTML"
+        RAW_TEXT = "RAW_TEXT"
+        TEXT_FILE = "TEXT_FILE"
```
lukeulrich, line 23: Two thoughts:
1. What about calling this `TEXT_URL` - as that better reflects what users will be providing?
2. Unfortunately, having to qualify with `_FILE` now makes this one inconsistent with the other values, which all expect a URL but don't have a suffix `_URL`. I don't suppose there is a good workaround, besides eventually making all of the others suffixed with `_URL` (or similar). <sigh> Not requesting any change just wondering if you have any thoughts on this.

lukeulrich (approved):

msokoloff1 (approved):

msokoloff1:
It might be good to add these to the tests
https://github.com/Labelbox/labelbox-python/blob/8c401d45a5a0d456d7db6b2b8e5793655ca140e1/tests/integration/test_data_rows.py#L483
https://github.com/Labelbox/labelbox-python/blob/8c401d45a5a0d456d7db6b2b8e5793655ca140e1/tests/integration/test_data_rows.py#L514
https://github.com/Labelbox/labelbox-python/blob/8c401d45a5a0d456d7db6b2b8e5793655ca140e1/tests/integration/test_data_rows.py#L547

lgluszek:
> It might be good to add these new types to the tests
> 
> https://github.com/Labelbox/labelbox-python/blob/8c401d45a5a0d456d7db6b2b8e5793655ca140e1/tests/integration/test_data_rows.py#L483
> 
> 
> https://github.com/Labelbox/labelbox-python/blob/8c401d45a5a0d456d7db6b2b8e5793655ca140e1/tests/integration/test_data_rows.py#L514
> 
> 
> https://github.com/Labelbox/labelbox-python/blob/8c401d45a5a0d456d7db6b2b8e5793655ca140e1/tests/integration/test_data_rows.py#L547

Thanks a lot, I'll add these tests in a follow-up

PR #661, opened 2022-08-08
[AL-0] Update to bulk create data row test

jtsodapop:
`Dataset.data_rows()` is running slow. Because the test being modified is not explicitly testing the effectiveness of the above, and because `Dataset.export_data_rows()` has been recently updated and optimized, we will be using this in the test moving forward.


jtsodapop:
Closing out in favor of just updating the resolvers to be faster on the backend

PR #662, opened 2022-08-08
[AL-3169] Update to shapely geoms

jtsodapop:
Iterating over multipart in shapely can now be done using `object.geoms` which returns a list.

Code is updated so that we now iterate over the list result of the `object`

Current documentation from shapely:

`...Members of a multi-point collection are accessed via the geoms property or via the iterator protocol using in or list()...`
[link to shapely manual](https://shapely.readthedocs.io/en/stable/manual.html)

```
pprint.pprint(list(points.geoms))
[<shapely.geometry.point.Point object at 0x...>,
 <shapely.geometry.point.Point object at 0x...>]
```

Since we are expecting everything to be multicollection objects outside of just a polygon, we run `.geoms` against anything that is not polygon here.


kkim-labelbox (approved):
LGTM

PR #669, opened 2022-08-10
Update config [AL-3061]

mmarinwilcke:
Adds methods/tests to:
- Update model run config
- Reset model run config
- Fetch model run config

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-# Version 3.26.1 (2022-08-18)
+# Version 0.0.0 (YYYY-MM-DD)
 ### Added
 * `ModelRun.update_config()`
     * Updates model run training metadata
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,6 @@
 # Changelog
 
-# Version 3.26.0 (2022-08-15)
-## Added
-* `Batch.delete()` which will delete an existing `Batch`
-* `Batch.delete_labels()`  which will delete all `Label`’s created after a `Project`’s mode has been set to batch.
-    * Note: Does not include labels that were imported via model-assisted labeling or label imports
-* Support for creating model config when creating a model run
-* `RAW_TEXT` and `TEXT_FILE` attachment types to replace the `TEXT` type.
-
-# Version 3.25.3 (2022-08-10)
-
+# Version 3.26.1 (2022-08-18)
 ### Added
 * `ModelRun.update_config()`
     * Updates model run training metadata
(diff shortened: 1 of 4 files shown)


tests/integration/annotation_import/test_model_run.py
```diff
@@ -56,6 +56,24 @@ def test_model_run_delete(client, model_run):
...
 
+def test_model_run_update_config(model_run_with_training_metadata):
+    new_config = {"batch_size": 2000}
+    res = model_run_with_training_metadata.update_config(new_config)
+    assert res["trainingMetadata"]["batchSize"] == new_config["batch_size"]
+
+
+def test_model_run_reset_config(model_run_with_training_metadata):
+    res = model_run_with_training_metadata.reset_config()
+    assert res["trainingMetadata"] == None
```
msokoloff1, line 67: nit: When you check if an object value is None you should use the identify check with the `is` operator. Technically the equality operator of whatever res["trainingMetadata"] returns can be defined by that particular class. We don't care if the object is equal to None by its own definition, we want to know if the object is None.

jtsodapop (changes requested):
Nice tests! Left some comments about the changelog and to add some additional docstrings

labelbox/schema/model_run.py
```diff
@@ -262,6 +262,39 @@ def update_status(self,
             },
             experimental=True)
 
+    @experimental
+    def update_config(self, config: Dict[str, Any]):
```
kkim-labelbox, line 266: Could you specify the return type for all the new methods?

labelbox/schema/model_run.py
```diff
@@ -262,6 +262,39 @@ def update_status(self,
...
+        res = self.client.execute(
+            """mutation resetModelRunConfigPyApi($modelRunId: ID!){
+                resetModelRunConfig(modelRun: {id : $modelRunId}){trainingMetadata}
+            }
+        """, {'modelRunId': self.uid},
+            experimental=True)
+        return res["resetModelRunConfig"]
+
+    @experimental
+    def fetch_config(self):
```
kkim-labelbox, line 290: Suggestion: `get_config` would follow the general naming convention of SDK methods better, unless the PM has already signed off on `fetch_config`

CHANGELOG.md
```diff
@@ -9,6 +9,19 @@
 * `RAW_TEXT` and `TEXT_FILE` attachment types to replace the `TEXT` type.
 
 # Version 3.25.3 (2022-08-10)
+
+### Added
+* `ModelRun.update_config()`
```
kkim-labelbox, line 14: The changelogs for these functions should go in a new section above version 3.26.0. We are now modifying changelogs each commit, and best practice is to add a new placeholder version at the top, like so:
```
# Version 0.0.0 (YYYY-MM-DD)
## Added
* `ModelRun.update_config()`
...
```

labelbox/schema/model_run.py
```diff
@@ -262,6 +262,56 @@ def update_status(self,
             },
             experimental=True)
 
+    @experimental
+    def update_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
+        """ 
+         Updates the model run's training metadata config
```
kkim-labelbox, line 268: Sorry to be a stickler, but could we capitalize `Model Run` everywhere in the comments? Just wanna make sure the docs are formalized, since they'll be public facing. Thanks!

kkim-labelbox (approved):
LGTM!

jtsodapop (approved):

PR #673, opened 2022-08-17
[AL-3329] Document Annotation Type for Export

jtsodapop:
This PR is designed to:
* create a new annotation type class for documents which will test against if an export annotation contains `page` and `unit`
* Add this to `extra` field when it exists, else it passes back `None`

Rough example output:

```
name='boxy' 
feature_schema_id='cl6xnuwt95lqq07330tbb3mfd' 
extra={'instanceURI': 'https://api.labelbox.com/masks/feature/cl6xnzjpq0dmx07yo0qh40z0n?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VySWQiOiJja2NjOWZtbXc0aGNkMDczOHFpeWM2YW54Iiwib3JnYW5pemF0aW9uSWQiOiJja2N6NmJ1YnVkeWZpMDg1NW8xZHQxZzlzIiwiaWF0IjoxNjYwNzYxNDUwLCJleHAiOjE2NjMzNTM0NTB9.X4-j6zee8o685PUrL9C6oC2m6TayKuJQHhN8iLgG8kI', 
'color': '#1CE6FF', 
'feature_id': 'cl6xnzjpq0dmx07yo0qh40z0n', 
'value': 'boxy', 
'page': 3, 
'unit': 'POINTS'} 
value=Rectangle(extra={}, start=Point(extra={}, x=83.34, y=298.12), end=Point(extra={}, x=83.72, y=501.95000000000005)) classifications=[] 
```


kkim-labelbox (commented):
Some questions around this:
1. What's the difference between labelbox_v1 representations of annotations versus stuff inside `serialization/ndjson` ?
2. Do we need a new rectangle object type specific to document? I don't see this pattern used for other data types, e.g. Video or Images

jtsodapop:
> Some questions around this:
> 
> 1. What's the difference between labelbox_v1 representations of annotations versus stuff inside `serialization/ndjson` ?
> 2. Do we need a new rectangle object type specific to document? I don't see this pattern used for other data types, e.g. Video or Images

not sure if you receive an update when I make a reply, so re-tagging @kkim-labelbox ! 

1. labelbox_v1 representation is just the transformations between the labelbox export and the annotations themselves. the ndjson converter is for transformations between the annotations and a labelbox import format (which will be done in a separate PR)
2. there will be nothing different between the existing bounding box annotation for other data types such as image, and this one. the only difference is the `page` and `unit` keys from the export. 

i set this up this way since i thought it would make more sense to just re-use the LBV1Rectangle code as there are no required additional changes, but if it would make more sense to paste the same code in for this new class, i am open to it

labelbox/data/serialization/labelbox_v1/objects.py
```diff
@@ -259,33 +259,40 @@ def from_common(cls, text_entity: TextEntity,
                    **extra)
 
 
+class LBV1DocumentRectangle(LBV1Rectangle):
```
kkim-labelbox, line 262: Starting a thread here in response to your reply, so we can better keep track of the conversation.

It doesn't seem like good practice to create a new Data type specific entity here, as it doesn't follow the current pattern. These geometric classes only hold data about the dimensions of the annotations, and not dimensions of the data types like Document. If we follow this pattern, we'd be creating a Document class for every type of annotation. I'd suggest either putting them in the base class `LBV1ObjectBase` with default values set to None (which seems like the current pattern, since `LBV1ObjectBase` contains optional fields like `color` and `instanceURI`), although this isn't ideal either, since we'd again, be storing data type specific attributes in a annotation related object. Another solution could be to create a new superclass of `LBV1LabelAnnotations` for documents specifically, which would follow the same pattern as video, and has its own `to_common` and `from_common` methods. @msokoloff1, would love your input here as well.

kkim-labelbox: Also, was there a reason why only `to_common` was implemented? Does `from_common` not need any code changes?

jtsodapop: Sounds good. One option I was looking at earlier was adding it under the [feature class here](https://github.com/Labelbox/labelbox-python/blob/73e2336fd06a4627c1054088295265dfe7e53768/labelbox/data/serialization/labelbox_v1/feature.py#L9). It houses `keyframe` but thought that it didn't necessarily make sense, and agree it would make more sense under `LBV1ObjectBase` 

No reason why only `to_common` was implemented. This was just up for discussion so I wanted a better agreement on where it should land for design. I'll implement `from_common` as well

jtsodapop: Actually, what is convenient is by adding it with `to_common` in the extras field, `from_common` doesn't need any update work at all 😄

labelbox/data/serialization/labelbox_v1/label.py
```diff
@@ -146,6 +146,7 @@ class LBV1Label(BaseModel):
     skipped: Optional[bool] = Extra('Skipped')
     media_type: Optional[str] = Extra('media_type')
     data_split: Optional[str] = Extra('Data Split')
+    global_key: Optional[str] = Extra('Global Key')
```
jtsodapop, line 149: @kkim-labelbox added this into labels class. also recognize that many old tests did not include this in the past (part of the conversation about ever changing interfaces).

kkim-labelbox (approved):
LGTM!

jtsodapop:
Will update changelog once import side has also been complete!

PR #674, opened 2022-08-19
[AL-3330] NDJsonConverter support for PDF Assets

jtsodapop:
# Changes

Allow users to use `NDJsonConverter.serialize()` and `NDJsonConverter.deserialize()` to transform from Labelbox common format to an importable format

Tested using the following script on a project that has existing pdf document labels

```
project = client.get_project("PROJECT_ID")
labels = project.label_generator()
labels = NDJsonConverter.serialize(labels)
labels = NDJsonConverter.deserialize(labels)
labels = NDJsonConverter.serialize(labels)
import_annotations = MALPredictionImport.create_from_objects(client=client, project_id = project.uid, name=f"import {str(uuid.uuid4())}", predictions=labels)
import_annotations.wait_until_done()

print(import_annotations.errors)
```

results in: 
``` [] ```

### Heads Up
* Only supports bbox annotations

### CODE REVIEW

File: tests/data/serialization/ndjson/test_classification.py [hunk 1]
```diff
@@ -4,13 +4,18 @@
 from labelbox.data.serialization.ndjson.converter import NDJsonConverter
 from labelbox.data.serialization.ndjson.objects import NDLine
 
+IGNORE_KEYS = ['unit', 'page']
```
kkim-labelbox, line 7: Instead of popping these keys off of the serialized result each time, can we remove these keys from the output of `serialize` if the value is None, at least for page and unit? I feel that it'd be confusing to users if they started seeing these fields set to None for any other type of data. 

Maybe play around with `exclude_none` setting for the pydantic `model.dict()` function?

jtsodapop: Updated the `serialize` logic to pop if the value is `None`. Opted for this instead of setting `exclude_none` because it would affect some other fields that we may not necessarily want to exclude (one example is that it would pop off feature schema in some scenarios). If we want to opt for using this field in the future, it will likely require some additional effort

tests/data/serialization/ndjson/test_export_video_objects.py
```diff
@@ -585,6 +585,10 @@ def video_serialized_bbox_label():
     }
 
 
+#ignore uuid because we randomize if there was none
```
kkim-labelbox, line 588: For `test_export_video_objects.py` and test_image.py`, was there a reason why the optimizations were put in place? Now that we don't need to ignore "page" and "unit", I don't think we need to create a separate `IGNORE_KEYS` list just to hold one item, and it's creating unnecessary commit noise for these lines of code. I commend the efforts to make things more future-proof though!

tests/integration/test_project.py
```diff
@@ -183,7 +183,7 @@ def test_same_ontology_after_instructions(
         configured_project_with_complex_ontology):
     project, _ = configured_project_with_complex_ontology
     initial_ontology = project.ontology().normalized
-    project.upsert_instructions('tests/data/assets/loremipsum.pdf')
+    project.upsert_instructions('tests/assets/loremipsum.pdf')
```
jtsodapop, line 186: note: this is just because i changed the path of the pdf so it isn't nested under the `data` folder and is more general

labelbox/data/serialization/ndjson/objects.py
```diff
@@ -21,6 +21,8 @@
 
 class NDBaseObject(NDAnnotation):
     classifications: List[NDSubclassificationType] = []
+    page: Optional[int] = None
```
kkim-labelbox, line 24: Wonder if these should be inside NDAnnotation. If we support PDFs for classifications in the future, we'd need to store the page/unit info inside the classification annotation

kkim-labelbox (approved):
LGTM!

jtsodapop:
Made some additional changes to the monorepo to allow page = 0 when doing more validation work. Once those push through, will do an additional push on these sdk tests. 

The sdk tests are expected to fail on the new test until the new monorepo changes are pushed in.

PR #676, opened 2022-08-22
[QQC-362] Document new project update restrictions regarding quality …

richardsun0713:
…mode

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -97,17 +97,17 @@ class Project(DbObject, Updateable, Deletable):
 
         Args:
             kwargs: a dictionary containing attributes to be upserted
-                * Note that the quality setting cannot be changed after a project has been created. The quality mode
-                for a project is inferred through the following attributes:
-                Benchmark:
-                    auto_audit_number_of_labels = 1
-                    auto_audit_percentage = 1.0
-                Consensus:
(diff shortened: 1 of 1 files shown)


labelbox/schema/project.py
```diff
@@ -93,6 +93,22 @@ class QueueMode(Enum):
         Dataset = "Dataset"
 
     def update(self, **kwargs):
+        """ Updates this project with the specified attributes
+
+        Args:
+            kwargs: a dictionary containing attributes to be upserted
+                * Note that the quality setting cannot be changed after a project has been created. The quality mode
```
kkim-labelbox, line 100: Not sure if github is screwing up the formatting, but could you put the Note on the same indent level as Args?

jtsodapop: **Question** When is project.update() meant to be used for `auto_audit_number_of_labels` and `auto_audit_percentage` if changing these values is supposed to be an illegal operation? If users cannot change between the two at all, should we jus exclude this information from the docstring?

richardsun0713: Users can adjust these values when a project is using consensus mode. For example, they might change the number of labels from 2 to 3, and the project's quality setting would remain in consensus mode (thus, a valid operation).

kkim-labelbox (approved):
LGTM!

PR #677, opened 2022-08-23
Modify get_config return value [AL-3352]

mmarinwilcke:
- We want to un-nest the return value for get_config

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 # Version 0.0.0 (YYYY-MM-DD)
+### Changed
+* `ModelRun.get_config()`
+    * Modifies get_config to return un-nested Model Run config
 ### Added
 * `ModelRun.update_config()`
     * Updates model run training metadata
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 
 # Version 0.0.0 (YYYY-MM-DD)
+### Changed
+* `ModelRun.get_config()`
+    * Modifies get_config to return un-nested Model Run config
```
jtsodapop, line 6: Maybe change the wording to something like:

`get_config` now returns un-nested Model Run config

also question, what is nested vs. un-nested?

mmarinwilcke: hm, cause basically we were returning the config object under the key 'trainingData' in the return object. Karen asked to just return the config object directly.

jtsodapop: Gotcha, i'm ok with these changes then

jtsodapop (approved):

kkim-labelbox (approved):

PR #682, opened 2022-09-03
[Al-3100] Added global key support to bulk import

olegtrygub:
https://labelbox.atlassian.net/browse/AL-3100


kkim-labelbox (approved):
LGTM! Thanks for adding tests

PR #683, opened 2022-09-03
[AL-3101] [AL-3102] Global Keys for Data Rows

jtsodapop:
Notes:
* Trying to get a global key that was deleted will not throw an error. added a TODO for consideration in regards to output to user
* Need to optimize performance on getting data row based on id, in `get_data_rows_for_global_keys()` by adding a bulk data row fetch function in the backend
* Bug around assigning a global_key whose data row has been deleted


labelbox/client.py
```diff
@@ -939,3 +941,133 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+                }
+                unmodifiedAssignments {
+                    dataRowId
+                    globalKey
+                }
+                accessDeniedAssignments {
+                    dataRowId
+                    globalKey
+                }}}
+        """
```
olegtrygub, line 997:
```suggestion
        get_failed_assignments_str = """query getDataRowsForGlobalKeysPyApi($jobId: ID!) {
            assignGlobalKeysToDataRowsResult(jobId: {id: $jobId}) {
                jobStatus
                data {
                    sanitizedAssignments {
                        dataRowId
                        globalKey
                    }
                    invalidGlobalKeyAssignments {
                        dataRowId
                        globalKey
                    }
                    unmodifiedAssignments {
                        dataRowId
                        globalKey
                    }
                    accessDeniedAssignments {
                        dataRowId
                        globalKey
                    }
                }}}
        """
```

labelbox/client.py
```diff
@@ -939,3 +941,136 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+        mutation_params = {
+            'globalKeyDataRowLinks': [
+                input.dict(by_alias=True)
+                for input in global_key_to_data_row_inputs
+            ]
+        }
+        assign_global_keys_to_data_rows_job = self.execute(
+            mutation_str, mutation_params)
+
+        get_failed_assignments_str = """query getDataRowsForGlobalKeysPyApi($jobId: ID!) {
```
kkim-labelbox, line 979: Looks like the query name should be updated for this function

kkim-labelbox, line 979: I would also name `get_failed_assignments_str` and `get_failed_assignments_param` something that reflects the output better, since it's not just returning failed assignments

labelbox/client.py
```diff
@@ -939,3 +941,136 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
             A ModelRun object.
         """
         return self._get_single(Entity.ModelRun, model_run_id)
+
+    def assign_global_keys_to_data_rows(
+            self,
+            global_key_to_data_row_inputs: List[AssignGlobalKeyToDataRowInput],
+            timeout=30) -> List[Dict[str, str]]:
```
kkim-labelbox, line 948: Let's name this `timeout_seconds` to be consistent with other async interfaces

labelbox/client.py
```diff
@@ -939,3 +941,136 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+        """
+        Assigns global keys to the related data rows.
+        
+        >>> global_key_data_row_inputs = [
```
kkim-labelbox, line 952: Great example! Makes it super clear how to use this function

labelbox/client.py
```diff
@@ -939,3 +941,136 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+                    }
+                }}}
+        """
+        get_failed_assignments_params = {
+            "jobId":
+                assign_global_keys_to_data_rows_job["assignGlobalKeysToDataRows"
+                                                   ]["jobId"]
+        }
+
+        while timeout >= 0:
```
kkim-labelbox, line 1007: Can we use the same approach as `project.export_labels()`, using `time.time()` to calculate the timeout?
```
sleep_time = 2
start_time = time.time()
while True:
  # logic here
  current_time = time.time()
  if current_time - start_time > timeout_seconds:
     raise Exception()
  time.sleep(sleep_time)
```

I remember this approach is less error prone, and we've seen timeouts in less than `timeout_seconds` with the current approach

kkim-labelbox (commented):
Leaving first set of comments!

labelbox/client.py
```diff
@@ -939,3 +941,136 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+                accessDeniedGlobalKeys
+                deletedDataRowGlobalKeys
+                } jobStatus}}
+            """
+        get_data_rows_params = {
+            "jobId":
+                data_rows_for_global_keys_job["dataRowsForGlobalKeys"]["jobId"]
+        }
+
+        while timeout >= 0:
```
kkim-labelbox, line 1067: Let's also convert this to use time.time(), as mentioned above

labelbox/client.py
```diff
@@ -939,3 +941,136 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+            res = self.execute(get_data_rows_str, get_data_rows_params)
+            if res["dataRowsForGlobalKeysResult"]['jobStatus'] == "COMPLETE":
```
kkim-labelbox, line 1069: Also need to handle jobStatus == FAILED

labelbox/schema/global_key.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class _CamelCaseMixin(BaseModel):
+
+    class Config:
+        allow_population_by_field_name = True
+        alias_generator = camel_case
+
+
+class AssignGlobalKeyToDataRowInput(_CamelCaseMixin):
```
kkim-labelbox, line 13: Discussed with @msokoloff1 offline on this. Let's use a dict as the parameter, for a few reasons:
- Most other places in code use primitives for parameters, so we want to follow the pattern for consistency
- Customers would need to have an extra line of import to use the assign function, which is moving away from our goal of improving usability
- Customers like dicts better as input params

We should remove this type, and `assign_global_keys_to_data_rows` should take in a dict, and validate the input.

labelbox/client.py
```diff
@@ -939,3 +940,306 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+        
+        Args:
+            A list of dicts containing data_row_id and global_key.
+        Returns:
+            Dictionary containing 'status', 'results' and 'errors'.
+
+            'Status' contains the outcome of this job. It can be one of
+            'Success', 'Partial Success', or 'Failure'.
+
+            'Results' contains the successful global_key assignments, including
```
jtsodapop, line 959: **nit** I think the word `Results` can be a bit misleading. A failure can also be a result. Is there any reason we do not choose to have it as `successes`?

labelbox/client.py
```diff
@@ -939,3 +940,306 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+                errors.extend(
+                    _format_failed_assignments(
+                        assignments=res['invalidGlobalKeyAssignments'],
+                        error_msg="Invalid global key"))
+                errors.extend(
+                    _format_failed_assignments(
+                        assignments=res['accessDeniedAssignments'],
+                        error_msg="Access denied to Data Row"))
+
+                if len(errors) == 0:
```
jtsodapop, line 1076: **suggestion** if errors is an empty list, we can rewrite this as


```suggestion
if not errors:
    status = "Success"
elif errors and results:
    status = "Partial Success"
else:
    status = "Failure"
                
                
```

jtsodapop: **nit** also, we have used enums to reflect possible paths instead of just text. might be better to have that? 

though i am unsure if it falls under the same category as the previous `input` where it was unnecessary, but users ideally would not need to do any additional importing/have to read it. and instead, it is possible to extend it if we wanted to continue on with messaging for success/partial/failure

```
Class SuccessType(enum):

SUCCESS = auto()
PARTIAL = auto()
FAILURE = auto()
```

labelbox/client.py
```diff
@@ -939,3 +940,306 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+            not necessarily contain all data rows requested.
+
+            'Errors' contains a list of global_keys that could not be fetched, along
+            with the failure reason
+        Examples:
+            >>> job_result = client.get_data_row_ids_for_global_keys(["key1","key2"])
+            >>> print(job_result['status'])
+            Partial Success
+            >>> print(job_result['results'])
+            ['cl7tv9wry00hlka6gai588ozv', 'cl7tv9wxg00hpka6gf8sh81bj']
```
jtsodapop, line 1129: **question** the more i think about this, the more i wonder if this should be a dict with global key and data row? i know Product said that this would be okay, but if i am user, is the expectation that i should know the direct mapping of datarow <> global key all the time after assigning it once?

labelbox/client.py
```diff
@@ -939,3 +940,306 @@ def get_model_run(self, model_run_id: str) -> ModelRun:
...
+                errors.extend(
+                    _format_failed_retrieval(data['notFoundGlobalKeys'],
+                                             "Data Row not found"))
+                errors.extend(
+                    _format_failed_retrieval(data['accessDeniedGlobalKeys'],
+                                             "Access denied to Data Row"))
+                errors.extend(
+                    _format_failed_retrieval(data['deletedDataRowGlobalKeys'],
+                                             "Data Row deleted"))
+                if len(errors) == 0:
```
jtsodapop, line 1178: **nit** same as my comments above, but again, a nit

labelbox/client.py
jtsodapop: spoke offline on this. keeping as `Results`

kkim-labelbox (approved):
LGTM!

PR #684, opened 2022-09-06
[AL-0] Updates to reduce the number of deprecation warnings

jtsodapop:


pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
-addopts = -s -vv -x --reruns 5 --reruns-delay 10 --durations=100
+addopts = -s -vv -x --reruns 5 --reruns-delay 10 --durations=20
```
jtsodapop, line 2: the number of returned tests i believe are a bit too many. maybe fetching only the top 20 is more useful.

labelbox/data/annotation_types/data/raster.py
```diff
@@ -187,7 +187,7 @@ def __repr__(self) -> str:
 
     class Config:
         # Required for sharing references
-        copy_on_model_validation = False
+        copy_on_model_validation = 'none'
```
jtsodapop, line 190: https://pydantic-docs.helpmanual.io/usage/model_config/

kkim-labelbox (approved):
lgtm!

PR #696, opened 2022-09-13
[AL-0] reduce the number of items docker will copy over

jtsodapop:
A result of a conversation about docker files providing bad marshal data led to the idea that docker is copying over unnecessarily all the files from the repo. we can reduce this to only the necessary files/folders


kkim-labelbox (approved):
Nice!

PR #701, opened 2022-09-14
[QQC-558] Remove test code updating queue_mode for a Project

richardsun0713:
Updating a project's queue mode will soon be disallowed through the API service, so this PR migrates the test code to instead set the queue mode upon project creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+# Version 0.0.0 (YYYY-MM-DD) - In Progress
+
+### Added
+### Changed
+* Updated docs for deprecated methods `_update_queue_mode` and `get_queue_mode` in `Project`
+    * Use the `queue_mode` attribute in `Project` to get and set the queue mode instead
+### Fixed
+
 # Version 3.27.1 (2022-09-16)
 ### Changed
 * Removed `client.get_data_rows_for_global_keys` until further notice
@@ -17,7 +25,7 @@
 ### Changed
 * Increase scalar metric value limit to 100m
 * Added deprecation warnings when updating project `queue_mode`
-## Fixed
+### Fixed
 * Fix bug in `feature_confusion_matrix` and `confusion_matrix` causing FPs and FNs to be capped at 1 when there were no matching annotations 
 
 # Version 3.26.2 (2022-09-06)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Changed
 * Updated docs for deprecated methods `_update_queue_mode` and `get_queue_mode` in `Project`
     * Use the `queue_mode` attribute in `Project` to get and set the queue mode instead
+    * For more information, visit https://docs.labelbox.com/reference/migrating-to-workflows#upcoming-changes
 ### Fixed
 
 # Version 3.27.1 (2022-09-16)
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -107,8 +107,6 @@ class Project(DbObject, Updateable, Deletable):
                 auto_audit_percentage <= 1.0
             Attempting to switch between benchmark and consensus modes is an invalid operation and will result
             in an error.
-
-        For more information, visit https://docs.labelbox.com/reference/migrating-to-workflows#upcoming-changes
         """
 
         media_type = kwargs.get("media_type")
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+# Version 0.0.0 (YYYY-MM-DD) - In Progress
+
+### Added
+### Changed
+* Removed the `_update_queue_mode` and `get_queue_mode` functions from `Project`
```
jtsodapop, line 7: We generally try avoiding taking away functionality right away because this can lead to user workflows breaking, which in turn makes them go to support and become unhappy.

From what I understand though this is not going to be allowed very soon from the API side. Could we reach out to Alex Cota on a proper deprecation plan for this and time it with the release of a new version to match removing this? 

* We need to notify MLSE/CSMs, to let users know that they will need to tell their customers that this will no longer be supported at date X
* Add deprecation warnings to the functions first, which will have the deprecation date X, and merge that as a PR first

jtsodapop (changes requested):

tests/integration/annotation_import/conftest.py
```diff
@@ -160,19 +160,19 @@ def configured_project_pdf(client, ontology, rand_gen, pdf_url):
 
 @pytest.fixture
 def configured_project_without_data_rows(client, configured_project, rand_gen):
-    project = client.create_project(name=rand_gen(str))
+    project = client.create_project(name=rand_gen(str),
```
jtsodapop, line 163: Could we also add this extra param in the example docstring here?

https://github.com/Labelbox/labelbox-python/blob/3b375d49b5e9ade23151e717876a4e4f3f312881/labelbox/client.py#L614

CHANGELOG.md
richardsun0713: The update queue mode functionality currently has undefined behavior when datarows have been attached to a project - this change is critical to preserve customer data integrity. Customers can still specify a queue mode for a project during creation time using the queue_mode attribute, so a migration path is readily available. Product has asked for the removal of this functionality immediately and without any feature flag rollout. That being said, I'll work with Pranoy and Alex to develop a communication plan to minimize surprise.

jtsodapop (approved):
lgtm :)

CHANGELOG.md
kkim-labelbox: I wonder if it's better to keep the `update queue_mode` functionality in code, but just throw an exception instead, so they are informed right when their script fails, instead of some vague "function missing error". What do you think @richardsun0713 ?

kkim-labelbox (commented):
Mostly looks good! I'd vote for adding an exception instead of removing the update queue_mode path, or even stripping down the `_update_queue_mode` method to just return an exception.

Also, if you haven't already, could you run the test suite against staging with the FF turned on?

richardsun0713:
Thanks for the reviews so far! There is a slight change of plan - we will not be introducing this breaking change to all customers immediately, but rather will hide the errors behind a FF that will then be rolled out in phases to various customer tiers. That means the `_update_queue_mode` and `get_queue_mode` functions will not be remove. Instead, this PR will update the docstrings + add a log warning to indicate the function will soon throw an error from the backend.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+# Version 0.0.0 (YYYY-MM-DD) - In Progress
+
+### Added
+### Changed
```
jtsodapop, line 6: **suggestion** would recommend changing this line:
https://github.com/Labelbox/labelbox-python/blob/3b375d49b5e9ade23151e717876a4e4f3f312881/labelbox/client.py#L614

to include queue_mode in the example:
>>> project = client.create_project(name="<project_name>", description="<project_description>", queue_mode=QueueMode.Batch)

labelbox/schema/project.py
```diff
@@ -105,10 +107,9 @@ def update(self, **kwargs):
                 auto_audit_percentage <= 1.0
             Attempting to switch between benchmark and consensus modes is an invalid operation and will result
             in an error.
+
+        For more information, visit https://docs.labelbox.com/reference/migrating-to-workflows#upcoming-changes
```
jtsodapop, line 111: like this info! wondering if instead of in project.update(), we ought to just include this in the changelog? 🤔

jtsodapop (approved):
lgtm!

kkim-labelbox (approved):

PR #702, opened 2022-09-14
update docstring and add warnings to mea to mal logic

msokoloff1:


mnoszczak (approved):
Logging errors, while still returning values is a nice DX improvement, thanks!

PR #703, opened 2022-09-15
[AL-0] Optimize get_data_rows_for_global_keys()

kkim-labelbox:


tests/integration/test_global_keys.py
```diff
@@ -2,14 +2,26 @@
 import pytest
 
 
+# Helper function to assert two lists of data rows are matchin since DataRows
+# constructed by get_data_rows_for_global_keys may not include certain fields
+def verify_data_rows_match(left_data_rows, right_data_rows):
```
jtsodapop, line 7: nice helper!

labelbox/client.py
```diff
@@ -1136,12 +1187,12 @@ def _format_failed_rows(rows: List[str],
...
             """
         params = {"globalKeys": global_keys}
-        data_rows_for_global_keys_job = self.execute(query_str, params)
+        data_rows_for_global_keys_job = self.execute(query_str, params, timeout=180)
 
         # Query string for retrieving job status and result, if job is done
         result_query_str = """query getDataRowsForGlobalKeysResultPyApi($jobId: ID!) {
             dataRowsForGlobalKeysResult(jobId: {id: $jobId}) { data { 
-                fetchedDataRows { id }
+                fetchedDataRows { id createdAt updatedAt externalId globalKey rowData }
```
jtsodapop, line 1195: based on the conversations yesterday, it sounds like this will revert back to just `id1

PR #704, opened 2022-09-16
increase mea upload timeout

msokoloff1:
Previously mea only supported 2k data rows so uploads were basically instant. Now that the limit has been raised substantially the client is timing out.

Ideally we will return an object representing the async task that allows users to check the status as needed. Currently the backend does not expose an endpoint compatible with out `Task` abstraction. Therefore, we have chosen not to introduce another async concept and this function will simply be blocking.


jtsodapop (approved):
lgtm

PR #714, opened 2022-09-27
[AL-3507] add message_id which is at the annotation level, and not specifically…

jtsodapop:
… data level for ner


jtsodapop:
Closing out in favor of other higher priority product features. Can be reopened in the future

PR #715, opened 2022-09-28
[AC-549] Introduce failed_data_rows to the Task class.

jakubborowski-lb:
- In order to support a backend change that is going to allow data row imports to succeed even when a small chunk of data rows might not get created, we're introducing the notion of `failed_data_rows`. These data row inputs will be returned in the results file on an import task.
- Remove unused imports.

Example output of a result file:
```jsonc
{
  "createdDataRows": [
    {
      "id": "cl8kui84n019hatrw5c17gdcl",
      "rowData": "https://storage.googleapis.com/image.jpeg",
      "datasetId": "cl8kui65h00bxatrwc2jifig0"
    },
    // additional created rows...
  ],
  "errors": [
    {
      "message": "Failed to create data rows after 5 retries. This chunk's data rows will not be imported.",
      "failedDataRows": [
        {
          "rowData": "https://storage.googleapis.com/image.jpeg"
        },
        // additional failed rows...
      ]
    },
    {
      "message": "Failed to create data rows after 5 retries. This chunk's data rows will not be imported.",
      "failedDataRows": [
        {
          "rowData": "https://storage.googleapis.com/image.jpeg"
        },
        // additional failed rows... 
      ]
    }
  ]
}
```

### DIFF
--- a/labelbox/schema/task.py
+++ b/labelbox/schema/task.py
@@ -1,8 +1,7 @@
 import logging
-import json
 import requests
 import time
-from typing import TYPE_CHECKING, TypeVar, Callable, Optional, Dict, Any, List
+from typing import TYPE_CHECKING, Callable, Optional, Dict, Any, List
 
 from labelbox.exceptions import ResourceNotFoundError
 from labelbox.orm.db_object import DbObject
@@ -64,7 +63,9 @@ class Task(DbObject):
         check_frequency = 2  # frequency of checking, in seconds
         while True:
             if self.status != "IN_PROGRESS":
-                if self.status == 'FAILED':
+                if self.status == "FAILED" or (self.status == "COMPLETE" and
+                                               self.failed_data_rows
+                                               is not None):
                     logger.warning(
                         "There are errors present. Please look at `task.errors` for more details"
                     )
@@ -80,16 +81,18 @@ class Task(DbObject):
 
     @property
     def errors(self) -> Optional[Dict[str, Any]]:
-        """ Downloads the result file from Task
+        """ Fetch the error associated with an import task.
         """
         if self.status == "FAILED":
             result = self._fetch_remote_json()
-            return result['error']
+            return result["error"]
+        elif self.status == "COMPLETE":
+            return self.failed_data_rows()
         return None
 
     @property
     def result(self) -> List[Dict[str, Any]]:
-        """ Fetch the result for a task
+        """ Fetch the result for an import task.
         """
         if self.status == "FAILED":
             raise ValueError(f"Job failed. Errors : {self.errors}")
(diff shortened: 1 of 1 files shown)


labelbox/schema/task.py
```diff
@@ -102,6 +101,19 @@ def result(self) -> List[Dict[str, Any]]:
                 'global_key': data_row.get('globalKey'),
             } for data_row in result['createdDataRows']]
 
+    @property
+    def failed_data_rows(self) -> List[Dict[str, Any]]:
```
kkim-labelbox, line 105: Glad we're introducing this partial failure concept!
One concern I have is, how will the customer know whether to call `failed_data_rows`. They wouldn't know that some data rows failed to be imported unless they explicitly called `Task.failed_data_rows`.
I think we should either:
1. Include failed data rows as part of `Task.result()`
2. Or include a warning message when they call `Task.result()`, if the result contains `errors`
(cc @olegtrygub )

kkim-labelbox: Also another question--if all data rows failed to be imported, will the Task status be "FAILED"?

jakubborowski-lb: Yeah I originally had wanted to include in `Task.result()` but realized if I did I would be breaking the interface of that property since it's just an array of data rows today. Ideally I'd return this in an object on the result that would look something like this.
```json
{
  "createdDataRows": [],
  "failedDataRows": []
}
```

jakubborowski-lb: @kkim-labelbox in response to your second question, currently the API does not mark that scenario as "FAILED" but I'll make a change to introduce that.

jakubborowski-lb: We'll only show the task as failed if all of the data rows have failed to be created, otherwise it'll be partial and the task will be marked as completed. In terms of showing the warning or the failed rows as part of the result, is it okay to introduce a breaking change here to the interface of the currently returned result?

kkim-labelbox: Ah yeah if the `Task.result()` is currently only returning the list of data rows, I don't see a good way to include failed data rows as part of it, without making a breaking change. 

At the same time, I feel that adding a new property may be confusing to the users. They may ask why there's `Task.errors` but also `Task.failed_data_rows`. They'd have to understand the nuances between a failed Task (all data rows failed) versus a completed Task with a partial failure. My proposal is to modify the `Task.errors` property to return the failed data rows when `status == "COMPLETE"` but result['error'] exists, and add an additional condition to check for `status == "COMPLETE" and result['errors'] is not None` to throw the existing warning message - https://github.com/Labelbox/labelbox-python/blob/20ee3a36149ede07ca0c24d23d6751ff6953a0fe/labelbox/schema/task.py#L68

What do you guys think?

jakubborowski-lb: I like that proposal, thanks @kkim-labelbox !!

kkim-labelbox (approved):
LGTM!

PR #721, opened 2022-10-12
Update video.ipynb

paultancre:
removing "media_type=MediaType.Video" from dataset since this cause an error : 
InvalidAttributeError: Field(s) ''media_type'' not valid on DB type 'Dataset'("Field(s) ''media_type'' not valid on DB type 'Dataset'", None)


paulnoirel (approved):
LGTM

PR #723, opened 2022-10-12
[X-0] Warn if queue_mode is not defined

mkozik-lb:
Since we'll be adjusting default createProject behavior to prefer batch projects, we want to give an early heads-up to our customers.


richardsun0713 (approved):

kkim-labelbox (approved):
LGTM!

PR #728, opened 2022-10-19
fix coco invalid polygons

msokoloff1:
There is a small chance that our algorithm for creating polygons from rasters creates small self intersections. Ultimately these polygons become invalid and errors are thrown when we attempt to use the invalid polygons.

### DIFF
--- a/labelbox/data/serialization/coco/panoptic_dataset.py
+++ b/labelbox/data/serialization/coco/panoptic_dataset.py
@@ -18,8 +18,6 @@ from .image import CocoImage, get_image, get_image_id, id_to_rgb
 from .annotation import PanopticAnnotation, SegmentInfo, get_annotation_lookup
 
 
-
-
 def vector_to_coco_segment_info(canvas: np.ndarray,
                                 annotation: ObjectAnnotation,
                                 annotation_idx: int, image: CocoImage,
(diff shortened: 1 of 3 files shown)


labelbox/data/serialization/coco/panoptic_dataset.py
```diff
@@ -70,20 +77,32 @@ def process_label(label: Label,
         for annotation_idx, annotation in enumerate(annotations[class_name]):
             categories[annotation.name] = hash_category_name(annotation.name)
             if isinstance(annotation.value, Mask):
-                segment, canvas = (mask_to_coco_segment_info(
+                coco_segment_info = mask_to_coco_segment_info(
```
kkim-labelbox, line 80: Questions around this.
1. How can coco_segment_info be set to `segment, canvas` when it's only returning `SegmentInfo` object?
2. When can coco_segment_info be None?

msokoloff1: 1. mask_to_coco_segment_info returns both of those things. It is hard to read because canvas looks like it is part of the SegmentInfo object but it isn't 
2. coco_segment_info can be None if the annotation is invalid. Like an empty seg mask is exported for whatever reason.

kkim-labelbox (approved):

PR #729, opened 2022-10-19
[AL-3274] Add Slice object and function to export data row IDs by slice

attila-papai:
- Added new `Slice` entity/dbobject
- Added `client.get_slice(id)` to fetch a Slice by ID
- Added `slice.get_data_row_ids()` to fetch data row ids of a Slice

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -1213,9 +1213,9 @@ class Client:
                 )
             time.sleep(sleep_time)
 
-    def get_slice(self, slice_id) -> CatalogSlice:
+    def get_catalog_slice(self, slice_id) -> CatalogSlice:
         """
-        Fetches a Slice by ID.
+        Fetches a Catalog Slice by ID.
 
         Args:
              slice_id (str): The ID of the Slice
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -33,7 +33,7 @@ from labelbox.schema.organization import Organization
 from labelbox.schema.user import User
 from labelbox.schema.project import Project
 from labelbox.schema.role import Role
-from labelbox.schema.slice import Slice
+from labelbox.schema.slice import CatalogSlice
 
 from labelbox.schema.media_type import MediaType
 
@@ -1213,7 +1213,7 @@ class Client:
                 )
             time.sleep(sleep_time)
 
-    def get_slice(self, slice_id) -> Slice:
+    def get_slice(self, slice_id) -> CatalogSlice:
         """
         Fetches a Slice by ID.
 
@@ -1235,4 +1235,4 @@ class Client:
             }
         """
         res = self.execute(query_str, {'id': slice_id})
-        return Entity.Slice(self, res['getSavedQuery'])
+        return Entity.CatalogSlice(self, res['getSavedQuery'])
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -33,6 +33,7 @@ from labelbox.schema.organization import Organization
 from labelbox.schema.user import User
 from labelbox.schema.project import Project
 from labelbox.schema.role import Role
+from labelbox.schema.slice import Slice
 
 from labelbox.schema.media_type import MediaType
 
@@ -963,7 +964,7 @@ class Client:
             timeout_seconds=60) -> Dict[str, Union[str, List[Any]]]:
         """
         Assigns global keys to data rows.
-        
+
         Args:
             A list of dicts containing data_row_id and global_key.
         Returns:
@@ -1211,3 +1212,27 @@ class Client:
                     "Timed out waiting for get_data_rows_for_global_keys job to complete."
                 )
             time.sleep(sleep_time)
+
+    def get_slice(self, slice_id) -> Slice:
+        """
+        Fetches a Slice by ID.
+
+        Args:
+             slice_id (str): The ID of the Slice
+        Returns:
+            Slice
+        """
+        query_str = """
+            query getSavedQueryPyApi($id: ID!) {
+                getSavedQuery(id: $id) {
+                    id
+                    name
+                    description
+                    filter
+                    createdAt
+                    updatedAt
+                }
+            }
+        """
+        res = self.execute(query_str, {'id': slice_id})
+        return Entity.Slice(self, res['getSavedQuery'])
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -1220,7 +1220,7 @@ class Client:
         Args:
              slice_id (str): The ID of the Slice
         Returns:
-            Slice
+            CatalogSlice
         """
         query_str = """
             query getSavedQueryPyApi($id: ID!) {
--- a/labelbox/schema/slice.py
+++ b/labelbox/schema/slice.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from labelbox.orm.db_object import DbObject
 from labelbox.orm.model import Field
 
@@ -9,7 +11,7 @@ class Slice(DbObject):
     updated_at = Field.DateTime("updated_at")
     filter = Field.Json("filter")
 
-    def get_data_row_ids(self) -> list[str]:
+    def get_data_row_ids(self) -> List[str]:
         """
         Fetches all data row ids that match this Slice
 
--- a/labelbox/schema/slice.py
+++ b/labelbox/schema/slice.py
@@ -5,7 +5,8 @@ from labelbox.pagination import PaginatedCollection
 
 class Slice(DbObject):
     """
-    A Slice is a saved set of filters (saved query)
+    A Slice is a saved set of filters (saved query).
+    This is an abstract class and should not be instantiated.
 
     Attributes:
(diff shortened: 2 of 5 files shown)


labelbox/schema/slice.py
```diff
@@ -0,0 +1,51 @@
+from typing import List
+
+from labelbox.orm.db_object import DbObject
+from labelbox.orm.model import Field
+
+
+class Slice(DbObject):
```
kkim-labelbox, line 7: I think we decided on CatalogSlice, since ModelSlice will be introduced in the future?

kkim-labelbox: Ah just read your message https://labelbox.slack.com/archives/C03FSV0K6MP/p1665642677853339?thread_ts=1665598292.363279&cid=C03FSV0K6MP.

If this `Slice` class is a parent class, I'm okay with it

labelbox/client.py
```diff
@@ -1211,3 +1212,27 @@ def _format_failed_rows(rows: List[str],
                     "Timed out waiting for get_data_rows_for_global_keys job to complete."
                 )
             time.sleep(sleep_time)
+
+    def get_slice(self, slice_id) -> Slice:
```
kkim-labelbox, line 1216: When we introduce `CatalogSlice` and `ModelSlice`, and have functions in those respective children classes, it won't be possible to do this:
```
catalog_slice = client.get_slice("catalog-slice-id")
catalog_slice.some_function_specific_to_catalog
```

labelbox/client.py
```diff
@@ -1211,3 +1212,27 @@ def _format_failed_rows(rows: List[str],
                     "Timed out waiting for get_data_rows_for_global_keys job to complete."
                 )
             time.sleep(sleep_time)
+
+    def get_slice(self, slice_id) -> CatalogSlice:
```
kkim-labelbox, line 1216: The return type should be `Slice` here, so that we prepare ourselves for ModelSlice later

labelbox/client.py
```diff
@@ -1211,3 +1212,27 @@ def _format_failed_rows(rows: List[str],
...
+                    name
+                    description
+                    filter
+                    createdAt
+                    updatedAt
+                }
+            }
+        """
+        res = self.execute(query_str, {'id': slice_id})
+        return Entity.CatalogSlice(self, res['getSavedQuery'])
```
kkim-labelbox, line 1238: We should think about when ModelSlice is introduced, how will this method distinguish between a CatalogSlice and ModelSlice. The method needs to be smart enough to figure that out, and create the corresponding Entity. Does SavedQuery have a "type" that we can return from graphql here?

labelbox/schema/slice.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    A Slice is a saved set of filters (saved query) used to find data rows matching a certain criteria
```
kkim-labelbox, line 9: Let's add a comment noting that this is an abstract class, and users should not instantiate this class

labelbox/schema/slice.py
```diff
@@ -0,0 +1,64 @@
...
+        data_row_ids: List[str] = []
+        total_count = 0
+        end_cursor = None
+        has_next_page = True
+        while has_next_page:
+            res = self.client.execute(query_str, {
+                'id': self.uid,
+                'after': end_cursor
+            })['getDataRowIdsBySavedQuery']
+            data_row_ids = data_row_ids + res['nodes']
```
kkim-labelbox, line 58: Are there limits in the backend as to how many data row ids can be returned? Can the python list handle that limit?

labelbox/schema/slice.py
```diff
@@ -0,0 +1,64 @@
...
+            total_count = res['totalCount']
+            has_next_page = res['pageInfo']['hasNextPage']
+            end_cursor = res['pageInfo']['endCursor']
+
+        assert total_count == len(data_row_ids)
```
kkim-labelbox, line 63: It may be better to throw a more transparent error message here, directing customers to support, as this implies there's something wrong with our code

kkim-labelbox (commented):
Left some comments! Could you also add some test cases?

labelbox/client.py
attila-papai: Yeah, there will be a `type` column in the corresponding `SavedQuery` table with possible values of `catalog` and `model`. Based on that, we could return an instance of the proper class.

labelbox/schema/slice.py
attila-papai: The backend returns 500 data rows at once. We merge the results here into a single list.
Hmm, should we rather use a generator? or `yield` the results?

labelbox/client.py
attila-papai: Hmm, in case of we also have `ModelSlice`, isn't the return type should be `Union[CatalogSlice, ModelSlice]` ?

msokoloff1: Will both kinds of slice objects always be exactly the same and support the same interface and the type is only used to indicate the source?

labelbox/schema/slice.py
```diff
@@ -0,0 +1,64 @@
...
+        description (datetime)
+        created_at (datetime)
+        updated_at (datetime)
+        filter (json)
+    """
+    name = Field.String("name")
+    description = Field.String("description")
+    created_at = Field.DateTime("created_at")
+    updated_at = Field.DateTime("updated_at")
+    filter = Field.Json("filter")
```
msokoloff1, line 22: Are all of these guaranteed to exist on the ModelSlice? I assume they will behave the exact same?

labelbox/schema/slice.py
msokoloff1: ```assert total_count == len(data_row_ids), "Unexpected result. Please contact support."```

labelbox/client.py
attila-papai: They are probably not 100% the same. I don't think it makes sense to have `get_data_row_ids` in a `ModelSlice`, right? The documents in the Model ES are annotation groups, so when you add filters on the Model UI you filter annotation groups. Maybe I'm wrong here, I don't know how Models work.

labelbox/schema/slice.py
attila-papai: I suggest to keep both catalog and model slices in the same Mysql table, so yes, these props would exists for both types of slices.

msokoloff1:
Also please add your new classes here:
https://github.com/Labelbox/labelbox-python/blob/226944a2e1d21eba8aed308128f0953933346d04/labelbox/orm/model.py#L359

attila-papai:
> Also please add your new classes here:
> 
> https://github.com/Labelbox/labelbox-python/blob/226944a2e1d21eba8aed308128f0953933346d04/labelbox/orm/model.py#L359

I saw this list but I wasn't sure. For example, `class FeatureSchema(DbObject):` is not listed here. Is that for a reason or a mistake?

labelbox/client.py
msokoloff1: Yeah there will potentially be differences. So this will prob return the union of the objects (in which case, how you have it now is fine). Although this might be confusing because you call get_slice and have to explicitly check what kind of object you have to know which functions you can use. Or we could create a dedicated function for each which is annoying but very explicit.

kkim-labelbox: Gotcha, yeah `Union[CatalogSlice, ModelSlice]` is better. We can leave this as is, and change the return type when we introduce ModelSlice

attila-papai: I like the idea better to have 2 functions `get_slice(id) -> CatalogSlice` and `get_model_slice(id) -> ModelSlice`

labelbox/schema/slice.py
msokoloff1: We should return a `PaginatedCollection` which is basically a generator.

msokoloff1:
> > Also please add your new classes here:
> > https://github.com/Labelbox/labelbox-python/blob/226944a2e1d21eba8aed308128f0953933346d04/labelbox/orm/model.py#L359
> 
> I saw this list but I wasn't sure. For example, `class FeatureSchema(DbObject):` is not listed here. Is that for a reason or a mistake?

It is a mistake.

labelbox/client.py
msokoloff1: I'm happy with either option. The only thing that I would ask is that if you choose to have two functions they should be ```get_catalog_slice(id) -> CatalogSlice and get_model_slice(id) -> ModelSlice```
(rename `get_slice` to `get_catalog_slice`)

attila-papai: Ok, I'll go with `get_catalog_slice(id) -> CatalogSlice`

labelbox/schema/slice.py
attila-papai: with the introduction of `PaginatedCollection` this is no longer necessary (I removed this code)

msokoloff1 (approved):
Great job!

msokoloff1 (approved):

PR #731, opened 2022-10-20
[X-0] Adjust integration tests to explicitly define queue mode

mkozik-lb:
Due to recent changes to default createProject behavior, we've started to default to `batches` if no queue mode is provided. Thus, we need to adjust integration tests.

Since the default behavior change is already on staging, I've run e2e tests against it. They all pass.

Tandem PR in API that adjusts the behavior: https://github.com/Labelbox/intelligence/pull/11294

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -160,7 +160,8 @@ def pdf_url(client):
 
 @pytest.fixture
 def project(client, rand_gen):
-    project = client.create_project(name=rand_gen(str))
+    project = client.create_project(name=rand_gen(str),
+                                    queue_mode=QueueMode.Dataset)
     yield project
     project.delete()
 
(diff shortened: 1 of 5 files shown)


msokoloff1:
Please format by either installing the pre-commit hooks or by just running:
```
yapf tests labelbox -i --verbose --recursive --parallel --style "google"
```

tests/integration/conftest.py
```diff
@@ -23,11 +23,11 @@
 
 
 class Environ(Enum):
-    LOCAL = 'local'
-    PROD = 'prod'
-    STAGING = 'staging'
-    ONPREM = 'onprem'
-    CUSTOM = 'custom'
+    LOCAL = "local"
```
kkim-labelbox, line 26: How come we're changing these double quotes? It's best practice to use single quotes these, iirc

msokoloff1 (approved):

PR #732, opened 2022-10-21
[AL-0] Polishing some sdk tests

kkim-labelbox:


tests/integration/test_delegated_access.py
```diff
@@ -28,7 +28,7 @@ def test_default_integration():
     ds.delete()
 
 
-@pytest.mark.skipif(os.environ.get("DA_GCP_LABELBOX_API_KEY") is None,
+@pytest.mark.skipif(os.environ.get("DA_GCP_LABELBOX_API_KEY") == "",
```
msokoloff1, line 31: This should be:
```@pytest.mark.skipif(os.environ.get("DA_GCP_LABELBOX_API_KEY"))```
That way this test doesn't run if either the env var is not set or if it is set to an empty string.

kkim-labelbox: `skipif` only accepts boolean as the condition, so I am doing `not os.environ.get("DA_GCP_LABELBOX_API_KEY")` instead, which should evaluate to false for empty string and None

kkim-labelbox: Good call nonetheless!

msokoloff1 (approved):

gszpak (approved):

PR #733, opened 2022-10-24
Include metadata in bulk create conversation in create_data_rows

ramy1951:
I have included the `metadata_fields` as an accepted key to the validation method of each message with creating multiple messages.

### DIFF
--- a/labelbox/schema/dataset.py
+++ b/labelbox/schema/dataset.py
@@ -295,8 +295,8 @@ class Dataset(DbObject, Updateable, Deletable):
             def check_message_keys(message):
                 accepted_message_keys = set([
                     "messageId", "timestampUsec", "content", "user", "align",
-                    "canLabel"
-                ])
+                    "canLabel", "metadata_fields"
+                ])   
                 for key in message.keys():
                     if not key in accepted_message_keys:
                         raise KeyError(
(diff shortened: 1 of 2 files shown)


karenkyang:
lgtm

labelbox/schema/dataset.py
```diff
@@ -295,7 +295,7 @@ def validate_conversational_data(conversational_data: list) -> None:
             def check_message_keys(message):
                 accepted_message_keys = set([
                     "messageId", "timestampUsec", "content", "user", "align",
-                    "canLabel"
+                    "canLabel", "metadata_fields"
```
msokoloff1, line 298: I don't think this is doing what you want. You are allowing users to provide metadata fields to individual messages here. Won't the below test case work even without your code change?

msokoloff1: What problem are we trying to solve? Data row metadata? Or you do want metadata per message somehow?

ramy1951: It should fail since all the keys in each message should be included in the set `accepted_message_keys`. I had to add this key to the list so that it can actually pass and not fail to raise a `KeyError`
Does that make sense or am I missing something?

The goal here was to have metadata fields added to each message. The validation would have failed if I did not add the key to the set `accepted_message_keys`.

kkim-labelbox: As Matt said, this is probably not doing what you want. This code snippet is only checking the `conversationalData` payload, whereas you're attempting to add `metadata_fields` outside of that

ramy1951: My understanding of the requirement that the metadata fields should be on per message basis not on payload basis

ramy1951: Since `metadata_fields` lives outside `conversationalData` list then no extra validation is needed and this PR is not needed.
Closing PR now.

PR #736, opened 2022-10-26
[QQC-755] Allow to pass batch consensus settings when creating a batch

mkozik-lb:
Tandem PR in API that needs to be merged before this: https://github.com/Labelbox/intelligence/pull/11383 (done)

This is part of project paradigm which is set to release on 2nd November. As part of the changes, we'll prefer customers to provide consensus settings per batch, rather than per project.

Updates to our SDK documentation are coming after this PR gets in.

### DIFF
--- a/labelbox/schema/consensus_settings.py
+++ b/labelbox/schema/consensus_settings.py
@@ -1,0 +1,18 @@
+from pydantic import BaseModel
+
+
+class ConsensusSettings(BaseModel):
+    """Container for holding consensus quality settings
+
+    >>> ConsensusSettings(
+    >>>    numberOfLabels = 2,
+    >>>    coveragePercentage = 0.2
+    >>>  )
+
+    Args:
+        numberOfLabels: Number of labels for consensus
+        coveragePercentage: Percentage of data rows to be labeled more than once
+    """
+
+    numberOfLabels: int
+    coveragePercentage: float
--- a/labelbox/schema/consensus_settings.py
+++ b/labelbox/schema/consensus_settings.py
@@ -1,18 +1,18 @@
-from pydantic import BaseModel
+from labelbox.utils import _CamelCaseMixin
 
 
-class ConsensusSettings(BaseModel):
+class ConsensusSettings(_CamelCaseMixin):
     """Container for holding consensus quality settings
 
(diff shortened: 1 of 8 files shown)


labelbox/schema/project.py
```diff
@@ -600,12 +605,15 @@ def create_batch(self, name: str, data_rows: List[str], priority: int = 5):
             }
         """ % (method, method, query.results_query_part(Entity.Batch))
 
+        consensus_settings_dict = consensus_settings.dict(
+        ) if consensus_settings else None
```
mkozik-lb, line 609: I don't want to fight the formatting here, but it's a pretty bad default ☝️

labelbox/schema/consensus_settings.py
```diff
@@ -0,0 +1,18 @@
+from pydantic import BaseModel
+
+
+class ConsensusSettings(BaseModel):
```
kkim-labelbox, line 4: A few things that I'd like to recommend here. @msokoloff1, would love your 2 cents here too.

- `numberOfLabels`and `coveragePercentage` should be snake_case to stay consistent
- In order to easily convert them to camelCase for graphql query, you can inherit `_CamelCaseMixin`, which currently exists in `schema.data_row_metadata`. This mixin also inherits from pydantic `BaseModel`. However, I would be keen to moving the mixin class to `labelbox.utils.py`, since we already have camelCase converters in that file.
- Once you inherit from _CamelCaseMixin, you can convert to camelCase when constructing the dictionary like so: `consensus_settings.dict(by_alias=True)`

labelbox/schema/project.py
```diff
@@ -561,7 +562,11 @@ def setup(self, labeling_frontend, labeling_frontend_options) -> None:
         timestamp = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
         self.update(setup_complete=timestamp)
 
-    def create_batch(self, name: str, data_rows: List[str], priority: int = 5):
+    def create_batch(self,
```
kkim-labelbox, line 565: Could you add test cases that use this new consensus_settings?

msokoloff1 (approved):

PR #738, opened 2022-10-27
[AL-4003] Adding DataRow and Dataset classes

kkim-labelbox:


labelbox_dev/data_row.py
```diff
@@ -0,0 +1,110 @@
+from typing import List, Optional, TypedDict, Union
+from labelbox_dev.entity import Entity
+from labelbox_dev.session import Session
+
+DATA_ROW_RESOURCE = "data-rows"
+
+
+class AttachmentsType(TypedDict):
```
mkozik-lb, line 8: Do we no longer prefer `BaseModel` from `pydantic` for data classes? AFAIK it's better in that it does runtime type check, so it may work better for SDK purposes.

kkim-labelbox: We will be performing pydantic validation internally, but that will be introduced later. For the interface itself, we don't want to use pydantic because 1. it requires users to import additional classes, and 2. it uses a lot of memory

PR #19, opened 2019-11-23
add code samples to dataset.py

alexandracota:
I added some samples to `dataset.create_data_rows` and `dataset.create_data_row` due to feedback from our support engineers that users are often confused by the documentation for these two. I also cleaned up the language a little bit.

PR #745, opened 2022-10-28
Update basics / basic project/datasets overview

ovalle15:


examples/basics/basics.ipynb
```diff
@@ -1,441 +1,468 @@
 {
```
msokoloff1, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    from labelbox.schema.media_type import MediaType
</pre><p>It would be nice to add these deps to the top level namespace.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/745/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

msokoloff1, line 1: <pre class="ql-syntax" spellcheck="false">Line #19.    print("Number of datasets in your org:", len(list(datasets)))
</pre><p>I don't think we want users pulling all projects and datasets. I would just do something like:</p><pre class="ql-syntax" spellcheck="false">dataset = next(client.get_dataset())
print("Dataset ID: ", dataset.uid)
print("Dataset Name:" , dataset.name)
</pre><p>Maybe just explain that both get_datasets() and get_projects() return a PaginatedCollection object which is just an iterable object. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/745/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

ovalle15: <p>I can do this moving forward , however, I am going to delete this two imports since I don't really use them on this notebook </p>

ovalle15: <p>Sounds good.</p>

ovalle15:
@msokoloff1 let me know if this notebook needs more attention. Thank you !

msokoloff1 (approved):

PR #749, opened 2022-10-28
Update video image import -  new paradigm

ovalle15:


examples/model_assisted_labeling/video_mal.ipynb
```diff
@@ -1,346 +1,557 @@
 {
```
msokoloff1, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    batch_datarows = [dr.uid for dr in list(dataset.export_data_rows())]
</pre><p>The list() is redundant since you are using list comprehension.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/749/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

msokoloff1:
I just had some nits around this. But one thought is that it would be good to include classifications in the video docs. I am asked about this every so often and there is no official example to point to.

examples/model_assisted_labeling/video_mal.ipynb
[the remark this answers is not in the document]
ovalle15: <p>Sounds good</p>

ovalle15: <p>Yes, Thank you !</p>

ovalle15:
@msokoloff1 I agree, our thinking was that for the new paradigm release, we should make the most important updates. I could go back after the initial release and either create more complicated examples (new colabs) or update this one. What do you think?

msokoloff1 (approved):
LGTM! Adding more complex examples at a later time is fine.

PR #755, opened 2022-11-04
[AL-4081] Wait for data rows to be processed when creating a batch

dubininsergey:
This PR adds a preliminary step to `create_batch` method. What it does is it periodically checks if specified data rows are finished processing. If waiting more than 1 hour - it throws an exception. 

Also, now SDK uses `createBatchV2` that also returns IDs of failed data rows. These data rows are available through `batch.failed_data_row_ids` property


tests/integration/test_batch.py
```diff
@@ -31,6 +31,23 @@ def small_dataset(dataset: Dataset):
...
+def dataset_with_invalid_data_rows(unique_dataset: Dataset):
+    upload_invalid_data_rows_for_dataset(unique_dataset)
+
+    yield unique_dataset
+
+
+def upload_invalid_data_rows_for_dataset(dataset: Dataset):
+    task = dataset.create_data_rows([
+        {
+            "row_data": 'https://jakub-da-test-primary.s3.us-east-2.amazonaws.com/dogecoin-whitepaper.pdf',
```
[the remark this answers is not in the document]
dubininsergey: > hm I'm not sure if such an URL should be part of the SDK test
> I saw we also use files from this domain https://storage.googleapis.com/labelbox-XXX
> could we reuse that somehow?

Good call, I found a bucket called `lb-test-private`, going to re-use the file from that bucket. It has no public access like the file that is already being used

labelbox/schema/batch.py
```diff
@@ -37,9 +37,10 @@ class Batch(DbObject):
     # Relationships
     created_by = Relationship.ToOne("User")
 
-    def __init__(self, client, project_id, *args, **kwargs):
+    def __init__(self, client, project_id, *args, failed_data_row_ids=None, **kwargs):
```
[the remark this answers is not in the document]
dubininsergey: This is a requirement only for Python 2.*, here is the PEP for 3.*: https://peps.python.org/pep-3102/

The intent with `failed_data_row_ids` is to make it **keyword only with default value**, if we place it before `args*` then it would fall into positional arguments section and raise `TypeError: __init__() got multiple values for argument 'failed_data_row_ids'`

tests/integration/test_batch.py
```diff
@@ -72,12 +89,69 @@ def test_batch_project(batch_project: Project, small_dataset: Dataset):
...
+
+    # upload data rows for this dataset and don't wait
+    upload_invalid_data_rows_for_dataset(unique_dataset)
+    unprocessed_data_rows = [dr.uid for dr in list(
+        unique_dataset.export_data_rows())]
+
+    data_row_ids = valid_data_rows + unprocessed_data_rows
+
+    with pytest.raises(ProcessingWaitTimeout):
+        batch_project._wait_processing_max_seconds = 0
```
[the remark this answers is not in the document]
dubininsergey: yeah, that makes sense totally, thanks!

labelbox/schema/batch.py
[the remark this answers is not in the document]
attila-papai: ah got it, thanks for the explanation!

attila-papai (approved):
LGTM!
@msokoloff1 mind taking a quick look as well?

labelbox/schema/batch.py
```diff
@@ -174,3 +180,7 @@ def delete_labels(self, set_labels_as_template=False) -> None:
                 },
             experimental=True)
         return res
+
+    @property
+    def failed_data_row_ids(self):
+        return self._failed_data_row_ids
```
msokoloff1, line 186: Is this a property so that in the future we can update the logic for how this is populated?

labelbox/schema/project.py
```diff
@@ -622,9 +631,12 @@ def create_batch(self,
...
                                   experimental=True)["project"][method]
-
-        res['size'] = len(dr_ids)
-        return Entity.Batch(self.client, self.uid, res)
+        batch = res['batch']
+        batch['size'] = len(dr_ids)
+        return Entity.Batch(self.client,
+                            self.uid,
+                            batch,
+                            failed_data_row_ids=res['failedDataRowIds'])
```
msokoloff1, line 639: What happens if this is 100k or more data row ids?

labelbox/schema/project.py
```diff
@@ -977,6 +989,42 @@ def _is_url_valid(url: Union[str, Path]) -> bool:
...
+        data_row_ids_param = "data_row_ids"
+
+        query_str = """query CheckAllDataRowsHaveBeenProcessedPyApi($%s: [ID!]!) {
+            queryAllDataRowsHaveBeenProcessed(dataRowIds:$%s) {
+                allDataRowsHaveBeenProcessed
+           }
+        }""" % (data_row_ids_param, data_row_ids_param)
+
+        params = {}
+        params[data_row_ids_param] = data_row_ids
```
msokoloff1, line 1023: you could just change this to:
```
params = {data_row_ids_param : data_row_ids}
```

tests/integration/conftest.py
```diff
@@ -191,6 +191,13 @@ def dataset(client, rand_gen):
     dataset.delete()
 
 
+@pytest.fixture(scope='function')
```
msokoloff1, line 194: This score is already the default and we usually don't set this.

labelbox/schema/project.py
dubininsergey: We don't allow more than 20K of data rows to be specified for a batch so here it can be 20K ids at most. That's about 500KB of data

labelbox/schema/batch.py
dubininsergey: That also works, but first of all to indicate that this is something that is read-only

msokoloff1 (approved):

PR #761, opened 2022-11-15
[AL-3578] Confidence support in SDK

dubininsergey:
This PR introduces "confidence" attribute into SDK area:

1) `AnnotationObject` and `ClassificationAnswer` classes were extended with a new attribute `confidence`
2) Added value validation: should be a valid number in [0, 1] range
3) NDJSON serializer/deserializer were adjusted to support this value's processing
4) Added pre-validation logic for MAL/Label imports which warns the user that confidence score is not supported and won't be imported


labelbox/data/annotation_types/annotation.py
```diff
@@ -1,6 +1,8 @@
 import abc
 from typing import Any, Dict, List, Optional, Union
 
+from labelbox.data.mixins import ConfidenceNotSupportedMixin, ConfidenceMixin
```
msokoloff1, line 4: `labelbox/data` contains both the models and utilities for working with those models. All models are defined under `labelbox/data/annotation_types`. The mixins should be moved to the models dir (annotation_types) since they will never be used anywhere else.

labelbox/data/serialization/ndjson/classification.py
```diff
@@ -241,6 +251,11 @@ def lookup_classification(
         }.get(type(annotation.value))
 
 
-NDSubclassificationType = Union[NDRadioSubclass, NDChecklistSubclass,
```
msokoloff1, line 244: How does this even run? This is use by the NDBaseObject.

labelbox/schema/confidence_presence_checker.py
```diff
@@ -0,0 +1,67 @@
...
+from labelbox.data.annotation_types.annotation import (
+    ClassificationAnnotation, ObjectAnnotation, VideoClassificationAnnotation,
+    VideoObjectAnnotation)
+from labelbox.data.annotation_types.classification.classification import (
+    Checklist, ClassificationAnswer, Dropdown, Radio, Text)
+from labelbox.data.annotation_types.label import Label
+from labelbox.data.annotation_types.metrics.confusion_matrix import \
+    ConfusionMatrixMetric
+from labelbox.data.annotation_types.metrics.scalar import ScalarMetric
+from labelbox.data.serialization.ndjson.converter import NDJsonConverter
```
msokoloff1, line 12: There are two main parts of the sdk. labelbox/data and labelbox/schema. data brings with it a whole bunch of dependencies that not all users might want (e.g. if they just want to make simple api calls without all of the utilities they would only install labelbox/schema). You can't require these modules in labelbox/schema or it breaks the pattern. 

Another consideration is that Pydantic is also not the fastest lib in the world so we don't want to keep converting between versions if it isn't necessary. 

Lastly, I think we want to minimize the number of client side checks. It allows for an inconstant experience (e.g. create_from_url wouldn't even benefit from this) and also it is easy to go stale / has to be maintained.

imo we shouldn't apply this check.

msokoloff1: If you really want a check, you could do something as simple as a regex on the json string for a confidence key anywhere and that would be much more performant and simple to maintain.

msokoloff1 (approved):

PR #763, opened 2022-11-16
[CCV-2223] Dont allow to call setup_editor multiple times

lgluszek:


msokoloff1:
Tests are failing because of other staging issues. You can merge this. I'll look into why these other tests are failing.

PR #770, opened 2022-11-17
[AL-4149] Validate metric names against reserved names

dubininsergey:
Validate metric names against the list of reserved ones


labelbox/data/metrics/iou/iou.py
```diff
@@ -62,7 +62,7 @@ def feature_miou_metric(ground_truths: List[Union[ObjectAnnotation,
         if value is None:
             continue
         metrics.append(
-            ScalarMetric(metric_name="iou", feature_name=key, value=value))
+            ScalarMetric(metric_name="custom_iou", feature_name=key, value=value))
```
dubininsergey, line 65: as we agreed in Slack discussion, we rename these to `custom_iou`

labelbox/data/annotation_types/metrics/scalar.py
```diff
@@ -28,6 +40,14 @@ class ScalarMetric(BaseMetric):
     value: Union[ScalarMetricValue, ScalarMetricConfidenceValue]
     aggregation: ScalarMetricAggregation = ScalarMetricAggregation.ARITHMETIC_MEAN
 
+    @validator('metric_name')
+    def validate_metric_name(cls, name: str):
+        clean_name = name.lower().strip()
+        if name.lower().strip() in RESERVED_METRIC_NAMES:
+            raise ValueError(f"`{clean_name}` is a reserved metric name. "
+                             "Please provide another value for `metric_name`.")
+        return clean_name
```
kozikkamil, line 49: I think we shouldn't force all lowercase, it's just for validation purpose

dubininsergey: I think we should be consistent, if we're validating against lower-cased value then we should store it that way as well

dubininsergey: but on the other hand it will break existing metrics, yeah 🤔

dubininsergey: I'd say validating against lower-cased metric name feels weird, we should just trim it

dubininsergey: @msokoloff1 @kozikkamil wdyt?

dubininsergey: alright, I think that's not so important

now we return the source name as it's currently in prod version 👌

labelbox/data/annotation_types/metrics/scalar.py
```diff
@@ -28,6 +33,14 @@ class ScalarMetric(BaseMetric):
     value: Union[ScalarMetricValue, ScalarMetricConfidenceValue]
     aggregation: ScalarMetricAggregation = ScalarMetricAggregation.ARITHMETIC_MEAN
 
+    @validator('metric_name')
+    def validate_metric_name(cls, name: str):
+        clean_name = name.lower().strip()
+        if name.lower().strip() in RESERVED_METRIC_NAMES:
```
kozikkamil, line 39: super nit, might want to change `name.lower().strip()` to `clean_name`

kozikkamil (approved):

msokoloff1 (approved):

PR #773, opened 2022-11-18
[AL-0] SDK support for clear global keys

olegtrygub:


labelbox/client.py
```diff
@@ -1229,6 +1229,110 @@ def _format_failed_rows(rows: List[str],
                 )
             time.sleep(sleep_time)
 
+    def clear_global_keys(
+            self,
+            global_keys: List[str],
+            timeout_seconds=60) -> Dict[str, Union[str, List[Any]]]:
+        """
+        Gets data row ids for a list of global keys.
```
kkim-labelbox, line 1237: Docstring should be updated

labelbox/client.py
```diff
@@ -1229,6 +1229,110 @@ def _format_failed_rows(rows: List[str],
...
+                notFoundGlobalKeys
+                accessDeniedGlobalKeys
+                } jobStatus}}
+            """
+        result_params = {
+            "jobId":
+                clear_global_keys_job["clearGlobalKeys"]["jobId"]
+        }
+
+        def _format_failed_rows(rows: List[str],
```
kkim-labelbox, line 1287: Nit: Could you define all helper functions right after the docstring?

labelbox/client.py
```diff
@@ -1229,6 +1229,110 @@ def _format_failed_rows(rows: List[str],
...
+                    _format_failed_rows(data['failedToClearGlobalKeys'],
+                                        "Clearing global key failed"))
+                errors.extend(
+                    _format_failed_rows(data['notFoundGlobalKeys'],
+                                        "Failed to find data row matching provided global key"))
+                errors.extend(
+                    _format_failed_rows(data['accessDeniedGlobalKeys'],
+                                        "Denied access to modify data row matching provided global key"))
+
+                # Invalid results may contain empty string, so we must filter
```
kkim-labelbox, line 1310: Does this still apply to clear_global_keys?

labelbox/client.py
olegtrygub: yeah, I just copypasted for the format. We should work on an informative docstring.

labelbox/client.py
olegtrygub: Sure!

labelbox/client.py
olegtrygub: It does! whatever you return is more of a discussion about backend

PR #785, opened 2022-12-02
add get_one and get_many methods to paginated collection

apollonin:
Adding more convenience to users when they want to get a predefined number of results from the paginated collection.

old way:
```
data_rows = dataset.data_rows()
data_row = next(data_rows)
```

new way:

```
data_row = dataset.data_rows().get_one()
```

old way:
```
data_rows = dataset.data_rows()
first_ten = []
for data_row in data_rows:
    first_ten.append(data_row)
    if len(first_ten) >= 10:
        break
```

new way: 
```
data_rows = dataset.data_rows().get_many(10)
```

important note. These methods are idempotent. That means: 
1. They can be called as many times as wanted and the result will be the same. E.x. get_one will **always** return first one result. It is not equivalent to calling next().
2. get_many will **not** call API in case it already has needed amount of data. This way to safe resources and time.

### DIFF
--- a/labelbox/pagination.py
+++ b/labelbox/pagination.py
@@ -45,7 +45,7 @@ class PaginatedCollection:
         """
         self._fetched_all = False
         self._data: List[Dict[str, Any]] = []
-        self._data_ind = 0
+        self._data_index = 0
 
         pagination_kwargs = {
             'client': client,
@@ -62,11 +62,11 @@ class PaginatedCollection:
                 **pagination_kwargs)
 
     def __iter__(self):
-        self._data_ind = 0
+        self._data_index = 0
         return self
 
     def __next__(self):
-        if len(self._data) <= self._data_ind:
+        if len(self._data) <= self._data_index:
             if self._fetched_all:
                 raise StopIteration()
 
@@ -75,9 +75,35 @@ class PaginatedCollection:
             if len(page_data) == 0:
                 raise StopIteration()
 
-        rval = self._data[self._data_ind]
-        self._data_ind += 1
-        return rval
+        next_value = self._data[self._data_index]
+        self._data_index += 1
+        return next_value
+
+    def get_one(self):
+      """Iterates over self and returns first value
+      This method is idempotent
+      """
+      for value in self:
+          return value
+
+    def get_many(self, n: int):
+      """Iterates over self and returns first n results
+      This method is idempotent
+
+      Args:
+          n (int): Number of elements to retrieve
+      """
+      results = []
+      i = 0
+
+      for value in self:
+          if i >= n:
+            break
+
+          results.append(value)
+          i += 1
+
+      return results
 
 
 class _Pagination(ABC):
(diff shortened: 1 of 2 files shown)


labelbox/pagination.py
```diff
@@ -45,7 +45,7 @@ def __init__(self,
         """
         self._fetched_all = False
         self._data: List[Dict[str, Any]] = []
-        self._data_ind = 0
+        self._data_index = 0
```
msokoloff1, line 48: I wouldn't rename anything even if it was poorly named. Technically someone could use this to determine the number of items in the collection after they iterate over it.

apollonin: naming is making me sad :( heh.
I know python does not have private variables, however, I would encourage our users not to reverse engineer and not use `_*` vars. 
But yea, I will rename it back

labelbox/pagination.py
```diff
@@ -79,6 +79,32 @@ def __next__(self):
         self._data_ind += 1
         return rval
 
+    def get_one(self):
+        """Iterates over self and returns first value
+        This method is idempotent
+        """
+        for value in self:
```
msokoloff1, line 86: It is kind of weird how each time you call this it fetches the next page but still returns the first result.

msokoloff1: Maybe this could return self._data[0]

apollonin: well, technically it will fetch next page only one time. Each consecutive call won't fetch the data page:


```
datasets = client.get_datasets()

datasets.get_one() # fetches next page
datasets.get_one() # no call to server

```

labelbox/pagination.py
```diff
@@ -79,6 +79,32 @@ def __next__(self):
...
+        """Iterates over self and returns first n results
+        This method is idempotent
+
+        Args:
+            n (int): Number of elements to retrieve
+        """
+        results = []
+        i = 0
+
+        for value in self:
```
msokoloff1, line 99: This has the same issue as above. Where you might want the first 200 but if you call the function twice it will fetch result 200-400. Even though this function only returns results 0 to 200

labelbox/pagination.py
msokoloff1: Each time you call get_one() it invokes the __next__ method which will fetch the next page and makes a call to the api.

labelbox/pagination.py
msokoloff1: Feel free to look closely at the work Kevin is doing for the next major SDK release. We want to be as meticulous as possible and extra 👀 are always helpful!

labelbox/pagination.py
apollonin: not exatly. I intentioally did loop here instead of next(), because it calls __iter__ firstly and sets _data_ind = 0
and the logic inside __next__ won't grab next page is previous is in memory

msokoloff1: oops sorry was reading one condition wrong (discussed in dms)

msokoloff1 (approved):

labelbox/pagination.py
cssetian: do we want next_page to continue iterating until it runs out of pages?

PR #789, opened 2022-12-06
pretty print objects values

apollonin:
Tested on nested dicts and dicts with nested objects.

Before: 
image

After:
image

Before: 
image

After: 
image


kozikkamil (approved):
Nice

PR #790, opened 2022-12-07
[QQC-1126] SDK support to create batch for 100k data rows

richardsun0713:
Allow the SDK to utilize the new async endpoints to create a batch if the batch size >= 10000 data rows.

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -18,10 +18,10 @@ from labelbox.orm.db_object import DbObject, Deletable, Updateable
 from labelbox.orm.model import Entity, Field, Relationship
 from labelbox.pagination import PaginatedCollection
 from labelbox.schema.consensus_settings import ConsensusSettings
+from labelbox.schema.data_row import DataRow
 from labelbox.schema.media_type import MediaType
 from labelbox.schema.queue_mode import QueueMode
 from labelbox.schema.resource_tag import ResourceTag
-from labelbox.schema.data_row import DataRow
 
 if TYPE_CHECKING:
     from labelbox import BulkImportRequest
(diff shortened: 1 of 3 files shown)


richardsun0713:
Tests added in this PR will require https://github.com/Labelbox/intelligence/pull/12285 to be deployed to staging before they will pass

labelbox/schema/project.py
```diff
@@ -608,22 +608,31 @@ def create_batch(self,
...
 
         if consensus_settings:
             consensus_settings = ConsensusSettings(**consensus_settings).dict(
                 by_alias=True)
+
+        if len(dr_ids) >= 10_000:
+            return self._create_batch_async(name, dr_ids, priority,
+                                            consensus_settings)
+        else:
+            return self._create_batch_sync(name, dr_ids, priority,
```
msokoloff1, line 620: Do we check that the sync create doesn't exceed 10k data rows on the backend? That might be a good safety check if we don't already. Especially since they can just call project._create_batch_sync() with more than 10k data rows.

richardsun0713: Yes, it is checked on the backend for the synchronous endpoints. There's a feature flag that controls the maximum

tests/integration/test_batch.py
```diff
@@ -55,6 +55,16 @@ def test_create_batch(batch_project: Project, big_dataset: Dataset):
...
+def test_create_batch_async(batch_project: Project, big_dataset: Dataset):
+    data_rows = [dr.uid for dr in list(big_dataset.export_data_rows())]
+    batch_project._wait_until_data_rows_are_processed(
+        data_rows, batch_project._wait_processing_max_seconds)
+    batch = batch_project._create_batch_async("big-batch", data_rows, 3)
+    assert batch.name == "big-batch"
+    assert batch.size == len(data_rows)
+
+
+@pytest.mark.skip()
```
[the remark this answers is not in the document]
richardsun0713: This was a mistake, reverting

msokoloff1 (approved):

PR #794, opened 2022-12-09
[AL-4044] Implement bulk data row import and async task interfaces for SDK4

kkim-labelbox:
https://labelbox.atlassian.net/browse/AL-4044 - DataRow bulk import
https://labelbox.atlassian.net/browse/AL-3890 - Async Task interface

### DIFF
--- a/labelbox_dev/data_row.py
+++ b/labelbox_dev/data_row.py
@@ -1,7 +1,8 @@
 from typing import Any, List, Optional, TypedDict, Union
 from labelbox_dev.entity import Entity
 from labelbox_dev.session import Session
-from labelbox_dev import utils
+from labelbox_dev.task import DataRowImportTask
+from labelbox_dev.exceptions import LabelboxError
 
 DATA_ROW_RESOURCE = "data-rows"
 
@@ -56,16 +57,25 @@ def get_by_global_keys(global_keys):
 
 
 def create(dataset_id, data_row: CreateDataRowType):
+    # TODO: Handle row_data as local file
     create_data_row_input = {'dataset_id': dataset_id, 'data_row': data_row}
-    # TODO: upload if row_data is local file
     data_row_json = Session.post_request(f"{DATA_ROW_RESOURCE}",
                                          json=create_data_row_input)
     return DataRow(data_row_json)
 
 
 def create_many(dataset_id, data_rows: List[CreateDataRowType]):
-    # TODO: Bulk creation and handling local files
-    pass
+    # TODO: Handle row_data as local file
+    create_data_rows_input = {'dataset_id': dataset_id, 'data_rows': data_rows}
+    task_json = Session.post_request(f"{DATA_ROW_RESOURCE}/bulkcreate",
+                                     json=create_data_rows_input)
+    task_id = task_json.get('task_id')
+
+    if task_id is None:
+        raise LabelboxError(
+            f"Failed to retrieve task information for `data_row.create_many()` operation"
+        )
+    return DataRowImportTask.get_by_id(task_id)
 
 
 class DataRow(Entity):
--- a/labelbox_dev/data_row.py
+++ b/labelbox_dev/data_row.py
@@ -1,8 +1,9 @@
 from typing import Any, List, Optional, TypedDict, Union
+
 from labelbox_dev.entity import Entity
+from labelbox_dev.exceptions import LabelboxError
 from labelbox_dev.session import Session
 from labelbox_dev.task import DataRowImportTask
-from labelbox_dev.exceptions import LabelboxError
 
 DATA_ROW_RESOURCE = "data-rows"
 
--- a/labelbox_dev/data_row.py
+++ b/labelbox_dev/data_row.py
@@ -56,7 +56,7 @@ def get_by_global_keys(global_keys):
     pass
 
 
-def create(dataset_id, data_row: CreateDataRowType):
+def create_one(dataset_id, data_row: CreateDataRowType):
     # TODO: Handle row_data as local file
     create_data_row_input = {'dataset_id': dataset_id, 'data_row': data_row}
     data_row_json = Session.post_request(f"{DATA_ROW_RESOURCE}",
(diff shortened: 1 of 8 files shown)


labelbox_dev/data_row.py
```diff
@@ -55,17 +56,39 @@ def get_by_global_keys(global_keys):
...
+        if task_id is None:
+            raise LabelboxError(
+                f"Failed to retrieve task information for `data_row.create()` async operation"
+            )
+        return DataRowImportTask.get_by_id(task_id)
+
+    # TODO: Need to paginate sync call. Currently, 100 data rows is the limit
+    data_rows = Session.post_request(f"{DATA_ROW_RESOURCE}/bulkcreate",
+                                     json=create_data_rows_input)
+    return [DataRow(data_row_json) for data_row_json in data_rows]
```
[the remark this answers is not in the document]
kkim-labelbox: Since we're only grabbing the first 100 data rows at the moment from the backend, there's no need to yield right now. I think we need to consider whether we want to introduce a paginated collection first, before implementing this return type as a generator

labelbox_dev/exceptions.py
```diff
@@ -0,0 +1,109 @@
+class LabelboxError(Exception):
```
msokoloff1, line 1: A lot of these exceptions are redundant and we should probably think through how best to handle them vs copying them over to the new code base. It makes it too easy to not fix if we re-use the same exceptions and I would argue that exception handling is pretty broken. 

Is there a plan for the exception hierarchy and types of exceptions that we will throw? This is a huge opportunity to fix this.

labelbox_dev/session.py
```diff
@@ -98,34 +94,35 @@ def _http_request(cls,
...
         except Exception as e:
-            raise labelbox.exceptions.LabelboxError(
+            raise LabelboxError(
                 "Unknown error during Client.query(): " + str(e), e)
 
         if response.status_code is requests.codes.no_content:
             return
 
         try:
+            print(response.content)
```
msokoloff1, line 109: ✂️

labelbox_dev/session.py
```diff
@@ -98,34 +94,35 @@ def _http_request(cls,
...
             r_json = response.json()
             print(
                 f"Response text: {response.text} Status code: {response.status_code}"
```
msokoloff1, line 112: i know this isn't from this pr, but we prob want a logger instead of print statements.

labelbox_dev/exceptions.py
kkim-labelbox: I think the exception handling being broken is due to the fact that our backend doesn't report exceptions nicely. This is handled better in our REST controllers, as we are practicing throwing appropriate HTTP exceptions. With that said, instead of copying over all exceptions from SDK 3, maybe it'll be good to start with only the exceptions we need, and build from there, since we do have an opportunity to clean up possible redundant exceptions

msokoloff1 (approved):

labelbox_dev/task.py
```diff
@@ -0,0 +1,186 @@
...
+from labelbox_dev.session import Session
+from labelbox_dev.utils import format_json_to_snake_case
+
+TASK_RESOURCE = "tasks"
+
+logger = logging.getLogger(__name__)
+
+
+class TaskStatus(Enum):
+    COMPLETE = "COMPLETE"
```
apollonin, line 22: Should it be `COMPLETED`?

labelbox_dev/task.py
```diff
@@ -0,0 +1,186 @@
...
+
+    task_type = TaskType.DATA_ROW_IMPORT
+
+    def __init__(self, json):
+        super().__init__(json)
+
+    @lru_cache()
+    def fetch_result(self) -> Dict[str, Any]:
+
+        def download_result():
```
apollonin, line 119: what is the reason for the method inside method?

labelbox_dev/task.py
```diff
@@ -0,0 +1,186 @@
...
+    def errors(self) -> Union[Dict[str, Any], None]:
+        task_result = self.fetch_result()
+        # TODO: Improve error message from backend
+        errors = {}
+        if 'errors' in task_result and len(task_result['errors']) != 0:
+            errors['errors'] = task_result['errors']
+        if 'failedDataRows' in task_result:
+            errors['failed_data_rows'] = task_result['failedDataRows']
+
+        if errors:
```
apollonin, line 158: `return errors if errors else None` - would be more "python-way" =)

labelbox_dev/task.py
```diff
@@ -0,0 +1,186 @@
...
+            return errors
+
+        return None
+
+    def wait_until_done(self, timeout_seconds=300) -> "TaskStatus":
```
apollonin, line 163: Why do we need to have timeout?

apollonin, line 163: yay! return value is a TaskStatus. I'm so happy =)

labelbox_dev/task.py
```diff
@@ -0,0 +1,186 @@
...
+        assert self.id is not None
+        check_frequency = 2  # frequency of checking, in seconds
```
apollonin, line 165: also current SDK allows you to set check_frequency. Don't we want to preserve it and have like 10 seconds by default?

labelbox_dev/task.py
kkim-labelbox: Task status in the backend is "COMPLETE"
```
export enum TaskStatus {
  Complete = 'COMPLETE',
  Failed = 'FAILED',
  InProgress = 'IN_PROGRESS',
}
```

kkim-labelbox: But we don't map the SDK task status to the backend anyway, so I can change it. `COMPLETED` is more consistent with `FAILED`

labelbox_dev/task.py
kkim-labelbox: No reason, just an artifact of copy pasta. Can remove

labelbox_dev/task.py
kkim-labelbox: Timeout ensures that there's nothing running indefinitely from the client side.

labelbox_dev/task.py
kkim-labelbox: I personally think customers don't need to worry about the check frequency.

apollonin: ah, true. it is only on label import. 
As soon as their requests don't add extra load on our servers - I'm fine with not having check_frequency configurable.

labelbox_dev/task.py
kkim-labelbox: Actually now that I'm implementing it, I think we should still keep it consistent with the backend TaskStatus for now. We are building this Enum in hopes that we can map them to the backend TaskStatus, and we're comparing SDK and backend TaskStatus at least for `IN_PROGRESS`. There is then an expectation that all TaskStatuses map correctly. It is not trivial to update the backend TaskStatus either, because SDK 3 customers may depend on it

PR #803, opened 2022-12-15
support local file as a source of data rows

apollonin:
https://labelbox.atlassian.net/browse/AL-4494.- followup ticket

### DIFF
--- a/labelbox_dev/data_row.py
+++ b/labelbox_dev/data_row.py
@@ -35,6 +35,26 @@ class UpdateDataRowType(TypedDict):
     external_id: Optional[str]
     row_data: Optional[str]
 
+def validate_keys(item):
+    if 'row_data' not in item:
+        raise LabelboxError("`row_data` missing when creating DataRow.")
+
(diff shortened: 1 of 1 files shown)


labelbox_dev/data_row.py
```diff
@@ -66,14 +67,34 @@ def create_one(dataset_id, data_row: CreateDataRowType):
 
 
 def create(dataset_id,
-           data_rows: List[CreateDataRowType],
+           data_rows: List[CreateDataRowType] = [],
+           data_files: List[str] = [],
```
kkim-labelbox, line 71: Instead of adding a `data_files` as a new parameter (we should avoid clogging the interface with additional parameters. Every new parameter will need to be understood by users), we should just handle the case when `data_row.row_data` is a local file

```
for dr in data_rows:
  if os.path.exists(dr.row_data);
     # Set row_data as the contents of the file
```

kkim-labelbox (approved):
LGTM!

PR #804, opened 2022-12-16
add get_by_name method to metadata ontology

apollonin:


labelbox/schema/data_row_metadata.py
```diff
@@ -147,6 +147,31 @@ def _build_ontology(self):
...
+    ) -> Union[DataRowMetadataSchema, Dict[str, DataRowMetadataSchema]]:
+        """ Get metadata by name
+
+        >>> mdo.get_by_name(name)
+
+        Args:
+            name (str): Name of metadata schema
+
+        Returns:
+            Metadata schema as `DataRowMetadataSchema` or dict, in case of nested metadata
```
kkim-labelbox, line 161: - `in case of Enum metadata` is probably more accurate
- It'd be nice to provide an example output of calling get_by_name

kkim-labelbox (approved):
One nit comment, but otherwise looks great!

labelbox/schema/data_row_metadata.py
apollonin: Thanks! I have changed the text as you suggesting. 
Separately, where do you want me to provide an example? I can not find any instance of Example in the repo. I will update the SDK doc of course after we release this version.

PR #807, opened 2022-12-20
[AL-4042] DataRow get_by_ids() and get_by_global_keys()

kkim-labelbox:


labelbox_dev/data_row.py
```diff
@@ -146,10 +59,122 @@ def from_json(self, json) -> "DataRow":
...
+
+    @staticmethod
+    def get_by_global_key(global_key: str):
+        params = {'is_global_key': True}
+        data_row_json = Session.get_request(f"{DATA_ROW_RESOURCE}/{global_key}",
+                                            params=params)
+        return DataRow(data_row_json)
+
+    @staticmethod
+    def get_by_ids(data_row_ids, run_async: bool = False):
```
apollonin, line 140: `get_by_ids` and `get_by_global_keys` are almost identical expect of body json param. Should we move the common parts to separate class for instance `GetBulkDataRow` and have static method there `get_many` or similar 

then these new methods will become: 


```
@staticmethod
 def get_by_ids(data_row_ids, run_async: bool = False):
   body = {'keys': data_row_ids, 'run_async': run_async}
   return DataRowshelper.getBulk(body)

 @staticmethod
 def get_by_global_keys(global_keys, run_async: bool = False):
   body = {'keys': global_keys, 'is_global_key': True, 'run_async': run_async}
   return DataRowshelper.getBulk(body)
```

kkim-labelbox: Definitely want to change this, but I want to give it some more thought. I'm going to create a follow-up ticket and address it as a bigger item to optimize

apollonin (approved):
this looks good to me

PR #808, opened 2022-12-20
[AL-4256] Create datarow metadata by name

whistler:
When adding metadata to a datarow, allow specifying the metadata name instead of the schema_id.
- For upsert operations, if schema_id is not present, load schema_id from name before making an API call
- Update example notebook to use names instead of schema_id

In addition to the unit tests, this has been tested using the updated example notebook.


kkim-labelbox (approved):
LGTM! Nice work, and great job making edits to the notebook!
Looks like the build is failing due to `yapf` linter. I'd suggest setting up a pre-commit hook for it - https://github.com/Labelbox/labelbox-python/blob/develop/CONTRIB.md#commits
```
pip install pre-commit
pre-commit install
```

labelbox/schema/data_row_metadata.py
```diff
@@ -172,6 +175,21 @@ def get_by_name(
         else:
             raise KeyError(f"There is no metadata with name {name}")
 
+    def _get_schema_by_name(
```
apollonin, line 178: Curious, why not use `self.get_by_name(name)` here? Looks like a bit of duplication and I wonder if we need to consider using this method here or creating a new with better naming with explicitly says "get_by_normalized_name" or similar

labelbox/schema/data_row_metadata.py
```diff
@@ -624,13 +644,15 @@ def _convert_metadata_field(metadata_field):
             if isinstance(metadata_field, DataRowMetadataField):
                 return metadata_field
             elif isinstance(metadata_field, dict):
-                if not all(key in metadata_field
-                           for key in ("schema_id", "value")):
+                if not "value" in metadata_field:
                     raise ValueError(
-                        f"Custom metadata field '{metadata_field}' must have 'schema_id' and 'value' keys"
+                        f"Custom metadata field '{metadata_field}' must a 'value' key"
```
apollonin, line 649: must "have" a value key?

labelbox/schema/data_row_metadata.py
```diff
@@ -664,11 +686,33 @@ def _upsert_schema(
         self.refresh_ontology()
         return _parse_metadata_schema(res)
 
+    def _find_option_id_by_name(self, schema_name: str, option_name: str) -> DataRowMetadataSchema:
+        # Precondition: already validated that schema_name is a valid Schema name
+        try:
+            return self._fields_by_name[schema_name][option_name].uid
```
apollonin, line 692: this one definitely should use `self.get_by_name(name)[option].uid`

apollonin:
I would definitely consider better naming or even reusage of existing method. 
Otherwise we end up with `get_by_name` and `_get_schema_by_name` methods that sounds identical, but do slightly different things.  The future maintenance would be cumbersome

labelbox/schema/data_row_metadata.py
whistler: For enums reseved_by_name and custom_by_name are dicts, they don't provide the id of the field, so I can't use `get_by_name`. I agree the duplication here is not ideal, I am open to ideas. When reading through the code, I personally found the term `normalized` a little confusing and wanted to avoid using that in the function.

whistler: I have extracted a common lookup function used by both functions and renamed the new function to `_ get_by_name_normalized`. `nested` and `flat` index might be better terminology, but don't want to change too much in this PR.

labelbox/schema/data_row_metadata.py
whistler: Good catch, fixed!

labelbox/schema/data_row_metadata.py
whistler: Good catch! I've updated it.

apollonin (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # Version 3.34.0 (...)
 ### Added
 * Added `get_by_name()` method to MetadataOntology object to access both custom and reserved metadata by name.
+* Added support for adding metadata by name when creating datarows using `dataset.create_data_rows()`.
```
kkim-labelbox, line 6:
```suggestion
* Added support for adding metadata by name when creating datarows using `DataRowMetadataOntology.bulk_upsert()`.
```

kkim-labelbox (changes requested):

kkim-labelbox (approved):

PR #809, opened 2022-12-20
[AL-4416] Increase DataRow limits to 150000

kkim-labelbox:
Load tested with and without metadata. As a result of recent changes to DataRow import processing, it seems that DataRow import is much faster.

150,000
w/ metadata: 1st run: 105 sec, 2nd run: 78 sec
w/o metadata: 1st run: 76 sec, 2nd run: 50 sec


apollonin (approved):
x5 to the moon!

PR #811, opened 2022-12-22
Al 4502 improved error handling for single annotation uploads

apollonin:


labelbox/schema/annotation_import.py
```diff
@@ -141,6 +141,18 @@ def _create_from_bytes(cls, client, variables, query_str, file_name,
         files = {file_name: file_data}
         return client.execute(data=data, files=files)
 
+    @classmethod
+    def _get_ndjson_from_objects(cls, data, slug):
```
kkim-labelbox, line 145: nit: 
Instead of `data` and `slug`, can we rename to `objects` and `object_name`, which stays more consistent with the function name?

kkim-labelbox, line 145: Also, can we define the return type here?

tests/unit/test_unit_label_import.py
```diff
@@ -33,3 +34,28 @@ def test_should_warn_user_about_unsupported_confidence():
             warning_mock.assert_called_once()
             "Confidence scores are not supported in Label Import" in warning_mock.call_args_list[
                 0].args[0]
+
+
+def test_invalid_labels_format():
+    """this test should confirm that labels are required to be in a form of dict"""
```
kkim-labelbox, line 40: `required to be a list`?

tests/unit/test_mal_import.py
```diff
@@ -39,3 +40,36 @@ def test_should_warn_user_about_unsupported_confidence():
             warning_mock.assert_called_once()
             "Confidence scores are not supported in MAL Prediction Import" in warning_mock.call_args_list[
                 0].args[0]
+
+
+def test_invalid_labels_format():
+    """this test should confirm that annotations are required to be in a form of dict"""
```
kkim-labelbox, line 46: `required to be a list`?

apollonin:
@kkim-labelbox thank you, I have addressed all comments

msokoloff1 (approved):

PR #812, opened 2022-12-22
[AL-0] Better error handling for label creation during test

kkim-labelbox:
Fixed two places where we were using a Generator to verify existence of the label, which was resulting in an error message that was hard to decipher

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -337,19 +337,13 @@ def configured_project_with_label(client, rand_gen, image_url, project, dataset,
         }
     }]
 
-    def create_label():
-        """ Ad-hoc function to create a LabelImport
+    upload_task = LabelImport.create_from_objects(
+        client, project.uid, f'label-import-{uuid.uuid4()}', predictions)
+    upload_task.wait_until_done(sleep_time_seconds=5)
+    assert upload_task.state == AnnotationImportState.FINISHED, "Label Import failed"
 
-        Creates a LabelImport task which will create a label
-        """
-        upload_task = LabelImport.create_from_objects(
-            client, project.uid, f'label-import-{uuid.uuid4()}', predictions)
-        upload_task.wait_until_done(sleep_time_seconds=5)
-        assert upload_task.state == AnnotationImportState.FINISHED
-
-    project.create_label = create_label
-    project.create_label()
-    label = next(project.labels())
+    label = project.labels().get_one()
+    assert label is not None, "Cannot fetch created label"
     yield [project, dataset, datarow, label]
 
     for label in project.labels():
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -339,17 +339,17 @@ def configured_project_with_label(client, rand_gen, image_url, project, dataset,
...
     def create_label():
         """ Ad-hoc function to create a LabelImport
-
         Creates a LabelImport task which will create a label
         """
         upload_task = LabelImport.create_from_objects(
             client, project.uid, f'label-import-{uuid.uuid4()}', predictions)
         upload_task.wait_until_done(sleep_time_seconds=5)
-        assert upload_task.state == AnnotationImportState.FINISHED
+        assert upload_task.state == AnnotationImportState.FINISHED, "Label Import failed"
```
msokoloff1, line 347: why would upload_task.state be equal to
```AnnotationImportState.FINISHED, "Label Import failed"```

File: tests/integration/test_data_row_media_attributes.py [hunk 2]
```diff
@@ -5,6 +5,8 @@ def test_export_empty_media_attributes(configured_project_with_label):
     project, _, _, _ = configured_project_with_label
     # Wait for exporter to retrieve latest labels
     sleep(10)
-    labels = project.label_generator()
-    label = next(labels)
-    assert label.data.media_attributes == {}
+    labels = project.label_generator().as_list()
```
msokoloff1, line 8: you can also just call list(project.label_generator()). It makes it more readable.

tests/integration/test_data_row_media_attributes.py
```diff
@@ -5,6 +5,8 @@ def test_export_empty_media_attributes(configured_project_with_label):
...
+    assert len(
+        labels
+    ) == 1, "Labels could not be fetched via `project.label_generator()`"
```
msokoloff1, line 11: I am already imagining this error message being posted all over slack. This makes it seem like the problem is the sdk. I would say that label export job unexpectedly returned an empty result set.

msokoloff1 (approved):
Just left some nits. Feel free to ignore.

tests/integration/test_data_row_media_attributes.py
apollonin: or even 
```
labels = [project.label_generator().get_one()]
```

to be consistent with the other test

apollonin: ah, len([None]) == 1, forget my comment )

tests/integration/test_data_row_media_attributes.py
kkim-labelbox: Thanks for the suggestions! Made the changes

PR #816, opened 2022-12-22
New NDJSON format - tile imagery

ovalle15:


msokoloff1:
This seems like a duplicate of the notebook here https://github.com/Labelbox/labelbox-python/pull/817. I would say let's consolidate this. MAL is a type of annotation import so idk why it needs to be treated differently for each type. We should highlight the fact that all of the schemas are the same whether you do mal / label import / mea import.

PR #818, opened 2022-12-22
increase max string len to 1024

msokoloff1:


kkim-labelbox (approved):

msokoloff1:
Keeping this branch around because some current tests rely on this. Please delete at any point in 2023

PR #819, opened 2022-12-22
[AL-0] Fix tests that were flaky due to label creation

kkim-labelbox:
- Added a step to wait for label to finish processing in `configured_project_with_label`
- Added a step to wait for data row to finish processing for tests that rely on media_attributes to be populated, specifically tests that use label exports (labels on data rows that have not been processed are not exported).

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -119,6 +119,49 @@ def get_invites(client):
     return invites
 
 
+def wait_for_data_row_processing(client, data_row):
+    """
+    Do not use. Only for testing.
+
+    Returns DataRow after waiting for it to finish processing media_attributes.
+    Some tests, specifically ones that rely on label export, rely on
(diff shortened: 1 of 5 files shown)


msokoloff1 (approved):

tests/integration/conftest.py
```diff
@@ -407,3 +407,56 @@ def configured_project_with_complex_ontology(client, rand_gen, image_url):
...
+
+    Returns DataRow after waiting for it to finish processing media_attributes.
+    Some tests, specifically ones that rely on label export, rely on
+    DataRow be fully processed with media_attributes
+    """
+
+    def func(client, data_row):
+        data_row_id = data_row.uid
+        timeout_seconds = 60
+        while True:
```
whistler, line 425: We could also use the [retry](https://pypi.org/project/retry/) package for this logic.

PR #823, opened 2022-12-23
add local guardian when running sdk tests

apollonin:
Help avoid issue when one of the required service is not running locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,18 +4,24 @@ build:
 
 
 test-local: build
+
+	@# if PATH_TO_TEST we asuume you know what you are doing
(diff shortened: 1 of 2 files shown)


kkim-labelbox (commented):
Awesome work! Could you add a check for rabbitmq as well?

apollonin:
> 

do you mean RabbitMq which is part of intelligence setuip? I assume if API works (by curl :8080), then rabbitmq as well as elastic search are alive as well

cssetian (approved):
Approving since Kevin's OOO and the risk of these changes is low (ie local test running)

PR #826, opened 2023-01-04
[AL-0] Add assertion for label import failure

kkim-labelbox:


cssetian (approved):

attila-papai (approved):
thanks @kkim-labelbox  !

PR #830, opened 2023-01-06
Add VSCode dev container

whistler:
Add configuration for creating a reproducible VSCode development environment inside a docker container.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,12 @@
+# Requirements for development environment. These are not needed for CI.
+
+# Requirements for development
+ipdb==0.13.11
+rich==12.6.0
+ipykernel==6.16.2
+ipywidgets==8.0.3
+pre-commit==2.21.0
+tox==4.0.16
+# Requirements for example notebooks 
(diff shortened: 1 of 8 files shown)


requirements-dev.txt
```diff
@@ -0,0 +1,12 @@
+# Requirements for development environment. These are not needed for CI.
```
apollonin, line 1: I know we use `poetry` all around the services to manage python dependencies. I think we can use it here as well. 
I wonder if we also can use poetry's virtual env instead of devcontainer to manage this environment? I don't have experience with VSCode devcontainer, though

whistler: True this can be done in a poetry virtualenv as well, but I feel this is more reproducible and more similar to the CI environment. Longer story: I was getting a bunch of segfaults in python libs when running the SDK on OSX, and had to reinstall python on two different occasions, so decided to develop in a docker container instead.

whistler:
Closing this PR.  Since SDK isnt meant to be run in a docker container, its better to use poetry instead. However dev containers can be useful for developing services.

PR #831, opened 2023-01-06
[AL-4418] [AL-3976] Make dataset tests parallelizable

whistler:
Remove the assumption that only the current test is modifying data stores for dataset integration tests. Tested on staging. 
Screenshot 2023-01-18 at 9 56 11 PM


apollonin (commented):
this is a great start! Do you want to merge these changes right away?

apollonin (approved):

tests/integration/test_dataset.py
```diff
@@ -48,9 +40,6 @@ def test_dataset(client, rand_gen):
     assert dataset.description == description
 
     dataset.delete()
```
kkim-labelbox, line 42: It would be good to add a check for the deleted dataset here by querying for the dataset
```
with pytest.raises(SomeException):
   client.get_dataset(dataset.uid)
```

tests/integration/test_pagination.py
```diff
@@ -0,0 +1,13 @@
+from copy import copy
+
+
+def test_get_one_and_many_dataset_order(client):
```
kkim-labelbox, line 4: Nice work adding coverage for pagination!

tests/integration/test_dataset.py
whistler: This already exists on line 44.

tests/integration/test_pagination.py
whistler: I felt that part of the test_dataset was really testing pagination so I moved that here.

tests/integration/test_client_errors.py
```diff
@@ -45,7 +45,7 @@ def test_semantic_error(client):
 def test_timeout_error(client, project):
     with pytest.raises(labelbox.exceptions.TimeoutError) as excinfo:
         query_str = """query getOntology { 
-        project (where: {id: $%s}) {
```
apollonin, line 48: how did it even work ?

tests/integration/test_data_row_metadata.py
```diff
@@ -17,7 +18,7 @@
 TEXT_SCHEMA_ID = "cko8s9r5v0001h2dk9elqdidh"
 CAPTURE_DT_SCHEMA_ID = "cko8sdzv70006h2dk8jg64zvb"
 PRE_COMPUTED_EMBEDDINGS_ID = 'ckrzang79000008l6hb5s6za1'
-CUSTOM_TEXT_SCHEMA_NAME = 'custom_text'
+CUSTOM_TEXT_SCHEMA_NAME = 'datarow_metadata_custom_text'
```
apollonin, line 21: I wonder if we want to use random schema name here for each test?

tests/integration/test_user_and_org.py
```diff
@@ -19,5 +19,5 @@ def test_user_and_org_projects(project):
 
     assert project.created_by() == user
     assert project.organization() == org
-    assert set(user.projects()) == user_projects.union({project})
-    assert set(org.projects()) == org_projects.union({project})
+    assert project in user_projects
```
apollonin, line 22: great catch! It has been intermittently failing recently on staging and with your change now should be fine!

apollonin (approved):

tests/integration/test_client_errors.py
whistler: maybe it timed out before this got evaluated.

tests/integration/test_data_row_metadata.py
whistler: I think this change is not needed anymore so I'll revert it, before we were creating and deleting this for each test, but that doesn't work with parallel tests.

PR #836, opened 2023-01-18
[AL-000] Fixed missing confidence attribute in NDline and NDTextEntity

dubininsergey:
In this PR we fix missing confidence attribute for Polyline and TextEntity objects, also adding missing tests for these cases

### DIFF
--- a/labelbox/data/serialization/ndjson/objects.py
+++ b/labelbox/data/serialization/ndjson/objects.py
@@ -96,7 +96,8 @@ class NDLine(NDBaseObject, ConfidenceMixin):
     def from_common(cls, line: Line,
                     classifications: List[ClassificationAnnotation], name: str,
                     feature_schema_id: Cuid, extra: Dict[str, Any],
-                    data: Union[ImageData, TextData]) -> "NDLine":
+                    data: Union[ImageData, TextData], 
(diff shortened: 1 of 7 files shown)


labelbox/data/serialization/ndjson/objects.py
```diff
@@ -96,7 +96,8 @@ def to_common(self) -> Line:
     def from_common(cls, line: Line,
                     classifications: List[ClassificationAnnotation], name: str,
                     feature_schema_id: Cuid, extra: Dict[str, Any],
-                    data: Union[ImageData, TextData]) -> "NDLine":
+                    data: Union[ImageData,
+                                TextData], confidence: float) -> "NDLine":
```
msokoloff1, line 100: Won't this break if confidence isn't set?

```
        optional_kwargs = {}
        if (annotation.confidence):
            optional_kwargs['confidence'] = annotation.confidence
...
       
        return obj.from_common(annotation.value, subclasses, annotation.name,
                               annotation.feature_schema_id, annotation.extra,
                               data, **optional_kwargs)
```
There is no positional argument for confidence or default

dubininsergey: correct, fixed

msokoloff1 (approved):

PR #840, opened 2023-01-23
[AL-4398] Add Model Run exports

mnoszczak:
Adds model run exports method method


labelbox/schema/model_run.py
```diff
@@ -446,6 +449,66 @@ def export_labels(
...
+    def export_labels_v2(self, task_name: str,
+                         filter: Optional[ModelRunExportFilter]) -> Task:
+        mutation_name = "exportDataRows"
+        create_task_query_str = """mutation exportDataRowsPyApi($input: ExportDataRowsInput!){
+          %s(input: $input) {taskId} }
+          """ % (mutation_name)
+        params = {
+            "input": {
+                "taskName": task_name,
+                "filters": {
```
msokoloff1, line 468: It is confusing that the backend calls the query a "filter" and the filters params. In the SDK we are calling this a filter. Ideally we could follow up and make the naming convention consistent.

labelbox/schema/model_run.py
```diff
@@ -446,6 +449,66 @@ def export_labels(
...
+                    "modelRunIds": [self.uid],
+                    "projectIds": []
+                },
+                "params": {
+                    "includeAttachments":
+                        filter["attachments"]
```
msokoloff1, line 474: It is also weird that the gql fields are different than the name we expose in the SDK. I can see a future where this causes confusion :). 

We should make these the same.

tests/integration/annotation_import/test_model_run.py
```diff
@@ -114,6 +114,15 @@ def test_model_run_export_labels(model_run_with_model_run_data_rows):
     assert len(labels) == 3
 
 
+def test_model_run_export_labels_v2(model_run_with_model_run_data_rows):
+    task_name = "test_task"
+    task = model_run_with_model_run_data_rows.export_labels_v2(
+        task_name, filter={"media_attributes: true"})
+    assert task.name == task_name
+    task.wait_till_done()
+    assert task.status == "COMPLETE"
```
msokoloff1, line 123: It would be good to check at least that the high level keys exist in the payload and the number of items returned was expected.

labelbox/schema/model_run.py
```diff
@@ -446,6 +449,66 @@ def export_labels(
...
+                        if filter and "attachments" in filter else False,
+                    "includeMediaAttributes":
+                        filter['media_attributes']
+                        if filter and 'media_attributes' in filter else False,
+                    "includeMetadata":
+                        filter['metadata_fields']
+                        if filter and 'metadata_fields' in filter else False,
```
msokoloff1, line 481: Maybe this would read better if at the top you did something like:
```
_filter = filter or {}
```
Then instead of all of these conditions you can just do:
```
"includeAttachments" : _filter.get('attachments', False)
```

mnoszczak: I like that idea

labelbox/schema/model_run.py
```diff
@@ -446,6 +449,66 @@ def export_labels(
...
 
+    """
+    Creates a model run export task with the given filter and returns the task.
+    
+    >>>    export_task = export_labels_v2("my_export_task", filter={"media_attributes": True})
+    
+    """
+
+    def export_labels_v2(self, task_name: str,
+                         filter: Optional[ModelRunExportFilter]) -> Task:
```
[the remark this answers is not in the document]
attila-papai:
```
filter={"media_attributes": True}
```
hm, this does not look like a filter to me but rather a parameter of the export

labelbox/schema/model_run.py
```diff
@@ -446,6 +449,66 @@ def export_labels(
...
+        mutation_name = "exportDataRows"
```
attila-papai, line 461: I still think this general mutation should be removed and two specialized mutations be added instead:
- exportDataRowsInProject
- exportDataRowsInModelRun

Right now, the user can call this mutation without any filters and that would export all the data rows in his organization, and if he has 100M data rows, it's going to add a huge load to our system.

labelbox/schema/model_run.py
mnoszczak: Would you update SDK or the API? Also, should SDK use camelCase or snake_case?

tests/integration/annotation_import/test_model_run.py
mnoszczak: Shouldn't that be in the scope of integration tests of API, not the SDK?

labelbox/schema/model_run.py
attila-papai: so when you do
```
model_run = client.get_model_run("model_run_id")
model_run.export_labels_v2(...)
```
isn't that constrained to a single model run?

mnoszczak: It is constrained to a single model run from within that SDK call:
```
                    "modelRunIds": [self.uid],
                    "projectIds": []
                },
```

labelbox/schema/model_run.py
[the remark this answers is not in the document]
mnoszczak: I have renamed from `ModelRunExportFilter` to `ModelRunExportParams`

labelbox/schema/model_run.py
attila-papai: so it would make sense if `mutation exportDataRowsInModelRun` accepts only a single model run id, because users can't export multiple model runs in a single call, right?
I can add this new mutation, shouldn't be more than half an hour.

mnoszczak: @attila-papai https://github.com/Labelbox/intelligence/pull/13104 I've added it, but it accepts multiple modelRunIds, should we:
a) make it general purpose and rename to `exportDataRowsInModelRuns`
b) limit it to just one `modelRunId` and keep the `exportDataRowsInModelRun` name

labelbox/schema/model_run.py
msokoloff1: We should update the api. I think it is fine if the api casing is camelcase and sdk is snake case just that the names should be the same.

tests/integration/annotation_import/test_model_run.py
msokoloff1: The SDK is the most useful end to end integration test. So it is nice to have it here to guarantee that things are working (also it blocks deployments). But technically it is weird that it serves that purpose.

labelbox/schema/model_run.py
```diff
@@ -446,6 +449,66 @@ def export_labels(
...
             time.sleep(sleep_time)
 
+    """
+    Creates a model run export task with the given filter and returns the task.
+    
+    >>>    export_task = export_labels_v2("my_export_task", filter={"media_attributes": True})
+    
+    """
+
+    def export_labels_v2(self, task_name: str,
```
msokoloff1, line 459: from @karenkyang:
We should call these functions `.export_v2(...)`

labelbox/schema/model_run.py
```diff
@@ -446,6 +449,66 @@ def export_labels(
                          self.uid)
             time.sleep(sleep_time)
 
+    """
+    Creates a model run export task with the given filter and returns the task.
+    
+    >>>    export_task = export_labels_v2("my_export_task", filter={"media_attributes": True})
```
msokoloff1, line 455: filter={"media_attributes": True})
^^ params

labelbox/schema/export_params.py
```diff
@@ -0,0 +1,25 @@
...
+
+class ProjectExportParams(DataRowParams):
+    include_project_details: Optional[bool]
+    include_label_details: Optional[bool]
+    include_performance_details: Optional[bool]
+
+
+class ModelRunExportParams(DataRowParams):
+    # TODO: Add model run fields
+    pass
```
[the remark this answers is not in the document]
mnoszczak: I think that we've discussed this point with Karen and ended up on picking TypedDict, because it requires less ceremony from the end user

msokoloff1 (approved):

PR #841, opened 2023-01-23
[AL-4397] Add project & data row import v2

mnoszczak:


msokoloff1 (approved):

labelbox/schema/project.py
```diff
@@ -371,6 +375,71 @@ def _validate_datetime(string_date: str) -> bool:
...
+    
+    >>>    export_task = export_v2("my_export_task", filter={"media_attributes": True})
+    
+    """
+
+    def export_v2(self, task_name: str,
+                  params: Optional[ProjectExportParams]) -> Task:
+        defaultParams: ProjectExportParams = {
+            "attachments": False,
+            "media_attributes": False,
```
msokoloff1, line 389: This always is requested. So let's remove it from the params.

mnoszczak: I'll fix in a follow up PR

PR #856, opened 2023-01-25
[AL-3765][AL-4666] Pagination class for v4

whistler:
Adds Identifier and Cursor-based pagination as described in the TDD: https://docs.google.com/document/d/1CssQvP9yZEG4MIJXLqI995dDZSdBMXIYZ8GvLChARGk/edit#

- DataRow `get_by_ids()` and `get_by_global_keys()` have been updated to use the new class  
- Refactored loading `api_key` and `api_host` in test suite to support v4 integration tests

Depends on updates in API in this PR: https://github.com/Labelbox/intelligence/pull/13005


labelbox_dev/pagination.py
```diff
@@ -0,0 +1,135 @@
...
+from labelbox_dev.session import Session
+from labelbox_dev.exceptions import ServiceError
+from labelbox_dev.entity import Entity
+
+DEFAULT_LIMIT = 100
+DATA_KEY = 'data'
+CURSOR_KEY = 'cursor'
+NEXT_CURSOR_KEY = 'next'
+LIMIT_KEY = 'limit'
+IDS_KEY = 'ids'
```
kkim-labelbox, line 14: `DEFAULT_IDENTIFIER_KEY` sounds more consistent than `IDS_KEY`

labelbox_dev/pagination.py
```diff
@@ -0,0 +1,135 @@
...
+        self.entity_class: Type[Entity] = entity_class
+        self.identifiers: List[str] = identifiers
+        self.identifiers_key: str = identifiers_key
+        self.limit: int = limit
+        self._current_idx = 0
+        super().__init__(resource, entity_class, params, limit)
+
+    def get_next_page(self):
+        end_idx = self._current_idx + self.limit
+        ids = self.identifiers[self._current_idx:end_idx]
```
kkim-labelbox, line 87: `identifiers_for_page` instead of `ids`?

labelbox_dev/pagination.py
```diff
@@ -0,0 +1,135 @@
...
+        response = self._session.get_request(self.resource, params)
+        self._cursor = response.get(NEXT_CURSOR_KEY)
+        if self._cursor is None:
+            self._is_last_page = True
+
+        try:
+            self._cursor = response['next']
+            return response['data']
+        except KeyError:
+            raise ServiceError(
```
kkim-labelbox, line 134: This error sounds pretty vague, so I don't think it's worth creating a new error type for this. We should try to handle most of the errors accordingly within the `http_request` method in Session. If for some reason, `data` does not exist in the response, we can just throw a generic `LabelboxError`, or even `KeyError` with a more detailed error message.

Also, looks like we're already checking for `next` within the response in line 126, so we don't need line 131.

kkim-labelbox (commented):
Leaving some comments. Thanks for adding test coverage!

tests/integration/labelbox_dev/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@pytest.fixture(scope="session")
+def datarows(dataset):
+    BASE_PATH = 'https://storage.googleapis.com/labelbox-datasets/image_sample_data/image-sample-'
+    data_rows = [{
+        "row_data": f"{BASE_PATH}{i}.jpg",
+        "global_key": f"{BASE_PATH}{i}.jpg" + str(uuid.uuid4()),
+    } for i in range(5)]
+    datarows = DataRow.create(dataset.id, data_rows)
```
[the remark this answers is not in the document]
whistler: I've renamed to `data_rows` for consistency. This is a fixture so should it have assertions for data row creation? Wouldn't `DataRow.create` throw an exception if it was not able to create these data rows? I think we should add a separate test for data row creation.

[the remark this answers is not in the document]
kkim-labelbox: I think it's still good to verify that the test data is created successfully, so we can decouple test setup failures from actual test failures, but maybe it's not necessary in this case, since `DataRow.create` will throw an error anyway. As a general rule though, in cases where test data can be created incorrectly, I think it's good to have assertions.

kkim-labelbox (approved):
Approved. Nice work! I would run a couple integration tests locally to make sure your changes around the fixtures do not break existing SDK tests, before you merge!

PR #857, opened 2023-01-25
Annotation Import Text - updates to ontology, tools, misc

ovalle15:
1. Replace ImageData with TextData 
2. Remove additional import (ImageData)
3. Removed URL signing function  All has been tested !


ovalle15:
@ezekielemerson I deleted the nested python annotation tool

ezekielemerson (approved):

PR #869, opened 2023-02-02
[AL-4764] Add Classification `name` and make `instructions` optional

whistler:
Depends on: https://github.com/Labelbox/intelligence/pull/13181/files


whistler:
Making instructions optional on the backend also need updates to the label editor, I have created a PR with a narrower scope: https://github.com/Labelbox/labelbox-python/pull/870

PR #876, opened 2023-02-06
[AL-4919] Make params optional, add task result to export task type

mnoszczak:


labelbox/schema/model_run.py
```diff
@@ -456,13 +456,16 @@ def export_labels(
     
     """
 
-    def export_v2(self, task_name: str,
-                  params: Optional[ModelRunExportParams]) -> Task:
-        _params = params or {}
+    def export_v2(self,
+                  task_name: Optional[str] = None,
+                  params: Optional[ModelRunExportParams] = {}) -> Task:
```
msokoloff1, line 461: Assigning a default argument of a dict is bad practice. It isn't re-defined for each function invokation, only when the function is defined.

https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python 

Technically you aren't modifying this so it should be fine. Also, optional means default value of None - not that it has a default. 

I would revert back to how you have it. But just set the value to None by default.

tests/integration/annotation_import/test_model_run.py
```diff
@@ -117,7 +117,7 @@ def test_model_run_export_labels(model_run_with_model_run_data_rows):
     assert len(labels) == 3
 
 
-@pytest.mark.skip(reason="feature under development")
+# @pytest.mark.skip(reason="feature under development")
```
msokoloff1, line 120: You can just delete this line instead of commenting it out

msokoloff1: Although I think exports v2 is broken in staging right now.. Maybe leave this until it is more stable.

msokoloff1 (approved):

PR #883, opened 2023-02-07
[AL-4858][AL-4860] Reduce imports and hide NDJsonConverter

whistler:
SDK objects can now be imported with:
```
import labelbox as lb
import labelbox.types as lb_types
```

* Update `__init__.py` file include imports used by notebooks but not in the root namespace.
* Create a types module that imports annotation types
* Update AnnotationImport and BulkImportRequest to serialize Python annotations.

Tested the dependency on labelbox[data] by uninstalling shapely and running a notebook that uses NDJson-based annotation import.

Notebooks will be updated in a subsequent PR, there have been a number of updates in `develop` since this PR was opened and a large number of conflicts. I'll redo the renaming instead of resolving conflicts.

### DIFF
--- a/labelbox/__init__.py
+++ b/labelbox/__init__.py
@@ -20,7 +20,7 @@ from labelbox.schema.webhook import Webhook
 from labelbox.schema.ontology import Ontology, OntologyBuilder, Classification, Option, Tool, FeatureSchema
 from labelbox.schema.role import Role, ProjectRole
 from labelbox.schema.invite import Invite, InviteLimit
-from labelbox.schema.data_row_metadata import DataRowMetadataOntology
+from labelbox.schema.data_row_metadata import DataRowMetadataOntology, DataRowMetadataField, DataRowMetadata, DeleteDataRowMetadata
 from labelbox.schema.model_run import ModelRun, DataSplit
 from labelbox.schema.benchmark import Benchmark
 from labelbox.schema.iam_integration import IAMIntegration
@@ -28,3 +28,4 @@ from labelbox.schema.resource_tag import ResourceTag
 from labelbox.schema.project_resource_tag import ProjectResourceTag
 from labelbox.schema.media_type import MediaType
 from labelbox.schema.slice import Slice, CatalogSlice
+from labelbox.schema.queue_mode import QueueMode
--- a/labelbox/data/annotation_types/geometry/line.py
+++ b/labelbox/data/annotation_types/geometry/line.py
@@ -4,6 +4,7 @@ import geojson
 import numpy as np
 import cv2
 from pydantic import validator
+
 from shapely.geometry import LineString as SLineString
 
 from .point import Point
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
 # Changelog
+# Version 3.38.0 (2023-xx-xx)
+
+### Added
+* All imports are available via `import labelbox as lb` and `import labelbox.types as lb_types`.
+
+## Changed
+* `LabelImport.create_from_objects()`, `MALPredictionImport.create_from_objects()`, `MEAPredictionImport.create_from_objects()`, `Project.upload_annotations()`, `ModelRun.add_predictions()` now support Python Types for annotations.
 
 # Version 3.37.0 (2023-02-08)
 ## Added
--- a/labelbox/data/annotation_types/__init__.py
+++ b/labelbox/data/annotation_types/__init__.py
@@ -34,3 +34,9 @@ from .metrics import ConfusionMatrixMetric
 from .metrics import ConfusionMatrixAggregation
 from .metrics import ScalarMetricValue
 from .metrics import ConfusionMatrixMetricValue
+
+from .data.tiled_image import EPSG
+from .data.tiled_image import EPSGTransformer
+from .data.tiled_image import TiledBounds
+from .data.tiled_image import TiledImageData
+from .data.tiled_image import TileLayer
--- a/labelbox/data/annotation_types/collection.py
+++ b/labelbox/data/annotation_types/collection.py
@@ -298,4 +298,4 @@ class LabelGenerator(PrefetchGenerator):
         return self._process(value)
 
 
-LabelCollection = Union[LabelList, LabelGenerator]
+LabelCollection = Union[LabelList, LabelGenerator, Iterable[Label]]
--- a/labelbox/data/annotation_types/geometry/geometry.py
+++ b/labelbox/data/annotation_types/geometry/geometry.py
@@ -3,8 +3,9 @@ from abc import ABC, abstractmethod
 
 import geojson
 import numpy as np
-from shapely import geometry as geom
 from pydantic import BaseModel
+
+from shapely import geometry as geom
 
 
 class Geometry(BaseModel, ABC):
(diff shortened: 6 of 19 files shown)


examples/annotation_import/basics.ipynb
```diff
@@ -78,12 +78,22 @@
     },
```
kkim-labelbox, line 78: <p>Should we delete this warning?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/883/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

examples/annotation_import/text.ipynb
```diff
@@ -105,16 +105,9 @@
       },
```
kkim-labelbox, line 105: <pre class="ql-syntax" spellcheck="false">Line #11.    print("   ")
</pre><p>Let's get rid of this print statement</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/883/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='32'/>

examples/annotation_import/tiled.ipynb
```diff
@@ -77,13 +77,13 @@
     },
```
kkim-labelbox, line 77: <pre class="ql-syntax" spellcheck="false">Line #11.    print("   ")
</pre><p>Let's get rid of this</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/883/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='37'/>

examples/test_notebooks.py
```diff
@@ -0,0 +1,72 @@
+import pathlib
```
kkim-labelbox, line 1: Can we put a docstring at the top of this file explaining what the purpose of this is?

kkim-labelbox (approved):

labelbox/__init__.py
```diff
@@ -2,6 +2,10 @@
 __version__ = "3.36.1"
 
 from labelbox.client import Client
+from labelbox.data import annotation_types as types
```
msokoloff1, line 5: This is going to break potentially. There are additional dependencies required for this module.

msokoloff1 (changes requested):
This will break. See comments here:
https://github.com/Labelbox/labelbox-python/pull/883/files#r1099422756

labelbox/__init__.py
msokoloff1: Since the models themselves and related utilities are tightly coupled, this only works if you import labelbox[data] to include the extra deps. If users only install labelbox, this import will fail.

labelbox/__init__.py
```diff
@@ -2,6 +2,10 @@
...
+from labelbox.data import metrics
+from labelbox.data.serialization import NDJsonConverter
+from labelbox.data.serialization import COCOConverter
```
msokoloff1, line 8: Same with these modules. But also no one should ever have to use these directly. Instead I recommend updating `create_from_objects()` to accept annotation types and we handle the conversion ourselves.

I don't see anything wrong with having obscure functions not in the top level lib.

labelbox/__init__.py
```diff
@@ -2,6 +2,10 @@
 __version__ = "3.36.1"
 
 from labelbox.client import Client
+from labelbox.data import annotation_types as types
+from labelbox.data import metrics
```
msokoloff1, line 6: We don't really want customers using this now that we have auto metrics. I would just remove this (it also has the same dep issue)

labelbox/__init__.py
msokoloff1: you could get around this with some sort of lazy loading of this module. Or you could have users do something like:
from labelbox.data import annotation_types as lb_types.

labelbox/__init__.py
msokoloff1: Basically follow this pattern that we use for projects
import: https://github.com/Labelbox/labelbox-python/blob/a356b0c993c76678b6b02dadc5289334f74792d3/labelbox/schema/project.py#L40-L43
catch at run-time:
https://github.com/Labelbox/labelbox-python/blob/a356b0c993c76678b6b02dadc5289334f74792d3/labelbox/schema/project.py#L1243-L1248

labelbox/__init__.py
whistler: There are a couple of notebooks that use metrics:
* examples/model_diagnostics
* examples/integrations/detectron2
should we remove metrics calculations from these notebooks?

labelbox/__init__.py
whistler: I can remove `COCOConverter` since only a few notebooks use that. @kkim-labelbox is there a ticket to hide NDJsonConverter?

labelbox/__init__.py
karenkyang: ideally we standarize to lb_*, 
from labelbox import annotation_types as lb_types
from labelbox import metrics as lb_metrics

labelbox/__init__.py
karenkyang: We should create a ticker for " we handle the conversion (between AnnotatinoTypes -> ndjson) ourselves." if we haven't/ This is part of annotation v0 scope

Ideally, no need for users to use NDJsonConverter at all

msokoloff1: It is a very small task. I would say it is part of this work basically.

labelbox/__init__.py
msokoloff1: You could lazy import or just in those notebooks (which are basically deprecated) keep the old import path.

karenkyang: @maximeLabelbox I don't think we need any metrics libraries anymore because of auto-metrics. Can you confirm? 
Only place that will need metrics will be custom metrics, and we should probably create a separate notebook for that

whistler: I have removed metrics from the root namespace, but the above notebooks still use metrics.

labelbox/__init__.py
whistler: I think a second import in the notebooks makes the dependency `labelbox[data]` more transparent. It would be nice to show this dependency in the import statement e.g.  `from labelbox.data import types as lb_types`. I'll update this.

labelbox/data/annotation_types/geometry/line.py
```diff
@@ -1,10 +1,12 @@
-from typing import List, Optional, Union, Tuple
+from typing import List, Optional, Union, Tuple, TYPE_CHECKING
 
 import geojson
 import numpy as np
 import cv2
```
msokoloff1, line 5: Open cv is an extra dep. You can see them all listed here.
https://github.com/Labelbox/labelbox-python/blob/2c688a7d69bce36984d7237b2d85e8af05aa8e29/setup.py#L27-L30

I would recommend pip installing from your branch from collab without the data package and try importing from annotation_import to see if an error is thrown to make sure this all works.

msokoloff1 (changes requested):
Please make sure to filter out all imports that rely on the data package during type checking.

msokoloff1 (approved):

labelbox/data/annotation_types/geometry/line.py
whistler: Tested that this works on both colab without the `labelbox[data]` as well as in a minimal docker container that does not have `data` dependencies installed.

PR #888, opened 2023-02-08
Revert [AL-4764] changes to notebooks

whistler:
Reverting changes to notebooks since the release does not have the changes yet and docs point to `develop` notebooks which causes public notebooks to be broken.


tytalus (approved):
lgtm

PR #909, opened 2023-02-14
[PTDT-1112] Add delete_unused_feature_schema method

lgluszek:
- calls a new REST endpoint, what should help with SDK4.0 migration
- delete_unused_feature_schema can be used to delete a feature schema

### How to test

If https://github.com/Labelbox/intelligence/pull/13388 is not merged, git checkout lgluszek/PTDT-1112 in `intelligence`

Usage:

```python
from labelbox import Client

client = Client(
    api_key="...",

res = client.delete_unused_feature_schema("...")
```

There should be errors when:

- feature schema is already deleted
- you try to remove a feature schema that is from a different organization
- a feature schema is used in annotations
- a feature schema is added to ontologies


labelbox/client.py
```diff
@@ -95,6 +96,13 @@ def __init__(self,
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
         self._data_row_metadata_ontology = None
+        #
```
kkim-labelbox, line 99: Hashtag typo!

labelbox/client.py
```diff
@@ -95,6 +96,13 @@ def __init__(self,
...
+        self.rest_endpoint = rest_endpoint
+        self.rest_endpoint_headers = {
+            "authorization": "Bearer %s" % self.api_key,
+            'X-User-Agent': 'python-sdk 0.0.0',
```
kkim-labelbox, line 103: Couple things here:
- Could we share the headers between graphql and rest?
- User agent is `0.0.0` here but using `SDK_VERSION` for graphql

labelbox/client.py
```diff
@@ -899,6 +907,30 @@ def create_ontology_from_feature_schemas(self,
...
+
+        Args:
+            feature_schema_id (str): The id of the feature schema to delete
+
+        Returns:
+            True if the feature schema was deleted
+        """
+
+        endpoint = self.rest_endpoint + "/feature-schemas/" + feature_schema_id
+        response = requests.delete(
```
kkim-labelbox, line 922: Nit: It would be good to create a function to handle all types of REST requests, similar to how it's done in SDK 4 Session https://github.com/Labelbox/labelbox-python/blob/sdk4-experimental/labelbox_dev/session.py#L83
since 3 of the PRs make REST requests, but you can feel free to do it in a separate ticket or as a polish item

labelbox/client.py
```diff
@@ -899,6 +907,30 @@ def create_ontology_from_feature_schemas(self,
...
+        Deletes a feature schema if it is not used by any ontologies or annotations
+
+        Args:
+            feature_schema_id (str): The id of the feature schema to delete
+
+        Returns:
+            True if the feature schema was deleted
+        """
+
+        endpoint = self.rest_endpoint + "/feature-schemas/" + feature_schema_id
```
kkim-labelbox, line 921: - Could you URL encode this endpoint?
- Could you also validate `feature_schema_id` to be a string of length 25? (cuid)

labelbox/client.py
```diff
@@ -899,6 +907,30 @@ def create_ontology_from_feature_schemas(self,
         normalized = {'tools': tools, 'classifications': classifications}
         return self.create_ontology(name, normalized, media_type)
 
+    def delete_unused_feature_schema(self, feature_schema_id):
```
kkim-labelbox, line 910: Could you add a test case for this function?

PR #910, opened 2023-02-14
[AL-4860] Add pre-commit hook to format notebooks

whistler:
Adds are pre-commit hook to remove extraneous metadata and consistently format notebook JSON to reduce merge conflicts in future.


kkim-labelbox (approved):
LGTM! Thanks for doing this!

PR #911, opened 2023-02-14
[AL-4860] Remove NDJsonConverter from example notebooks

whistler:
Add test suite to check if notebooks run without errors

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,7 @@ deps =
     pytest < 7.0.0
     pytest-cases
     pytest-rerunfailures
+    nbconvert
 passenv = 
     LABELBOX_TEST_ENVIRON
     LABELBOX_TEST_API_KEY_PROD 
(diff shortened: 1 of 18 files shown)


tox.ini
```diff
@@ -17,6 +18,6 @@ passenv =
     LABELBOX_TEST_API_KEY_ONPREM 
     LABELBOX_TEST_API_KEY_CUSTOM     
     LABELBOX_TEST_GRAPHQL_API_ENDPOINT     
-    DA_GCP_LABELBOX_API_KEY 
+    DA_GCP_LABELBOX_API_KEY
 
-commands = pytest {posargs}
+commands = pytest -k "not notebooks" {posargs}
```
kkim-labelbox, line 23: Curious, does this exclude the `test_notebooks.py` from the automation?

kkim-labelbox (approved):
So clean!! Awesome job

tox.ini
whistler: Yes, it excludes them from the ci job. Eventually we can add it there, but need to sort out certain things like what happens when we reach the max limit of models we are allowed to have in an org.

PR #915, opened 2023-02-15
[AL-4858] Update example notebook imports

whistler:
Updates notebooks to use:

```
import labelbox as lb
import labelbox.types as lb_types
```


kkim-labelbox (approved):
Looks great!

PR #920, opened 2023-02-16
[PTDT-1113] Add delete_unused_ontology method

lgluszek:
- calls a new REST endpoint, what should help with SDK4.0 migration
- delete_unused_ontology can be used to delete an ontology

### How to test

If https://github.com/Labelbox/intelligence/pull/13388 is not merged, git checkout lgluszek/PTDT-1113 in `intelligence`

Usage:

```python
from labelbox import Client

client = Client(
    api_key="...",

res = client.delete_unused_ontology("...")
```

There should be an error when user tries to delete an ontology that is connected to at least 1 project


karenkyang:
Is it possible to show which project is connect to the ontology if it failed to be deleted?

labelbox/client.py
```diff
@@ -931,6 +931,30 @@ def delete_unused_feature_schema(self, feature_schema_id):
...
+        Deletes an ontology if it is not used by any annotations
+
+        Args:
+            ontology_id (str): The id of the ontology to delete
+
+        Returns:
+            True if the ontology was deleted
+        """
+
+        endpoint = self.rest_endpoint + "/ontologies/" + ontology_id
```
kkim-labelbox, line 945: Could you URL encode this and validate `ontology_id` as str of length 25 as well?

labelbox/client.py
```diff
@@ -931,6 +931,30 @@ def delete_unused_feature_schema(self, feature_schema_id):
                 "Failed to delete the feature schema, message: " +
                 response.json()['message'])
 
+    def delete_unused_ontology(self, ontology_id):
```
kkim-labelbox, line 934: Please add a test case for this function as well!

PR #922, opened 2023-02-16
fix exports v2, enable tests

msokoloff1:
* Removed a param in gql and needed to update here.
* Tests were added in because we are starting internal testing and exports should be stable.

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -433,6 +433,8 @@ class Project(DbObject, Updateable, Deletable):
                         _params.get('project_details', False),
                     "includePerformanceDetails":
                         _params.get('performance_details', False),
+                    "includeLabelDetails":
+                        _params.get('label_details', False)
                 },
             }
         }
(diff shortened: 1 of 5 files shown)


mnoszczak (approved):

labelbox/schema/project.py
```diff
@@ -431,8 +430,6 @@ def export_v2(self,
                         _params.get('data_row_details', False),
                     "includeProjectDetails":
                         _params.get('project_details', False),
-                    "includeLabels":
```
mnoszczak, line 434: Shouldn't that be labelDetails instead?
https://github.com/Labelbox/intelligence/blob/develop/apps/api/src/exports-v2/export-data.graphql#L17

labelbox/schema/project.py
```diff
@@ -431,8 +431,6 @@ def export_v2(self,
...
```
mnoszczak, line 434: Here it's still missing, right?

msokoloff1: oops fixed

mnoszczak (approved):

PR #923, opened 2023-02-16
Skip notebook tests and add test dependencies to docker image

whistler:
This is a new test added that was only meant to be run locally, but it is breaking the SDK tests. Adding dependencies for this test and skipping it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,5 @@ pyproj
 pygeotile
 typing-extensions
 pytest-xdist
+nbformat~=5.7.0
+nbconvert~=7.2.6
--- a/examples/test_notebooks.py
+++ b/examples/test_notebooks.py
@@ -64,7 +64,11 @@ run_notebook_paths = [
 ]
 
 
-@pytest.mark.notebooks
+@pytest.mark.skip(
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -14,4 +14,6 @@ imagesize
 pyproj
 pygeotile
 typing-extensions
-pytest-xdist
\ No newline at end of file
+pytest-xdist
+nbformat~=5.7.0
```
vbrodsky, line 18: will this address the codefresh failues like https://g.codefresh.io/build/63ee46d0b77e2079fc9625a2?
But I assume nbformat is not required for development... if this is true, we should do smth like this https://realpython.com/lessons/production-vs-development-dependencies/

PS We need to switch to poetry asap

whistler: Yes, this is meant to fix codefresh failures. nbformat is required to run examples/test_notebooks.py, so its required for development. Good idea to split production vs dev requirements, but isn't that already the case since package requirements are listed in `setup.py` and dev requirements in `requirements.txt`?

kkim-labelbox (approved):

PR #925, opened 2023-02-17
[AL-4860] Remove lb_serializers imports and fix invalid notebook

whistler:
* lb_serialzers imports were unused, removing them
* examples/label_export/images.ipynb was invalid, it was missing some required metadata and not in the format the pre-commit hook expects. Fixing that.

### DIFF
--- a/examples/annotation_import/image.ipynb
+++ b/examples/annotation_import/image.ipynb
@@ -84,8 +84,7 @@
       "metadata": {},
       "source": [
         "import labelbox as lb\n",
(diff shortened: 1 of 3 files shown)


vbrodsky (approved):
Searched in the code outside of notebooks and did not even find `lb_serializers`. Does it make sense?

whistler:
> Searched in the code outside of notebooks and did not even find `lb_serializers`. Does it make sense?
Yes the `as` renames the `serialization` module to `lb_serialization` only for the current file.

PR #926, opened 2023-02-18
[AL-0] Remove custom embeddings

kkim-labelbox:
Remove references to custom embeddings in SDK


whistler:
Why do we not support custom embeddings anymore? Is this a breaking change?

kkim-labelbox:
@whistler Custom embeddings have been removed from the system already. Users currently cannot create custom embeddings metadata, so this isn't a breaking change. I got the sign off from @karenkyang, @sqlboy and @tomislav-peharda to delete it from the SDK

whistler (approved):

manuaero:
FYI: We will be soon supporting custom embeddings in SDK again (once ADV ships).

tomislav-peharda (approved):

kkim-labelbox:
Ongoing discussion on whether to go ahead with this change, because custom embeddings will be reintroduced. However, we do not yet know whether we want to reintroduce it as a dataRow metadata, or its own class.

kkim-labelbox:
Closing PR, since custom embedding will be part of dataRow metadata when it's shipped via ADV.

PR #929, opened 2023-02-20
Vb/remove UUID notebooks al 4862

vbrodsky:
JIRA story: https://labelbox.atlassian.net/browse/AL-4862

This removes uuid generation from notebook ndjson. It is not necessarty any more because we already support filling in uuids in the prediction service

I WAS going to followup, in this PR, and remove uuid's from sdk tests and test harnesses, but that led to many tests failing. Most of our NDJSON serialization classes generate a uuid by default. And this is not a problem, since our effort goes to making our customer life simpler in notebooks in literal NDJSON strings... So the I have left the sdk tests as is, and when we use NDJSON objects, those will come with uuid pre-populated


examples/integrations/detectron2/coco_object.ipynb
```diff
@@ -460,7 +460,7 @@
     {
       "metadata": {},
       "source": [
-        "pairs = get_label_pairs(val_labels, labels_mea, filter_mismatch = True)\n",
+        "pairs = lb.get_label_pairs(val_labels, labels_mea, filter_mismatch = True)\n",
```
whistler, line 463: Functions related to metrics are not included in the `lb` namespace since this has mostly been replaced by autometrics. 

This is how it can be imported instead:
```
from labelbox.data.metrics.group import get_label_pairs
```

You can also leave out this change, I can fix this for the rest of `get_label_pairs` usages in notebooks.

whistler (approved):

examples/integrations/detectron2/coco_object.ipynb
vbrodsky: thanks, @whistler . I will do the import you suggest... o.w. the notebooks it broken (I will test it in python, since this notebook does not run to my knowlesge)

PR #930, opened 2023-02-20
Vb/remove labellist sdk al 4965

vbrodsky:
JIRA story: https://labelbox.atlassian.net/browse/AL-4965

### DIFF
--- a/labelbox/data/annotation_types/__init__.py
+++ b/labelbox/data/annotation_types/__init__.py
@@ -24,8 +24,6 @@ from .data import TextData
 from .data import VideoData
 
 from .label import Label
-
-from .collection import LabelList
 from .collection import LabelGenerator
 
 from .metrics import ScalarMetric
--- a/labelbox/data/annotation_types/__init__.py
+++ b/labelbox/data/annotation_types/__init__.py
@@ -24,6 +24,7 @@ from .data import TextData
 from .data import VideoData
 
 from .label import Label
+from .collection import LabelList
 from .collection import LabelGenerator
 
 from .metrics import ScalarMetric
--- a/labelbox/data/annotation_types/collection.py
+++ b/labelbox/data/annotation_types/collection.py
@@ -295,4 +295,4 @@ class LabelGenerator(PrefetchGenerator):
         return self._process(value)
 
 
-LabelCollection = Union[LabelList, LabelGenerator, Iterable[Label]]
+LabelCollection = Union[LabelGenerator, Iterable[Label]]
--- a/labelbox/data/annotation_types/collection.py
+++ b/labelbox/data/annotation_types/collection.py
@@ -22,10 +22,9 @@ class LabelList:
     Use on smaller datasets.
     """
 
-
     def __init__(self, data: Optional[Iterable[Label]] = None):
         warnings.warn("LabelList is deprecated and will be "
-                    "removed in a future release.")
+                      "removed in a future release.")
 
         if data is None:
             self._data = []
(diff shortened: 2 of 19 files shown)


labelbox/data/metrics/group.py
```diff
@@ -79,8 +79,8 @@ def all_have_key(features: List[FeatureSchema]) -> Tuple[bool, bool]:
     return all_schemas, all_names
 
 
-def get_label_pairs(labels_a: LabelList,
-                    labels_b: LabelList,
+def get_label_pairs(labels_a: list,
```
vbrodsky, line 82: looks like this function is only used in Notebooks, btw

labelbox/data/annotation_types/collection.py
```diff
@@ -179,22 +179,14 @@ def _apply_threaded(self, fns, max_concurrency, *args):
...
-    than the LabelList but will be much more memory efficient.
+    A container for interacting with a large collection of labels. 
+    For a small number of labels, just use a list of Label objects.
     """
 
     def __init__(self, data: Generator[Label, None, None], *args, **kwargs):
         self._fns = {}
         super().__init__(data, *args, **kwargs)
 
-    def as_list(self) -> "LabelList":
```
whistler, line 192: Shoudn't we wait till the 4.0 release to make a breaking change? Until then we could mention that this will be deprecated in the docstring.

vbrodsky: This is why I am doing this PR :) should I revert this change? cc @kkim-labelbox

vbrodsky: However, keep in mind, I have also removed all usages of this method both in the sdk codebase and Notebooks

kkim-labelbox: I think it's sufficient to leave the warning here, and don't remove the method, since it is a breaking change. As long as we remove usage of LabelList and `as_list` in our codebase, that should satisfy our deprecation requirements

labelbox/data/annotation_types/__init__.py
```diff
@@ -24,8 +24,6 @@
 from .data import VideoData
 
 from .label import Label
-
-from .collection import LabelList
```
whistler, line 28: Lets keep this import to keep backwards compatibility?

labelbox/data/annotation_types/collection.py
```diff
@@ -179,10 +181,8 @@ def _apply_threaded(self, fns, max_concurrency, *args):
...
     """
-    A container for interacting with a collection of labels.
-
-    Use this class if you have larger data. It is slightly harder to work with
-    than the LabelList but will be much more memory efficient.
+    A container for interacting with a large collection of labels. 
+    For a small number of labels, just use a list of Label objects.
     """
 
     def __init__(self, data: Generator[Label, None, None], *args, **kwargs):
```
kkim-labelbox, line 188:
```
def as_list(self) -> "LabelList":
        warnings.warn("This method is deprecated and will be "
                      "removed in a future release. LabeList"
                      " class will be deprecated.")
```
Could you fix the spelling on `LabeList`?

labelbox/data/annotation_types/collection.py
vbrodsky: @whistler restored

labelbox/data/annotation_types/collection.py
vbrodsky: @kkim-labelbox done

whistler (approved):

PR #931, opened 2023-02-22
[AL-4886] Introduce ModelSlice

kozikkamil:
Adds the new ModelSlice entity and get method

### DIFF
--- a/labelbox/schema/slice.py
+++ b/labelbox/schema/slice.py
@@ -15,6 +15,7 @@ class Slice(DbObject):
         updated_at (datetime)
         filter (json)
     """
+
     name = Field.String("name")
     description = Field.String("description")
     created_at = Field.DateTime("created_at")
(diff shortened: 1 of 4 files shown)


labelbox/schema/slice.py
```diff
@@ -57,3 +58,40 @@ def get_data_row_ids(self) -> PaginatedCollection:
...
+                        endCursor
+                        hasNextPage
+                    }
+                }
+            }
+        """
+        return PaginatedCollection(
+            client=self.client,
+            query=query_str,
+            params={'id': self.uid},
```
vbrodsky, line 94: according to gql above, $first is required, but I do not see it in the params. Is it ok?

vbrodsky: prob handled in `PaginatedCollection`?

labelbox/schema/slice.py
```diff
@@ -57,3 +58,40 @@ def get_data_row_ids(self) -> PaginatedCollection:
...
+            dereferencing=['getDataRowIdsBySavedQuery', 'nodes'],
+            obj_class=lambda _, data_row_id: data_row_id,
```
vbrodsky, line 96: not following here... according to PaginatedCollection documentation, obj_class is a class of an object...

labelbox/schema/slice.py
kozikkamil: Yup https://github.com/Labelbox/labelbox-python/blob/develop/labelbox/pagination.py#L148

labelbox/schema/slice.py
kozikkamil: That documentation is 4 years old, I think it's outdated. I was actually just copying the code from CatalogSlice here 😅

dubininsergey:
The change looks good, @kozikkamil can we add a basic fuse test for the method? 
I know we can't do e2e test because it's not possible to create Model Slice via SDK, however we could mock HTTP response and make sure SDK behaves in expected way given the expected API response

kozikkamil:
> The change looks good, @kozikkamil can we add a basic fuse test for the method? I know we can't do e2e test because it's not possible to create Model Slice via SDK, however we could mock HTTP response and make sure SDK behaves in expected way given the expected API response

It won't actually be giving any value though. We'll be effectively unit testing PaginatedCollection class, which I am sure has enough coverage already from the e2e tests. It seems like it's not worth to spend the time right now to set that up. ModelSlice has basically no logic in itself

dubininsergey:
> > The change looks good, @kozikkamil can we add a basic fuse test for the method? I know we can't do e2e test because it's not possible to create Model Slice via SDK, however we could mock HTTP response and make sure SDK behaves in expected way given the expected API response
> 
> It won't actually be giving any value though. We'll be effectively unit testing PaginatedCollection class, which I am sure has enough coverage already from the e2e tests. It seems like it's not worth to spend the time right now to set that up. ModelSlice has basically no logic in itself

IMO we should have a fuse test that would catch for example the case where the name of public method has been changed accidentally, or it has been removed at all. If we have something mentioned in SDK docs we should be sure that it's in the code too. 

Not going to block this though

dubininsergey (approved):

PR #938, opened 2023-02-25
[AL-4866] Upsert data rows to model run using global keys

whistler:
Depends on: https://github.com/Labelbox/intelligence/pull/13597


apollonin (approved):

labelbox/schema/model_run.py
```diff
@@ -80,27 +80,40 @@ def upsert_labels(self, label_ids, timeout_seconds=3600):
...
         """ Adds data rows to a Model Run without any associated labels
         Args:
-            data_row_ids (list): data row ids to add to mea
+            data_row_ids (list): data row ids to add to model run
+            global_keys (list): global keys for data rows to add to model run
             timeout_seconds (float): Max waiting time, in seconds.
         Returns:
             ID of newly generated async task
         """
+        if data_row_ids is None == global_keys is None:
```
kkim-labelbox, line 95: Do we need this validation if we're handling it on the backend?

labelbox/schema/model_run.py
```diff
@@ -80,27 +80,40 @@ def upsert_labels(self, label_ids, timeout_seconds=3600):
...
+        if data_row_ids is not None and len(data_row_ids) < 1:
             raise ValueError("Must provide at least one data row id")
 
+        if global_keys is not None and len(global_keys) < 1:
+            raise ValueError("Must provide at least one global key")
+
         mutation_name = 'createMEAModelRunDataRowRegistrationTask'
-        create_task_query_str = """mutation createMEAModelRunDataRowRegistrationTaskPyApi($modelRunId: ID!, $dataRowIds : [ID!]!) {
-          %s(where : { id : $modelRunId}, data : {dataRowIds: $dataRowIds})}
+        create_task_query_str = """mutation createMEAModelRunDataRowRegistrationTaskPyApi($modelRunId: ID!, $dataRowIds: [ID!], $globalKeys: [ID!]) {
```
kkim-labelbox, line 107: Is this a breaking change for customers who are on prev versions of SDK that still have `$dataRowIds: [ID!]!`, if the backend is now accepting `[ID!]`? Might be worth testing

labelbox/schema/model_run.py
cssetian: Might not be a bad idea to validate this on the resolver. Would encourage us to remember to include the validation in the controller in SDK4 as well. In SDK4 we should probably include this into OpenAPI based validation, and implement something like `openapi-typescript-validator` (or not sure if there's anything more native) on the controllers so this is all defined declaratively and we can generate documentation from it.

whistler: This is already being validated in the resolver so I have removed it from here.

labelbox/schema/model_run.py
whistler: I tested this by passing dataRowIds and not globalKeys in the GraphQL call.

whistler: Also ran `tests/integration/annotation_import/test_model_run.py::test_model_run_upsert_data_rows` in develop branch to test backwards compatibility of the API.

PR #939, opened 2023-02-25
[PTDT-1108] Added client method unarchive_root_schema_node

Tim-Kerr:
How to test:
```
from labelbox import Client

client = Client(
    api_key=<LOCALHOST_API_KEY>,

res = client.unarchive_feature_schema_node(<ontology_id>, <feature_schema_id>)
print(res)
```

Given a valid API key that owns the valid ontology and root feature schema node expect `True`

* Important: Test unauthorized access *

Use a valid api key for a user in another org. Attempt to unarchive the root feature schema node belonging to another org. Expect to get a "404" response.

https://www.loom.com/share/379eaebf4f5c483fa0fc0ca6656475d9


labelbox/client.py
```diff
@@ -55,7 +55,8 @@ def __init__(self,
                  api_key=None,
                  endpoint='https://api.labelbox.com/graphql',
                  enable_experimental=False,
-                 app_url="https://app.labelbox.com"):
+                 app_url="https://app.labelbox.com",
+                 rest_endpoint="https://api.labelbox.com/api/v1"):
```
Tim-Kerr, line 59: Added a new `rest_endpoint` param to the constructor that points to the base url for the new api rest endpoints.

labelbox/client.py
```diff
@@ -1385,6 +1396,27 @@ def get_catalog_slice(self, slice_id) -> CatalogSlice:
         res = self.execute(query_str, {'id': slice_id})
         return Entity.CatalogSlice(self, res['getSavedQuery'])
 
+    def unarchive_root_feature_schema_node(
```
Tim-Kerr, line 1399: New unarchive client method reaches out to the api's rest endpoint to unarchive the feature schema in the ontology.

labelbox/client.py
```diff
@@ -1385,6 +1392,27 @@ def get_catalog_slice(self, slice_id) -> CatalogSlice:
...
+                                      root_feature_schema_id: str) -> bool:
+        """
+        Returns true if the root feature schema node was successfully unarchived, false otherwise
+        Args:
+            root_feature_schema_id (str): The ID of the root level feature schema
+            ontology_id (str): The ID of the ontology
+        Returns:
+            bool
+        """
+        ontology_endpoint = self.rest_endpoint + "/ontologies/" + ontology_id + '/feature-schemas/' + root_feature_schema_id + '/unarchive'
```
lgluszek, line 1405: please encode ontology_id and root_feature_schema_id using urllib.parse.quote

labelbox/client.py
```diff
@@ -1385,6 +1392,27 @@ def get_catalog_slice(self, slice_id) -> CatalogSlice:
...
+        response = requests.patch(
+            ontology_endpoint,
+            headers=self.rest_endpoint_headers,
+        )
+        if response.status_code == 200:
+            return response.json()['unarchived']
+        else:
+            raise labelbox.exceptions.LabelboxError(
+                "Failed unarchive root feature schema node: ", response.text)
```
lgluszek, line 1414: This is the convention other methods follow

```suggestion
                "Failed unarchive root feature schema node, message: ", response.text)
```

labelbox/client.py
Tim-Kerr: [Updated](https://github.com/Labelbox/labelbox-python/pull/939/commits/3f98a70ed7bb65f28b22272ef9b55fd7d0281ba5)

labelbox/client.py
```diff
@@ -1385,6 +1392,27 @@ def get_catalog_slice(self, slice_id) -> CatalogSlice:
...
+            ontology_id (str): The ID of the ontology
+        Returns:
+            bool
+        """
+        ontology_endpoint = self.rest_endpoint + "/ontologies/" + ontology_id + '/feature-schemas/' + root_feature_schema_id + '/unarchive'
+        response = requests.patch(
+            ontology_endpoint,
+            headers=self.rest_endpoint_headers,
+        )
+        if response.status_code == 200:
```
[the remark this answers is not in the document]
Tim-Kerr: [Udpated](https://github.com/Labelbox/labelbox-python/pull/939/commits/734dd3bb0a3b4ff0f15b3c04b7d14454c5eb71ca)

labelbox/client.py
Tim-Kerr: [Updated](https://github.com/Labelbox/labelbox-python/pull/939/commits/4764f3e85dc2ce7fcff2f038389fd8790c392af2)

labelbox/client.py
```diff
@@ -1574,6 +1574,30 @@ def get_catalog_slice(self, slice_id) -> CatalogSlice:
         res = self.execute(query_str, {'id': slice_id})
         return Entity.CatalogSlice(self, res['getSavedQuery'])
 
+    def unarchive_feature_schema_node(self, ontology_id: str,
+                                      root_feature_schema_id: str) -> bool:
+        """
+        Returns true if the root feature schema node was successfully unarchived, false otherwise
```
kkim-labelbox, line 1580: Can we specify in the docstring here that only root level feature schemas are unarchivable?

labelbox/client.py
```diff
@@ -1574,6 +1574,30 @@ def get_catalog_slice(self, slice_id) -> CatalogSlice:
...
+        """
+        ontology_endpoint = self.rest_endpoint + "/ontologies/" + urllib.parse.quote(
+            ontology_id) + '/feature-schemas/' + urllib.parse.quote(
+                root_feature_schema_id) + '/unarchive'
+        response = requests.patch(
+            ontology_endpoint,
+            headers=self.headers,
+        )
+        if response.status_code == requests.codes.ok:
+            return response.json()['unarchived']
```
kkim-labelbox, line 1595: It's probably fine to return `None` here, since it'll just return true when successfully unarchived anyway. Unless there's a case where unarchived = false

kkim-labelbox (commented):
Could you add a test case for this function? Other than that, left a couple small comments!

labelbox/client.py
Tim-Kerr: [Updated](https://github.com/Labelbox/labelbox-python/pull/939/commits/85e77f6ca6072d90e71ea6cab45aa998cbb6775f)

labelbox/client.py
Tim-Kerr: It'll return false if the user passes in a root feature schema id that does not belong to the ontology.

kkim-labelbox: If the response is false, maybe we can just raise an exception then with an error message indicating that the unarchive was unsuccessful? I know the docstring specifies the behavior, but customers may not necessarily check for the output of this function, so it may be safer to throw an explicit message indicating the behavior

Tim-Kerr: [Updated](https://github.com/Labelbox/labelbox-python/pull/939/commits/2c45ec7cf3362b941a2630c6ef6c1ffde7b2fc00)

labelbox/client.py
```diff
@@ -1578,6 +1578,35 @@ def get_catalog_slice(self, slice_id) -> CatalogSlice:
...
+        ontology_endpoint = self.rest_endpoint + "/ontologies/" + urllib.parse.quote(
+            ontology_id) + '/feature-schemas/' + urllib.parse.quote(
+                root_feature_schema_id) + '/unarchive'
+        response = requests.patch(
+            ontology_endpoint,
+            headers=self.headers,
+        )
+        if response.status_code == requests.codes.ok:
+            unarchived = bool(response.json()['unarchived'])
+            if (unarchived == False):
```
kkim-labelbox, line 1601: You cna do:
```
if not unarchived
```

Tim-Kerr: [Updated](https://github.com/Labelbox/labelbox-python/pull/939/commits/65b814bf8d3e78793c4d83a26ea2c4cb3248fe39)

kkim-labelbox (approved):

PR #953, opened 2023-03-01
[AL-4867] Create batch using global keys

whistler:
Depends on: https://github.com/Labelbox/intelligence/pull/13652

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -669,16 +669,20 @@ class Project(DbObject, Updateable, Deletable):
         timestamp = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
         self.update(setup_complete=timestamp)
 
-    def create_batch(self,
-                     name: str,
-                     data_rows: List[Union[str, DataRow]],
-                     priority: int = 5,
-                     consensus_settings: Optional[Dict[str, float]] = None):
-        """Create a new batch for a project. Batches is in Beta and subject to change
+    def create_batch(
+        self,
+        name: str,
+        data_rows: Optional[List[Union[str, DataRow]]] = None,
+        priority: int = 5,
+        consensus_settings: Optional[Dict[str, float]] = None,
+        global_keys: Optional[List[str]] = None,
+    ):
+        """Create a new batch for a project. One of `global_keys` or `data_rows` must be provided but not both.
 
         Args:
             name: a name for the batch, must be unique within a project
-            data_rows: Either a list of `DataRows` or Data Row ids
+            data_rows: Either a list of `DataRows` or Data Row ids. 
+            global_keys: global keys for data rows to add to the batch. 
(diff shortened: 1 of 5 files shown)


labelbox/schema/project.py
```diff
@@ -688,35 +690,40 @@ def create_batch(self,
...
-            else:
-                raise ValueError("You can DataRow ids or DataRow objects")
+        if data_rows is not None:
+            for dr in data_rows:
+                if isinstance(dr, Entity.DataRow):
+                    dr_ids.append(dr.uid)
+                elif isinstance(dr, str):
+                    dr_ids.append(dr)
+                else:
+                    raise ValueError("You can DataRow ids or DataRow objects")
```
apollonin, line 700: I know it is unrelated to this PR, but should pretty easy to improve. The error text is weird and would be cool to add "provide" word in there =)

labelbox/schema/project.py
```diff
@@ -1260,38 +1270,47 @@ def _is_url_valid(url: Union[str, Path]) -> bool:
...
+            queryAllDataRowsHaveBeenProcessed(dataRowIds:$%s, globalKeys:$%s) {
                 allDataRowsHaveBeenProcessed
            }
-        }""" % (data_row_ids_param, data_row_ids_param)
+        }""" % (data_row_ids_param, global_keys_param, data_row_ids_param,
+                global_keys_param)
 
-        params = {}
-        params[data_row_ids_param] = data_row_ids
+        params = {
```
vbrodsky, line 1310: do we have to send both if only one will be ever present?

labelbox/schema/project.py
```diff
@@ -671,14 +671,16 @@ def setup(self, labeling_frontend, labeling_frontend_options) -> None:
 
     def create_batch(self,
                      name: str,
-                     data_rows: List[Union[str, DataRow]],
+                     data_rows: List[Union[str, DataRow]] = None,
+                     global_keys: List[str] = None,
                      priority: int = 5,
                      consensus_settings: Optional[Dict[str, float]] = None):
-        """Create a new batch for a project. Batches is in Beta and subject to change
+        """Create a new batch for a project. One of `global_keys` or `data_rows` must be provided but not both.
```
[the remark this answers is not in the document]
whistler: This is validated on the backend.

labelbox/schema/project.py
whistler: Good catch, updated!

labelbox/schema/project.py
whistler: I'll see if I can pass just one.

whistler: Updated so that only one of globalKeys or dataRowIds is passed.

kkim-labelbox (approved):
LGTM!

whistler:
The one test that doesn't pass is also broken on develop, so merging this.

PR #964, opened 2023-03-02
addind_queue_dataset

paultancre:
Based on the prediction import the expected project queue set is currently should be dataset (since the model is querying dataset to make prediction on). This would need to be rework for batch(es)


paulnoirel (approved):
LGTM

PR #969, opened 2023-03-03
fix test by waiting until data rows are processed

attila-papai:


tests/integration/test_project.py
```diff
@@ -43,7 +43,9 @@ def test_project(client, rand_gen):
 
 
 def test_project_export_v2(configured_project_with_label):
-    project, _, _, label = configured_project_with_label
+    project, _, data_row, label = configured_project_with_label
+    project._wait_until_data_rows_are_processed(
+        [data_row.id], wait_processing_max_seconds=3600, sleep_interval=5)
```
attila-papai, line 48: should we just put this code after this line? 🤔 
https://github.com/Labelbox/labelbox-python/blob/63451820f9a9a9941161bef658a0ef5fe702c949/tests/integration/conftest.py#L341

tytalus (approved):
lgtm

tests/integration/test_project.py
whistler: Maybe we should instead call `_wait_until_data_rows_are_processed()` `inside export_v2()`, thats the pattern we use for `create_batch()`. What do you think @kkim-labelbox?

attila-papai: `_wait_until_data_rows_are_processed` needs a list of data row ids that we don't have when calling an export

whistler (approved):

tests/integration/test_project.py
whistler: I think it would be good to keep it in the test instead of the fixture, so we know in which cases the user may have to wait for data_rows to be processed.

Makefile
```diff
@@ -1,11 +1,10 @@
 
-build:
+build-image:
```
attila-papai, line 2: `make build` is a reserved command, it won't run this here

apollonin (approved):

PR #976, opened 2023-03-08
[AL-5192] Add last_activity_at / label_created_at export v2 support

mnoszczak:
https://labelbox.atlassian.net/browse/AL-5192

Added support for `last_activity_at / label_created_at` filters which are transformed into search query params before making GQL call

### DIFF
--- a/labelbox/schema/export_filters.py
+++ b/labelbox/schema/export_filters.py
@@ -2,9 +2,11 @@ import sys
 
 from typing import Optional
 if sys.version_info >= (3, 8):
-    from typing import TypedDict, Tuple
+    from typing import TypedDict
 else:
     from typing_extensions import TypedDict
+
+from typing import Tuple
 
 
 class ProjectExportFilters(TypedDict):
--- a/labelbox/schema/export_filters.py
+++ b/labelbox/schema/export_filters.py
@@ -1,0 +1,12 @@
+import sys
+
+from typing import Optional
+if sys.version_info >= (3, 8):
+    from typing import TypedDict, Tuple
+else:
+    from typing_extensions import TypedDict
+
+
+class ProjectExportFilters(TypedDict):
+    label_created_at: Optional[Tuple[str, str]]
+    last_activity_at: Optional[Tuple[str, str]]
--- a/labelbox/schema/export_filters.py
+++ b/labelbox/schema/export_filters.py
@@ -11,4 +11,18 @@ from typing import Tuple
 
 class ProjectExportFilters(TypedDict):
     label_created_at: Optional[Tuple[str, str]]
+    """ Date range for labels created at
+    Formatted "YYYY-MM-DD" or "YYYY-MM-DD hh:mm:ss"
+    Examples: 
+    >>>   ["2000-01-01 00:00:00", "2050-01-01 00:00:00"]
+    >>>   [None, "2050-01-01 00:00:00"]
+    >>>   ["2000-01-01 00:00:00", None]
+    """
     last_activity_at: Optional[Tuple[str, str]]
+    """ Date range for last activity at
+    Formatted "YYYY-MM-DD" or "YYYY-MM-DD hh:mm:ss"
+    Examples: 
+    >>>   ["2000-01-01 00:00:00", "2050-01-01 00:00:00"]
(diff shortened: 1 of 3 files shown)


labelbox/schema/export_filters.py
```diff
@@ -0,0 +1,12 @@
...
+
+from typing import Optional
+if sys.version_info >= (3, 8):
+    from typing import TypedDict, Tuple
+else:
+    from typing_extensions import TypedDict
+
+
+class ProjectExportFilters(TypedDict):
+    label_created_at: Optional[Tuple[str, str]]
```
kkim-labelbox, line 11: This looks like a good place to add a docstring to describe each of these filter items in detail, and maybe put the behavior for how we fetch the timezones as well

mnoszczak: I think we could do a follow up PR with improved docs. At this point we're linking to the docs through URL in export_v2's docstring, which should give good guidance. 

I don't want to block the release with the docs.

labelbox/schema/project.py
```diff
@@ -46,6 +47,20 @@
 logger = logging.getLogger(__name__)
 
 
+def _validate_datetime(string_date: str) -> bool:
```
vbrodsky, line 50: Is there a utils etc package to move it to?

labelbox/schema/project.py
```diff
@@ -400,16 +402,27 @@ def _validate_datetime(string_date: str) -> bool:
...
-    """
-    Creates a project run export task with the given params and returns the task.
-    
-    >>>    export_task = export_v2("my_export_task", filter={"media_attributes": True})
-    
-    """
-
     def export_v2(self,
                   task_name: Optional[str] = None,
+                  filters: Optional[ProjectExportFilters] = None,
```
vbrodsky, line 407: is the intend of default `= None` to allow to pass filters value as `None` as opposed to just missing variable `filters`?  If not, a default is not required. Also see here https://docs.python.org/3.9/library/typing.html#typing.Optional 
PS This seems to be the pattern we have been following, so just curious

labelbox/schema/project.py
```diff
@@ -420,15 +433,33 @@ def export_v2(self,
             "label_details": False
         })
 
+        _filters = filters or ProjectExportFilters({
```
vbrodsky, line 436: another nitpick, I like to wrap those in a default fun like
```
class ProjectExportFilters
  @classmethod
  def null_filter():
   self({...})
```

labelbox/schema/project.py
```diff
@@ -420,15 +433,33 @@ def export_v2(self,
...
+            "last_activity_at": None,
+            "label_created_at": None
+        })
+
+        def _get_timezone() -> str:
+            timezone_query_str = """query CurrentUserPyApi { user { timezone } }"""
```
vbrodsky, line 442: again, this is prob more for my own education, but don't we have some sort of sdk-side authenticated user object that should already expose time sone?

vbrodsky (approved):
had some nitpicks and questions

labelbox/schema/project.py
mnoszczak: Nope - right now we're using that only in project file scope, I would move it into utils once there's use case outside of that file

labelbox/schema/project.py
mnoszczak: I've looked into the codebase myself and asked @kkim-labelbox about that - looks like we haven't had an use case to fetch timezone before.

labelbox/schema/project.py
mnoszczak: If the default value is not provided, a function call fails with:
```
 export_v2() missing 2 required positional arguments: 'task_name' and 'params'
```

labelbox/schema/project.py
```diff
@@ -446,6 +477,84 @@ def export_v2(self,
                 },
             }
         }
+
+        if "last_activity_at" in _filters and _filters[
```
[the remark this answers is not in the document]
mnoszczak: I've started with that approach, but if I use unsafe getter, `mypy` will be bothered with this piece:
```
            values = _filters['last_activity_at']
            start, end = values
```
Perhaps `mypy` is not good enough with typings?

PR #979, opened 2023-03-09
[AL-4990] Add media type override

mnoszczak:
Adds support for media type overrides filed

### DIFF
--- a/labelbox/schema/export_params.py
+++ b/labelbox/schema/export_params.py
@@ -1,6 +1,8 @@
 import sys
 
 from typing import Optional
+
+from labelbox.schema.media_type import MediaType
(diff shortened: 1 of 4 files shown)


labelbox/schema/export_params.py
```diff
@@ -11,6 +13,7 @@ class DataRowParams(TypedDict):
     data_row_details: Optional[bool]
     metadata_fields: Optional[bool]
     attachments: Optional[bool]
+    media_type_override: Optional[MediaType]
```
kkim-labelbox, line 16: Can you add a test case with media_type_override in the exporter?

attila-papai (approved):

tytalus (approved):
lgtm

PR #987, opened 2023-03-12
Vb/document entity annotation al 5106

vbrodsky:
Story: https://labelbox.atlassian.net/browse/AL-5106
Adding support for `DocumentEntity` proper annotation type


labelbox/data/annotation_types/ner/document_entity.py
```diff
@@ -0,0 +1,20 @@
...
+
+class DocumentTextSelection(BaseModel):
+    tokenIds: List[str]
+    groupId: str
+    page: int
+
+    @validator("page")
+    def validate_page(cls, v):
+        if v < 1:
+            raise ValueError("Page must be greater than 1")
```
mnoszczak, line 14: Do we need this validation? If you try to import text selection that doesn't belong to a page, it will be rejected by the import pipeline anyway. I think that one of the stretch goals was to support 0-indexed documents as well (but I don't know where that conversation ended up).



This seems to be duplicated (and incomplete as import pipeline checks for way more things) logic.

tests/integration/annotation_import/conftest.py
```diff
@@ -152,14 +159,42 @@ def configured_project_pdf(client, ontology, rand_gen, pdf_url):
...
+    data_row = dataset.create_data_row(pdf_entity_row_data)
+    data_row_ids.append(data_row.uid)
+    yield dataset, data_row_ids
+    dataset.delete()
+
+
+@pytest.fixture
+def configured_project_pdf_entity(client, ontology, rand_gen,
+                                  dataset_pdf_entity):
+    project = client.create_project(name=rand_gen(str),
```
whistler, line 183: Couldn't we reuse `configured_project_without_data_rows` fixture here? Also, we should prefer `QueueMode.Batch` over `QueueMode.Dataset`, `QueueMode.Dataset` is an older paradigm.

labelbox/data/annotation_types/ner/document_entity.py
vbrodsky: our sdk documentation here https://docs.labelbox.com/reference/import-document-annotations#entity state page is one-indexed and prediction service does not check for this (it does checks like page counts etc)... @kkim-labelbox any thoughts?

tests/integration/annotation_import/conftest.py
vbrodsky: will look into both

tests/integration/annotation_import/conftest.py
```diff
@@ -152,14 +159,42 @@ def configured_project_pdf(client, ontology, rand_gen, pdf_url):
...
+    data_row_ids.append(dataset.create_data_row(pdf_url).uid)
     project.datasets.connect(dataset)
     project.data_row_ids = data_row_ids
     yield project
     project.delete()
     dataset.delete()
 
 
+@pytest.fixture
+def dataset_pdf_entity(client, rand_gen, pdf_entity_row_data):
```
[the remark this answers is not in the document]
vbrodsky: done

examples/annotation_import/pdf.ipynb
whistler, line 1: Interesting that we're getting such a large diff. Maybe some of our commits don't have databooks running, perhaps we should add databooks as a CI job too. No action needed for now, I've made a note for myself.

labelbox/data/annotation_types/ner/document_entity.py
```diff
@@ -0,0 +1,21 @@
+from typing import List
+
+from pydantic import BaseModel, validator
+
+
+class DocumentTextSelection(BaseModel):
+    tokenIds: List[str]
```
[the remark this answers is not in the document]
whistler: Using camel case for python object properties is not the best user experience. I believe `labelbox.utils._CamelCaseMixin` can be used to keep properties snake case but serialization to and deserialize from camel case. I just read this post to understand this better: https://medium.com/analytics-vidhya/camel-case-models-with-fast-api-and-pydantic-5a8acb6c0eee

whistler:
Should we also have NDJSON serialization tests similar to other examples under `tests/data/serialization/ndjson`?

vbrodsky:
> Should we also have NDJSON serialization tests similar to other examples under `tests/data/serialization/ndjson`?

good idea

examples/annotation_import/pdf.ipynb
whistler: yes, thats a good idea, I've noticed using jupyter creates huge diffs compared to vscode

labelbox/data/annotation_types/ner/document_entity.py
```diff
@@ -2,10 +2,12 @@
 
 from pydantic import BaseModel, validator
 
+from labelbox.utils import _CamelCaseMixin
 
-class DocumentTextSelection(BaseModel):
-    tokenIds: List[str]
-    groupId: str
+
+class DocumentTextSelection(_CamelCaseMixin, BaseModel):
```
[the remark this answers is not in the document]
vbrodsky: k

examples/annotation_import/pdf.ipynb
kkim-labelbox: It may also be because someone working on this notebook did not have the pre-commit hooks installed.

tests/integration/annotation_import/conftest.py
vbrodsky: this is done

examples/annotation_import/pdf.ipynb
vbrodsky: totally possible as Andrea has created this notebook

labelbox/data/annotation_types/ner/document_entity.py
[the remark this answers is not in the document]
vbrodsky: yeah... but pylance complaince, see the discussion here https://github.com/microsoft/pylance-release/issues/361 too

I could just do `class DocumentTextSelection(_CamelCaseMixin):` but will be a little less readable and a little more correct from Python typing point of view?

labelbox/data/serialization/ndjson/objects.py
```diff
@@ -372,6 +372,36 @@ def from_common(cls,
                    confidence=confidence)
 
 
+class NDDocumentEntity(NDBaseObject, ConfidenceMixin):
+    name: str
+    text_selections: List[DocumentTextSelection]
+
+    def to_common(self) -> DocumentEntity:
+        return TextEntity(name=self.name, text_selections=self.text_selections)
```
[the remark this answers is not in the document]
vbrodsky: done

whistler (approved):
Great work! Thanks for the updates!

PR #989, opened 2023-03-13
[AL-5044] Last activity end

tytalus:


tests/integration/test_export.py
```diff
@@ -142,4 +145,9 @@ def test_export_filtered_activity(client,
         last_activity_start=(datetime.datetime.now() +
                              datetime.timedelta(days=1)).strftime("%Y-%m-%d"),
     )
+
+    empty_export = project.export_labels(
+        download=True,
+        last_activity_end=(datetime.datetime.now() -
+                           datetime.timedelta(days=1)).strftime("%Y-%m-%d"))
```
mnoszczak, line 152: NIT: Could also add `%Y-%m-%d %h:%m` to make sure we support both formats

mnoszczak (approved):

PR #990, opened 2023-03-13
[AL-5103] Add new data type classes to annotation types

whistler:
Adds new classes for missing data types. The data types classes do not provide additional helper functions like existing subclasses of BaseData, but are required to allow creation of Labels through using Python objects. 

A test has been added to ensure that LabelImport can use these data classes.


tests/integration/annotation_import/conftest.py
```diff
@@ -11,6 +11,167 @@
...
+            "https://lb-test-data.s3.us-west-1.amazonaws.com/image-samples/sample-image-1.jpg",
+        "global_key":
+            f"https://lb-test-data.s3.us-west-1.amazonaws.com/image-samples/sample-image-1.jpg-{rand_gen(str)}",
+        "media_type":
+            "IMAGE",
+    }
+
+
+@pytest.fixture()
+def document_data_row(rand_gen):
```
vbrodsky, line 95: I have a very similar fixture in my PR, as you know, so this will be a dupe..no action I suppose, when both are merged we can get rid of one

tests/integration/annotation_import/test_label_import.py
```diff
@@ -110,3 +113,42 @@ def test_wait_till_done(client, configured_project, predictions):
...
+        client, project_id, f'test-import-{data_type_string}', labels)
+    label_import.wait_until_done()
+
+    assert label_import.state == AnnotationImportState.FINISHED
+    assert len(label_import.errors) == 0
+    exported_labels = configured_project.export_labels(download=True)
+    objects = exported_labels[0]['Label']['objects']
+    classifications = exported_labels[0]['Label']['classifications']
+    assert len(objects) + len(classifications) == len(labels)
+    DataRow.bulk_delete([data_row])
```
kkim-labelbox, line 154: any reason for using `DataRow.bulk_delete` over `data_row.delete()`?

whistler: I can't find the `data_row.delete()` method does it exist?

kkim-labelbox (commented):
Can we put the tests inside a new file, since that's the subject of the tests? `test_data_types.py`

Also, can we add a case for VideoData? I see the fixtures are set up for it already

tests/integration/annotation_import/conftest.py
whistler: I've combined the two fixtures.

kkim-labelbox (approved):

PR #994, opened 2023-03-14
[AL-4940]Add initial data row embedding support for create/delete/getAll

shaochen-lb:
This is an initial step to integrate ADV custom embedding with SDK:

1. Add list/create/delete method
2. Return Json format as results as list method.
3. Noted that ADV has a service path as https://api.labelbox.com/adv/adv/v1/...

TODO: add import method for ndjson(a bigger PR)

Ran testing on Staging:
Screenshot 2023-03-14 at 3 28 59 PM

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -21,6 +21,7 @@ from labelbox.orm import query
 from labelbox.orm.db_object import DbObject
 from labelbox.orm.model import Entity
 from labelbox.pagination import PaginatedCollection
+from labelbox.schema.data_row_embedding import DataRowEmbedding
 from labelbox.schema.data_row_metadata import DataRowMetadataOntology
 from labelbox.schema.dataset import Dataset
 from labelbox.schema.enums import CollectionJobStatus
@@ -58,7 +59,8 @@ class Client:
                  endpoint='https://api.labelbox.com/graphql',
                  enable_experimental=False,
                  app_url="https://app.labelbox.com",
-                 rest_endpoint="https://api.labelbox.com/api/v1"):
+                 rest_endpoint="https://api.labelbox.com/api/v1",
+                 adv_rest_endpoint="https://api.labelbox.com/adv"):
         """ Creates and initializes a Labelbox Client.
 
         Logging is defaulted to level WARNING. To receive more verbose
@@ -92,6 +94,7 @@ class Client:
         self.app_url = app_url
         self.endpoint = endpoint
         self.rest_endpoint = rest_endpoint
+        self.adv_rest_endpoint = adv_rest_endpoint
         self.headers = {
             'Accept': 'application/json',
             'Content-Type': 'application/json',
@@ -99,6 +102,7 @@ class Client:
             'X-User-Agent': f'python-sdk {SDK_VERSION}'
         }
         self._data_row_metadata_ontology = None
+        self._data_row_embedding = None
 
     @retry.Retry(predicate=retry.if_exception_type(
         labelbox.exceptions.InternalServerError,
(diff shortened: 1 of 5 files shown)


labelbox/client.py
```diff
@@ -58,7 +59,8 @@ def __init__(self,
                  endpoint='https://api.labelbox.com/graphql',
                  enable_experimental=False,
                  app_url="https://app.labelbox.com",
-                 rest_endpoint="https://api.labelbox.com/api/v1"):
+                 rest_endpoint="https://api.labelbox.com/api/v1",
+                 adv_rest_endpoint="https://api.labelbox.com/adv"):
```
cssetian, line 63: I don't think we want to expose the concept of 'adv' vs 'api' externally, we should only use 1 base gateway api route externally and then just manage different route suffixes per area of functionality

cssetian: The routes for embeddings should probably be something like 
`https://api.labelbox.com/api/v1/data-rows/embedding:POST|PUT|DELETE`, and follow a similar pattern to how we expose other external datarow routes in SDK4 (as something like `https://api.labelbox.com/api/v1/data-rows/`)
cc @kkim-labelbox

shaochen-lb: It should be already there at expressjs, I will check it out later.

tomislav-peharda: yep, that sounds right. we should have one proxy route which will then decide where to forward a request

shaochen-lb: I think our offline discussion is that since Embedding is new and doesn't have migration issue as data row, so we can use ADV endpoint directly for now.

sqlboy: > The routes for embeddings should probably be something like
> `https://api.labelbox.com/api/v1/data-rows/embeddings:POST|PUT|DELETE`, and follow a similar pattern to how we expose other external datarow routes in SDK4 (as something like `https://api.labelbox.com/api/v1/data-rows/`)

@cssetian that will route the call through the API service, which is what we do not want to do, its an architectural layering violation.  IE, whenver ADV adds an endpoint, that doesn't mean we add a duplicate endpoint in the API service.

labelbox/schema/data_row_embedding.py
```diff
@@ -0,0 +1,60 @@
...
+            raise labelbox.exceptions.LabelboxError(
+                "Failed to create custom embedding, message: ", response.text)
+
+    def get_all_custom_embeddings(
+            self) -> List[Dict[str, Union[str, List[Any]]]]:
+        adv_embedding_endpoint_url = f"{self._client.adv_rest_endpoint}{ADV_EMDBEDDING_PATH}"
+        response = requests.get(url=adv_embedding_endpoint_url,
+                                headers=self._client.headers)
+
+        if response.status_code == requests.codes.ok:
```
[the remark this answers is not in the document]
shaochen-lb: We can borrow the `_make_request`

tests/integration/conftest.py
```diff
@@ -78,6 +78,20 @@ def rest_url(environ: str) -> str:
     return 'http://host.docker.internal:8080/api/v1'
 
 
+def adv_rest_url(environ: str) -> str:
```
sqlboy, line 81: im not we we going to expose this publically

labelbox/client.py
sqlboy: Note that after talking to chris, we actually have to setup a secondary endpoint system in nginx.

sqlboy:
I think we can close this PR, until such time that custom embeddings are ready for use.

PR #996, opened 2023-03-15
[AL-5107] Vb/conversational annotation al 5107 2

vbrodsky:
story: https://labelbox.atlassian.net/browse/AL-5107

### DIFF
--- a/labelbox/data/serialization/ndjson/objects.py
+++ b/labelbox/data/serialization/ndjson/objects.py
@@ -505,8 +505,8 @@ class NDObject:
 # NOTE: Deserialization of subclasses in pydantic is a known PIA, see here https://blog.devgenius.io/deserialize-child-classes-with-pydantic-that-gonna-work-784230e1cf83
 # I could implement the registry approach suggested there, but I found that if I list subclass (that has more attributes) before the parent class, it works
 # This is a bit of a hack, but it works for now
-NERTextType = Union[NDConversationEntity, NDTextEntity]
+NDEntityType = Union[NDConversationEntity, NDTextEntity]
 NDObjectType = Union[NDLine, NDPolygon, NDPoint, NDRectangle, NDMask,
-                     NERTextType, NDDocumentEntity]
+                     NDEntityType, NDDocumentEntity]
 
 NDFrameObjectType = NDFrameRectangle, NDFramePoint, NDFrameLine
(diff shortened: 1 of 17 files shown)


labelbox/data/serialization/ndjson/objects.py
```diff
@@ -471,7 +502,11 @@ def lookup_object(
         return result
 
 
+# NOTE: Deserialization of subclasses in pydantic is a known PIA, see here https://blog.devgenius.io/deserialize-child-classes-with-pydantic-that-gonna-work-784230e1cf83
+# I could implement the registry approach suggested there, but I found that if I list subclass (that has more attributes) before the parent class, it works
+# This is a bit of a hack, but it works for now
+NERTextType = Union[NDConversationEntity, NDTextEntity]
```
kkim-labelbox, line 508: `NDEntityType` or `NDNerType` would follow the naming pattern better here

tests/integration/annotation_import/test_conversation_import.py
```diff
@@ -0,0 +1,43 @@
...
+from labelbox.data.annotation_types.data.text import TextData
+from labelbox.data.annotation_types.ner import ConversationEntity
+
+from labelbox.schema.annotation_import import MALPredictionImport
+
+
+def test_conversation_entity(client, configured_project_without_data_rows,
+                             dataset_conversation_entity, rand_gen):
+
+    conversation_entity_annotation = ConversationEntity(name="named-entity",
```
kkim-labelbox, line 13: Don't think you need `name` here, since the name is set in `ObjectAnnotation`

kkim-labelbox (commented):
Left some comments. Also, I see that `DocumentEntity` has a `name` field, which doesn't seem necessary, since `ObjectAnnotation` already has a name tied to it. Could you make that change in this PR as well? Will probably need to update the tests as well

tests/integration/annotation_import/test_conversation_import.py
```diff
@@ -0,0 +1,43 @@
...
+                  ]))
+
+    import_annotations = MALPredictionImport.create_from_objects(
+        client=client,
+        project_id=configured_project_without_data_rows.uid,
+        name=f"import {str(uuid.uuid4())}",
+        predictions=labels)
+    import_annotations.wait_until_done()
+
+    assert import_annotations.errors == []
```
whistler, line 43: It might be a good idea to check that we can retrieve the newly created labels. For video annotations, I saw that there are no import errors but the label does not exist.

vbrodsky: k

kkim-labelbox (approved):
LGTM!

PR #997, opened 2023-03-15
Set typeguard version as 2.13.2 to fix tox error

shaochen-lb:
as title


kkim-labelbox (approved):

shaochen-lb:
There are snyk failure on license and security, which I don't have access to check yet, request waiting to be approved.

PR #1005, opened 2023-03-17
[AL-5109][AL-5110] Add Video and Raster Segmentation Annotation Kinds

whistler:
* Create new module for Video and DICOM annotation kinds
* Add `VideoMaskAnnotation` and `DICOMMaskAnnotation`
* Update `NDJsonConverter` to support new annotation kinds
* Update case conversion util to produce acronyms in uppercase i.e `URI` instead of `Uri`.

Note: This PR includes changes from #999, please review #999 first.

### DIFF
--- a/tests/data/serialization/ndjson/test_dicom.py
+++ b/tests/data/serialization/ndjson/test_dicom.py
@@ -1,7 +1,11 @@
 from copy import copy
+import pytest
 import labelbox.types as lb_types
 from labelbox.data.serialization import NDJsonConverter
-from labelbox.data.serialization.ndjson.objects import NDDicomSegments, NDDicomSegment, NDDicomLine, NDSegments, NDSegment, NDFrameLine
+from labelbox.data.serialization.ndjson.objects import NDDicomSegments, NDDicomSegment, NDDicomLine
+"""
+Polyline test data
+"""
 
 dicom_polyline_annotations = [
     lb_types.DICOMObjectAnnotation(uuid="dicom-object-uid",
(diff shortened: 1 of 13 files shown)


tests/data/serialization/ndjson/test_dicom.py
```diff
@@ -0,0 +1,183 @@
...
+
+video_mask_label = lb_types.Label(data=lb_types.VideoData(uid="test-uid"),
+                                  annotations=[video_mask_annotation])
+
+dicom_mask_annotation_ndjson = copy(video_mask_annotation_ndjson)
+dicom_mask_annotation_ndjson['groupKey'] = 'axial'
+dicom_mask_annotation_ndjson['name'] = 'dicom_mask'
+
+
+def test_serialize_dicom_polyline_annotation():
```
vbrodsky, line 117: I really prefer that we start comparing class types and actual attributes instead of serialize/deserialize stuff. I have discovered some tests passed while the actual object we deserialized into was the wrong type (I think our test fixture was wrong in that case).
Do we agree on this? Perhaps do not need to do for this PR if too much work

vbrodsky (commented):
I can approve this once I have approved a dependent PR

tests/data/serialization/ndjson/test_dicom.py
whistler: I agree with you, comparing python objects instead of only using a json to deserialize then serialize catches more problems.

vbrodsky (approved):

PR #1009, opened 2023-03-17
Fix annotation data type coersion by Pydantic

whistler:
When creating a `Label`, the annotation data type class was being changed and the global_key field was disappearing. Example:

```
label = []
label.append(
  lb_types.Label(
    data=lb_types.ConversationData(
      global_key=global_key
    ),
    annotations=[
      ner_annotation
    ]
  )
)
```
Output
```
[Label(uid=None, data=VideoData(file_path=None,frames=None,url=None), annotations=[ObjectAnnotation(confidence=None, name='ner', feature_schema_id=None, extra={}, value=ConversationEntity(start=0, end=8, extra={}, message_id='4'), classifications=[])], extra={})]
```
The problem was that Pydantic tries to coerce objects to the first class listed in a Union. See [Pydantic docs](https://docs.pydantic.dev/usage/types/#unions) for more details.

This is fixed by adding a discriminator field (`class_name`) which is ignored during serialization.

### DIFF
--- a/labelbox/data/annotation_types/data/dicom.py
+++ b/labelbox/data/annotation_types/data/dicom.py
@@ -1,5 +1,7 @@
+from labelbox.typing_imports import Literal
+from labelbox.utils import _NoCoercionMixin
 from .base_data import BaseData
 
 
-class DicomData(BaseData):
-    ...
+class DicomData(BaseData, _NoCoercionMixin):
+    class_name: Literal["DicomData"] = "DicomData"
--- a/labelbox/data/annotation_types/label.py
+++ b/labelbox/data/annotation_types/label.py
@@ -11,11 +11,14 @@ from .annotation import (ClassificationAnnotation, ObjectAnnotation,
                          VideoClassificationAnnotation, VideoObjectAnnotation,
                          DICOMObjectAnnotation)
 from .classification import ClassificationAnswer
-from .data import DicomData, VideoData, TextData, ImageData
+from .data import AudioData, ConversationData, DicomData, DocumentData, HTMLData, ImageData, MaskData, TextData, VideoData
 from .geometry import Mask
 from .metrics import ScalarMetric, ConfusionMatrixMetric
 from .types import Cuid
 from ..ontology import get_feature_schema_lookup
+
+DataType = Union[VideoData, ImageData, TextData, TiledImageData, AudioData,
+                 ConversationData, DicomData, DocumentData, HTMLData]
 
 
 class Label(BaseModel):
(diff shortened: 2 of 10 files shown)


tests/data/annotation_types/test_label.py
```diff
@@ -181,3 +182,14 @@ def test_schema_assignment_confidence():
                   ])
 
     assert label.annotations[0].confidence == 0.914
+
+
+def test_initialize_label_no_coersion():
```
kopreschko, line 187: small typo here, `coersion` => `coercion`

labelbox/data/annotation_types/data/dicom.py
```diff
@@ -1,5 +1,7 @@
+from labelbox.typing import Literal
+from labelbox.utils import _NoCoercionMixin
 from .base_data import BaseData
 
 
-class DicomData(BaseData):
-    ...
\ No newline at end of file
+class DicomData(BaseData, _NoCoercionMixin):
+    data_type: Literal["DicomData"] = "DicomData"
```
kkim-labelbox, line 7: should this be `class_name`?

labelbox/data/annotation_types/label.py
```diff
@@ -37,7 +40,7 @@ class Label(BaseModel):
         extra: additional context
     """
     uid: Optional[Cuid] = None
-    data: Union[VideoData, ImageData, TextData, TiledImageData]
+    data: DataType
```
vbrodsky, line 43: is this a generic pydantic type? sorry pydantic documentation links do not work for me for some reason

vbrodsky (approved):
Fantastic, we should extend it to all our union types... as you know they are all susceptible

labelbox/data/annotation_types/label.py
whistler: I've defined this type on line 19 above since the list was getting too long.

tests/data/annotation_types/test_label.py
whistler: Good catch! fixed!

labelbox/data/annotation_types/data/dicom.py
whistler: good catch! fixed it.

tests/data/annotation_types/test_label.py
kopreschko: awesome work BTW!

kkim-labelbox (approved):

apollonin (approved):

PR #1013, opened 2023-03-20
[AL-5296] Add global key support to ConversationEntity, DocumentEntity, DicomSe…

kkim-labelbox:
…gments

- Added global key to serialization of ConversationEntity, DocumentEntity and DicomSegments
- Rearranged conversation and pdf tests, and added global_key test cases to the 3 types above


whistler (approved):
LGTM!

PR #1004, opened 2023-03-16
Vb/classification missing confidence al 5239

vbrodsky:
Story: https://labelbox.atlassian.net/browse/AL-5239

This adds the _confidence_ attribute to all annotation values

PR #1015, opened 2023-03-21
[AL-5297] Fix MEA import test

whistler:


vbrodsky (commented):
While I think there is no harm in this PR, I do not think it will address the real issue for annotations to fail - incorrect dimension sizes for image annotations in fixtures. 
I have fixed it on my branch already 
https://github.com/Labelbox/labelbox-python/pull/1004/commits/0786cfd71c6d651b9102662d127f2a3132330dfa in tests/integration/annotation_import/conftest.py

PR #1018, opened 2023-03-21
[AL-0] Make mea tests more reliable

kkim-labelbox:
- Add assertions for label creation in mea tests
- Add check for DR processing

### DIFF
--- a/tests/integration/annotation_import/conftest.py
+++ b/tests/integration/annotation_import/conftest.py
@@ -292,6 +292,32 @@ def ontology():
 
 
 @pytest.fixture
+def wait_for_label_processing():
+    """
+    Do not use. Only for testing.
+
+    Returns project's labels as a list after waiting for them to finish processing.
+    If `project.labels()` is called before label is fully processed,
+    it may return an empty set
+    """
+
+    def func(project):
+        timeout_seconds = 10
+        while True:
+            labels = list(project.labels())
+            if len(labels) > 0:
+                return labels
+            timeout_seconds -= 2
+            if timeout_seconds <= 0:
+                raise TimeoutError(
+                    f"Timed out waiting for label for project '{project.uid}' to finish processing"
+                )
+            time.sleep(2)
+
+    return func
+
+
+@pytest.fixture
 def configured_project(client, ontology, rand_gen, image_url):
     project = client.create_project(name=rand_gen(str),
                                     queue_mode=QueueMode.Dataset)
@@ -303,6 +329,7 @@ def configured_project(client, ontology, rand_gen, image_url):
     data_row_ids = []
     for _ in range(len(ontology['tools']) + len(ontology['classifications'])):
         data_row_ids.append(dataset.create_data_row(row_data=image_url).uid)
+    project._wait_until_data_rows_are_processed(data_row_ids=data_row_ids)
     project.datasets.connect(dataset)
     project.data_row_ids = data_row_ids
     yield project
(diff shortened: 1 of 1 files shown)


tests/integration/annotation_import/conftest.py
```diff
@@ -291,6 +291,32 @@ def ontology():
     return {"tools": tools, "classifications": classifications}
 
 
+@pytest.fixture
+def wait_for_label_processing():
+    """
+    Do not use. Only for testing.
```
vbrodsky, line 297: it is in the testing dir.... isn't it obvious? :)

tests/integration/annotation_import/conftest.py
```diff
@@ -291,6 +291,32 @@ def ontology():
...
+
+    Returns project's labels as a list after waiting for them to finish processing.
+    If `project.labels()` is called before label is fully processed,
+    it may return an empty set
+    """
+
+    def func(project):
+        timeout_seconds = 10
+        while True:
+            labels = list(project.labels())
```
vbrodsky, line 307: the best we can do? no other api, to, say, check for labels by some sort of id? or just a heuristics till we come up with smth better?

tests/integration/annotation_import/conftest.py
```diff
@@ -615,14 +643,21 @@ def model_run_with_training_metadata(rand_gen, model):
...
 def model_run_with_model_run_data_rows(client, configured_project,
-                                       model_run_predictions, model_run):
+                                       model_run_predictions, model_run,
+                                       wait_for_label_processing):
     configured_project.enable_model_assisted_labeling()
 
     upload_task = LabelImport.create_from_objects(
         client, configured_project.uid, f"label-import-{uuid.uuid4()}",
         model_run_predictions)
     upload_task.wait_until_done()
```
vbrodsky, line 653: so this is not working?

tests/integration/annotation_import/conftest.py
```diff
@@ -291,6 +291,32 @@ def ontology():
...
+def wait_for_label_processing():
+    """
+    Do not use. Only for testing.
+
+    Returns project's labels as a list after waiting for them to finish processing.
+    If `project.labels()` is called before label is fully processed,
+    it may return an empty set
+    """
+
+    def func(project):
```
whistler, line 304: We could also use retry decorators for this type of logic which IMO is cleaner. This is how it might look like.
```
from google.api_core import retry

@pytext.fixture
@retry.Retry(deadline=10)
def wait_for_label_processing(project):
     labels = list(project.labels())
     if len(labels) > 0:
         return labels
     else:
         raise ValueError
```

tests/integration/annotation_import/conftest.py
kkim-labelbox: Right, no way to fetch labels by id for a project

tests/integration/annotation_import/conftest.py
kkim-labelbox: The fixture was added to check for DR processing to be finished

tests/integration/annotation_import/conftest.py
kkim-labelbox: Copy pasted code from another fixture haha

whistler (approved):

tests/integration/annotation_import/conftest.py
```diff
@@ -615,14 +643,21 @@ def model_run_with_training_metadata(rand_gen, model):
...
         model_run_predictions)
     upload_task.wait_until_done()
-    label_ids = [label.uid for label in configured_project.labels()]
+    assert upload_task.state == AnnotationImportState.FINISHED, "Label Import did not finish"
+    assert len(
+        upload_task.errors
+    ) == 0, f"Label Import {upload_task.name} failed with errors {upload_task.errors}"
+    labels = wait_for_label_processing(configured_project)
+    label_ids = [label.uid for label in labels]
+    print(label_ids)
```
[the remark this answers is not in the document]
kkim-labelbox: Removed

vbrodsky (approved):
ship it :)

PR #1028, opened 2023-03-23
[AL-5120][AL-5121] Integration tests for Audio and HTML import

whistler:


tests/integration/annotation_import/test_conversation_import.py
```diff
@@ -39,16 +39,4 @@ def test_conversation_entity(client, configured_project_without_data_rows,
         predictions=labels)
     import_annotations.wait_until_done()
 
-    assert import_annotations.errors == []
-
-    exported_labels = configured_project_without_data_rows.label_generator()
```
whistler, line 44: Updated to make it clear that MAL Labels are not part of the exports. This returns zero labels, and the loop below doesn't get run.

vbrodsky (approved):

kkim-labelbox (approved):

PR #1032, opened 2023-03-28
[AL-5110] Fix video raster annotation import

whistler:
* The name field at the root of VideoMaskAnnotation is not needed since this is part of `instances`, therefore it has been removed.
* Added an integration test to ensure VideoMaskAnnotation can be imported.

### DIFF
--- a/tests/integration/annotation_import/test_data_types.py
+++ b/tests/integration/annotation_import/test_data_types.py
@@ -21,12 +21,44 @@ checklist_annotation = lb_types.ClassificationAnnotation(
 text_annotation = lb_types.ClassificationAnnotation(
     name="text", value=lb_types.Text(answer="sample text"))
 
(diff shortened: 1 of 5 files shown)


tests/integration/annotation_import/test_data_types.py
```diff
@@ -153,16 +188,7 @@ def test_import_label_annotations(client, configured_project,
     data_row.delete()
 
 
-@pytest.mark.parametrize('data_type, data_class, annotations', [
```
vbrodsky, line 156: why moved this one?

whistler: The two tests have the same parameters so I extracted them out to a variable.

kkim-labelbox (approved):
LGTM!

PR #1035, opened 2023-03-28
[AL-5113] Add notebook for HTML annotation import

whistler:


examples/annotation_import/html.ipynb
ovalle15, line 1: @whistler do you have time to add nested classification examples? We should support those I think.

ovalle15 (approved):

examples/annotation_import/html.ipynb
whistler: Yes, are there existing nested classification examples that I can use as a template?

ovalle15: Approved, if you do add a nested class let me know and I review again.

PR #1039, opened 2023-03-31
Fix label import test

whistler:
The API now validates that an ontology is linked to a project before LabelImport can be done. The two tests were using a project without ontology for label import, updating to use project with ontology.


mnoszczak:
image

PR #1045, opened 2023-04-07
[AL-5276] Fix ISO timestamp parsing

whistler:
When creating a data row, datetime metadata in the following iso format was not supported: `2011-05-07T14:34:14+00:00`. Since Python 3.7-3.10 did not have full parsing support for ISO timestamps we were doing datetime manipulation in our parsing code, which resulted in this problem. Python 3.11 has more general support for parsing ISO datetime strings, I have installed a backport to bring Python 3.11 datetime parsing to earlier versions. Tests have been adding for adding / updating datetime metadata.

### DIFF
--- a/labelbox/__init__.py
+++ b/labelbox/__init__.py
@@ -1,6 +1,10 @@
 name = "labelbox"
 __version__ = "3.43.0"
 
+from backports.datetime_fromisoformat import MonkeyPatch
+
+MonkeyPatch.patch_fromisoformat()
+
 from labelbox.client import Client
 from labelbox.schema.project import Project
 from labelbox.schema.model import Model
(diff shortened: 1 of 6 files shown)


labelbox/__init__.py
```diff
@@ -1,6 +1,10 @@
 name = "labelbox"
 __version__ = "3.43.0"
 
+from backports.datetime_fromisoformat import MonkeyPatch
+
+MonkeyPatch.patch_fromisoformat()
```
vbrodsky, line 6: is this change hard to implement so a backport is justified? in the words of their creator would possibly break when a feature becomes present -  see here https://pypi.org/project/backports/ - last paragraph titled **What if the feature is present?**

whistler: While supporting the reported case is easy enough to fix, there are several edge cases for parsing datetimes and we might miss some of them with a custom implementation. This is best left to a library. The risk of this breaking in a future version is very low since the backport is based on the CPython code itself.

kopreschko (approved):
lgtm

vbrodsky (approved):

PR #1048, opened 2023-04-11
Fix specs due to models --> experiments change in export schemas

vbrodsky:
Fixed sdk integration tests for export v2 to support https://labelbox.atlassian.net/browse/AL-5450 - change of attribute names for models/model_run to experiments/run


whistler (approved):

vbrodsky:
the spec will fail till I have released this pr https://github.com/Labelbox/python-monorepo/pull/1091

PR #1053, opened 2023-04-17
[AL-5519] Add SDK Validation

mnoszczak:


labelbox/schema/export_params.py
```diff
@@ -34,3 +36,18 @@ class CatalogExportParams(DataRowParams):
 class ModelRunExportParams(DataRowParams):
     # TODO: Add model run fields
     pass
+
+
+def validate_array_length(array, max_length, array_name):
```
kkim-labelbox, line 41: Probably good to prefix this with `_` since it's a private function
`_validate_array_length`

labelbox/schema/export_params.py
```diff
@@ -34,3 +36,18 @@ class CatalogExportParams(DataRowParams):
...
+    if len(array) > max_length:
+        raise ValueError(f"{array_name} cannot exceed {max_length} items")
+
+
+def validate_catalog_export_params(params: CatalogExportParams):
+    if "model_run_ids" in params and params["model_run_ids"] is not None:
```
kkim-labelbox, line 47: `params["model_run_ids"] is not None` can just be `params["model_run_ids"]`, same with the line below

mnoszczak: 🤔 that's how I started and it was causing issues, I'll revisit to see if that's expected

labelbox/schema/export_params.py
```diff
@@ -34,3 +36,18 @@ class CatalogExportParams(DataRowParams):
...
+        validate_array_length(params["model_run_ids"], EXPORT_LIMIT,
```
mnoszczak, line 48:
```suggestion
        _validate_array_length(params["model_run_ids"], EXPORT_LIMIT,
```

labelbox/schema/export_params.py
```diff
@@ -34,3 +36,18 @@ class CatalogExportParams(DataRowParams):
...
+                              "model_run_ids")
+
+    if "project_ids" in params and params["project_ids"] is not None:
+        validate_array_length(params["project_ids"], EXPORT_LIMIT,
```
mnoszczak, line 52:
```suggestion
        _validate_array_length(params["project_ids"], EXPORT_LIMIT,
```

labelbox/schema/export_params.py
mnoszczak, line 41:
```suggestion
def _validate_array_length(array, max_length, array_name):
```

PR #1058, opened 2023-04-20
[AL-5533][AL-5538][AL-5539]Vb/conversational v2 export al 5538

vbrodsky:
Stories:
- https://labelbox.atlassian.net/browse/AL-5553
- https://labelbox.atlassian.net/browse/AL-5538
- https://labelbox.atlassian.net/browse/AL-5539

Essentially updating existing fixtures to fix export v2 tests for Conversational, Document and DICOM data

### DIFF
--- a/tests/integration/annotation_import/conftest.py
+++ b/tests/integration/annotation_import/conftest.py
@@ -37,8 +37,6 @@ def conversation_data_row(rand_gen):
             "https://storage.googleapis.com/labelbox-developer-testing-assets/conversational_text/1000-conversations/conversation-1.json",
         "global_key":
             f"https://storage.googleapis.com/labelbox-developer-testing-assets/conversational_text/1000-conversations/conversation-1.json-{rand_gen(str)}",
-        "media_type":
-            "CONVERSATIONAL",
     }
 
 
--- a/tests/integration/annotation_import/conftest.py
+++ b/tests/integration/annotation_import/conftest.py
@@ -34,9 +34,9 @@ def audio_data_row(rand_gen):
 def conversation_data_row(rand_gen):
     return {
         "row_data":
-            "https://storage.googleapis.com/labelbox-datasets/conversational-sample-data/sample-conversation-1.json",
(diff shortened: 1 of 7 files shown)


tests/integration/annotation_import/conftest.py
```diff
@@ -27,11 +28,9 @@ def audio_data_row(rand_gen):
 def conversation_data_row(rand_gen):
     return {
         "row_data":
```
vbrodsky, line 30: using the same data as our notebooks.... I thought in the beginning this was one of the reasons for failing conversational export tests

tests/integration/annotation_import/conftest.py
```diff
@@ -173,6 +178,34 @@ def annotations_by_data_type(polygon_inference, rectangle_inference,
     }
 
 
+@pytest.fixture
+def annotations_by_data_type_v2(
```
vbrodsky, line 182: Separated from export v1 annotation mappings, cleaner also it was breaking exports v1

tests/integration/annotation_import/conftest.py
```diff
@@ -519,13 +623,52 @@ def point_inference(prediction_id_mapping):
 
 
 @pytest.fixture
-def entity_inference(prediction_id_mapping):
+def entity_inference(request, prediction_id_mapping):
```
kkim-labelbox, line 626: What is `request`?

tests/integration/annotation_import/conftest.py
```diff
@@ -1,3 +1,4 @@
+import glob
```
kkim-labelbox, line 1: Is this an unused import?

tests/integration/annotation_import/conftest.py
vbrodsky: `request` represents the calling function context
I was experimenting with changing fixture data based on whether the caller is export_v1 or _v2... however even if I could get it to work, it would make our code just more complex... so just ended up writing more fixtures
will remove `request`

tests/integration/annotation_import/conftest.py
vbrodsky: yep

kkim-labelbox (approved):

PR #1060, opened 2023-04-21
[AL-5332] allow user to update datarow metadata by global keys

yamini-labelbox:
https://labelbox.atlassian.net/browse/AL-5332

This is to add the feature to update datarowmetada using a global key.

Todo: 
fix integration tests. Going to update the PR after I fix those.


kkim-labelbox (commented):
LGTM. A few things to keep in mind for SDK PRs:
- Set up pre-commit tool, which will use `yapf` to automatically catch any formatting issues upon git commits. Instructions are inside `CONTRIB.md` file
- It'd be good to add some integration tests that call `bulk_upsert` using global_keys inside `test_data_row_metadata.py` file
- However, keep in mind that above test will fail before intelligence changes are merged and deployed to staging, because tests in the PR run against `staging` environment

labelbox/schema/data_row_metadata.py
```diff
@@ -480,6 +483,7 @@ def bulk_upsert(
 
 
         >>> metadata = DataRowMetadata(
+        >>>                 global_key="global-key",
```
kkim-labelbox, line 486: Can we specify in the docstring here that users should supply either global_key or data_row, and show usage of only one of them in the example? I feel like it's confusing to show us setting both in the example
```
        """Upsert metadata to a list of data rows.

        You may specify data row by either global_key or data_row_id

        >>> metadata = DataRowMetadata(
        >>>                 global_key="global-key", # Alternatively, set data_row_id="data-row-id"
        >>>                 fields=[
        >>>                        DataRowMetadataField(schema_id="schema-id", value="my-message"),
        >>>                        ...
        >>>                    ]
        >>>    )
        >>> mdo.batch_upsert([metadata])
```

kkim-labelbox (approved):

PR #1076, opened 2023-04-28
[AL-5458] Add our own library to use Python json parser to parse ndjson

vbrodsky:
Story: https://labelbox.atlassian.net/browse/AL-5458

This is the first step in replacing ndjson library with our own. This PR adds a new module that has exactly the same methods as we use in ndjson library BUT we implement them ourselves using Python json parser. 

The is the same as ndjson uses, so the end result is very similar in terms of parsing and performance. I have added tests, where I generated expected results using the ndjson library


gszpak (approved):
LGTM!

apollonin (approved):

PR #1077, opened 2023-04-30
[AL-5264] Add interpolated frames

mnoszczak:
add interpolated frames support


vbrodsky (commented):
It's all cool, but we should modify an existing integration spec, I think. to prove stuff is/will be working for longevity. 
For example, there is a test tests/integration/annotation_import/fixtures/video_annotations.py (there maybe others etc.)

kkim-labelbox (approved):

PR #1107, opened 2023-05-24
improved metadata nb

ezekielemerson:


ovalle15 (approved):
I confirmed that the bulk_export  only accepts data row ids

PR #1108, opened 2023-05-24
[AL-5857]Add ISO format to exports V2 date filters

yamini-labelbox:
https://labelbox.atlassian.net/browse/AL-5857


labelbox/schema/export_filters.py
```diff
@@ -45,19 +45,33 @@ class DatasetExportFilters(SharedExportFilters):
...
             except ValueError:
                 pass
         raise ValueError(f"""Incorrect format for: {string_date}.
-        Format must be \"YYYY-MM-DD\" or \"YYYY-MM-DD hh:mm:ss\"""")
+        Format must be \"YYYY-MM-DD\" or \"YYYY-MM-DD hh:mm:ss\" or ISO 8061 format \"YYYY-MM-DDThh:mm:ss±hhmm\""""
+                        )
     return True
 
 
+def covert_to_utc_if_iso8061(string_date: str, timezone_str: Optional[str]):
```
kkim-labelbox, line 63: `covert_to_utc_if_iso8061` --> `convert_to_utc_if_iso8061`
can we also rename `string_date` everywhere to `datetime_str` to stay consistent with `timezone_str`?

labelbox/schema/export_filters.py
```diff
@@ -45,19 +45,33 @@ class DatasetExportFilters(SharedExportFilters):
...
+    """helper function to convert datetime to UTC if it is in ISO_8061_FORMAT and set timezone to UTC"""
+    try:
+        date_obj = datetime.strptime(string_date, ISO_8061_FORMAT)
+        date_obj_utc = date_obj.replace(tzinfo=timezone.utc)
+        string_date = date_obj_utc.strftime(ISO_8061_FORMAT)
+        timezone_str = "UTC"
+    except ValueError:
```
kkim-labelbox, line 70: For my own understanding, if the date is not in ISO8601 format, will it fail at the first line, `date_obj = datetime.strptime(string_date, ISO_8061_FORMAT)`, and therefore `timezone_str` will not be set to `UTC`?

kkim-labelbox:
Also would be great to add a test case using ISO8601 datetime!

labelbox/schema/export_filters.py
yamini-labelbox: yes, that is correct. It will throw a ValueError which we are catching.

kkim-labelbox (approved):
LGTM!

PR #1116, opened 2023-05-25
[AL-5662] Add batch export v2

mnoszczak:
```
...
batches = project.batches()
batch = #find_interesting_batch
task = project.export_v2(filters={ "batch_id": batch.id  })
...
```


tests/integration/test_batch.py
```diff
@@ -1,6 +1,10 @@
 from labelbox.exceptions import ProcessingWaitTimeout
 import pytest
 from labelbox import Dataset, Project
+from typing import Tuple
```
kkim-labelbox, line 4: Unnecesary imports?

tests/integration/test_project.py
```diff
@@ -42,6 +45,30 @@ def test_project(client, rand_gen):
     assert project not in projects
 
 
+def test_batch_project_export_v2(
```
kkim-labelbox, line 48: Could you add multiple batches to this project, so we can test the filters properly? Also would be good to verify the data row ids as well

tests/integration/test_project.py
```diff
@@ -42,6 +45,43 @@ def test_project(client, rand_gen):
...
+        configured_batch_project_with_label: Tuple[Project, Dataset, DataRow,
```
kkim-labelbox, line 49: Nit: We don't need to define all the return types since this is a test. May be cleaner to not have to import all of the types too

PR #1126, opened 2023-06-06
Vb/freetext confidence al 6008

vbrodsky:
Story: https://labelbox.atlassian.net/browse/AL-6008

We are able to support confidence on both free text and nested free text by just moving the `confidence` attribute to `Text` when we convert date to_ from_common. The attribute itself has already been defined via pydantic

Also added unit tests, integration test and test via a notebook

### DIFF
--- a/tests/data/serialization/ndjson/test_text.py
+++ b/tests/data/serialization/ndjson/test_text.py
@@ -1,5 +1,5 @@
 from labelbox.data.annotation_types.annotation import ClassificationAnnotation
-from labelbox.data.annotation_types.classification.classification import Checklist, ClassificationAnswer, Radio, Text
+from labelbox.data.annotation_types.classification.classification import ClassificationAnswer, Radio, Text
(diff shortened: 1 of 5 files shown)


tests/data/serialization/ndjson/test_text.py
```diff
@@ -13,40 +13,92 @@ def test_serialization():
                       text="This is a test",
                   ),
                   annotations=[
-                      ClassificationAnnotation(
-                          name="radio_question_geo",
-                          confidence=0.5,
-                          value=Text(answer="first_radio_answer"))
+                      ClassificationAnnotation(name="free_text_annotation",
```
kkim-labelbox, line 16: Is this the best place to put this test? `test_text.py` is testing for Text DataRow cases, so it doesn't seem like a good place to put tests on text _classification_.

kkim-labelbox (approved):

PR #1147, opened 2023-06-22
Reduce wait time for DR processing in tests

kkim-labelbox:
- Reduced timeout for `project._wait_until_data_rows_are_processed` to 30s with sleep interval set to 5s
- Reduced number of retries on github action from 5-->3, with reruns-delay from 10s-->3s
- Removed @attila-papai 's` _wait_until_data_rows_are_processed` calls, since it's taken care of in `create_batch`

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -89,4 +89,4 @@ jobs:
 
           DA_GCP_LABELBOX_API_KEY: ${{ secrets[matrix.da-test-key] }}
         run: |
-          tox -e py -- -n 10 -svv --reruns 5 --reruns-delay 10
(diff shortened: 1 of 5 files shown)


yamini-labelbox (approved):

kkim-labelbox:
> > Removed @attila-papai 's _wait_until_data_rows_are_processed calls, since it's taken care of in create_batch
> 
> @kkim-labelbox this was explicitly introduced to fix integration tests yesterday, can this change cause a test regression?

@mnoszczak , this function is already embedded into create_batch(), so the tests should pass without it. The issue with integration tests yesterday was due to how ES count is fetched to determine which data rows are stuck in processing. That's been fixed in intelligence, so we should be good here

PR #1148, opened 2023-06-22
[AL-6339] Fix numpy version

mnoszczak:


MeTaNoV:
Need probably to be able to run the coordinator properly...

```
ImportError: cannot import name '_GenericAlias' from 'numpy._typing' (/usr/local/lib/python3.9/site-packages/numpy/_typing/__init__.py)
```

MeTaNoV:
After fixing it in the "model-training" repo, the training itself have the same issue...

image

PR #1151, opened 2023-06-26
Exclude feature_schema_id from exporter v2 expected results

vbrodsky:
Removing feature_schema_id from exporter v2 tests

1. This is a new field, BUT it is not a 'static' field thus can not assert the value
2. However in the near future we should add an assertion on presence of the key


vbrodsky:
> THANK YOU so much for looking into this, val!

of course... I am curious why this did not fail on merge

PR #1166, opened 2023-07-05
[AL-6476] Integration tests: convert all project queue_mode from DATASET to BATCH

vbrodsky:
Batch is now considered a malformed query, see https://github.com/Labelbox/intelligence/blob/develop/apps/api/src/projects/modules/project-resolvers/project-settings.ts#L66-L68

So we have updated all our fixture projects to be `queue_mode=QueueMode.Batch`

In the process of converting our test projects to batch, we ended up making the following changes:
- Updating `validate_labeling_parameter_overrides` on the Project, we do not allow overwriting num_of_labels for batch projects
- Removed any tests for creating reviews, that is not supported for batch projects I am told
- Adding waits on data row / project data row creations
  - Note our mutation to check on project data row completions only makes sure the data rows completed processing, it does not consider success / failure 
- Also added media types to all projects 
- Fixing various other test glitches:
  - Moving test artifacts to our own bucket to avoid ES CORS errors
  - Fixing document annotation type data


tests/integration/conftest.py
```diff
@@ -204,25 +204,30 @@ def pdf_entity_data_row(client):
...
     project = client.create_project(name=rand_gen(str),
-                                    queue_mode=QueueMode.Dataset)
+                                    queue_mode=QueueMode.Batch,
+                                    media_type=MediaType.Image)
     yield project
     project.delete()
 
 
 @pytest.fixture
 def batch_project(client, rand_gen):
```
attila-papai, line 227: this should be removed now and all references should be replaced to use `project` (which is the same fixture now)

tests/integration/test_sorting.py
```diff
@@ -9,13 +9,13 @@
 def test_relationship_sorting(client):
     a = client.create_project(name="a",
```
attila-papai, line 10: here we still need to add the media_type param

tests/integration/test_review.py
```diff
@@ -1,50 +0,0 @@
-import pytest
```
attila-papai, line 1: why was this file deleted?

tests/integration/test_relationships.py
```diff
@@ -1,46 +0,0 @@
-import pytest
```
attila-papai, line 1: why was this file deleted?

tests/integration/test_project.py
```diff
@@ -342,7 +342,7 @@ def test_queued_data_row_export(configured_project):
 
 def test_queue_mode(configured_project: Project):
     # ensures default queue mode is dataset
```
attila-papai, line 344: wrong comment

tests/integration/test_project.py
```diff
@@ -19,7 +19,7 @@ def test_project(client, rand_gen):
     data = {
         "name": rand_gen(str),
         "description": rand_gen(str),
-        "queue_mode": QueueMode.Dataset
+        "queue_mode": QueueMode.Batch
```
attila-papai, line 22: add `media_type` param

tests/integration/test_batch.py
```diff
@@ -12,7 +12,7 @@ def big_dataset(dataset: Dataset):
             "row_data": IMAGE_URL,
             "external_id": "my-image"
         },
-    ] * 250)
+    ] * 2)
```
attila-papai, line 15: this is a big dataset for a reason, please revert this back to 250

tests/integration/conftest.py
```diff
@@ -499,11 +534,26 @@ def _setup_ontology(project):
...
+                                    queue_mode=QueueMode.Batch,
+                                    media_type=MediaType.Image)
+    dataset = initial_dataset
     data_row = dataset.create_data_row(row_data=image_url)
+    data_row_ids = [data_row.uid]
+
+    project._wait_until_data_rows_are_processed(
+        data_row_ids=data_row_ids,
+        wait_processing_max_seconds=DATA_ROW_PROCESSING_WAIT_TIMEOUT_SECONDS,
+        sleep_interval=DATA_ROW_PROCESSING_WAIT_SLEEP_INTERNAL_SECONDS)
```
attila-papai, line 549: the below `create_batch` function calls `_wait_until_data_rows_are_processed` internally, so no need for this here

tests/integration/conftest.py
```diff
@@ -356,17 +366,37 @@ def project_based_user(client, rand_gen):
...
+
+@pytest.fixture
+def configured_project(project, initial_dataset, client, rand_gen, image_url):
+    dataset = initial_dataset
+    data_row_id = dataset.create_data_row(row_data=image_url).uid
+
+    project._wait_until_data_rows_are_processed(
+        data_row_ids=[data_row_id],
+        wait_processing_max_seconds=DATA_ROW_PROCESSING_WAIT_TIMEOUT_SECONDS,
+        sleep_interval=DATA_ROW_PROCESSING_WAIT_SLEEP_INTERNAL_SECONDS)
```
attila-papai, line 392: the below `create_batch` function calls `_wait_until_data_rows_are_processed` internally, so no need for this here

tests/integration/conftest.py
```diff
@@ -356,17 +366,37 @@ def project_based_user(client, rand_gen):
...
         proj.delete()
 
 
 @pytest.fixture
-def configured_project(project, client, rand_gen, image_url):
-    dataset = client.create_dataset(name=rand_gen(str), projects=project)
-    dataset.create_data_row(row_data=image_url)
+def initial_dataset(client, rand_gen):
+    dataset = client.create_dataset(name=rand_gen(str))
+    yield dataset
```
attila-papai, line 381: missing `dataset.delete()` after `yield`

tests/integration/annotation_import/test_data_types.py
```diff
@@ -115,21 +117,43 @@ def get_annotation_comparison_dicts_from_export(export_result, data_row_id,
     return converted_annotations
 
 
+def create_data_row_for_project(project, dataset, data_row_ndjson, batch_name):
+    data_row = dataset.create_data_row(data_row_ndjson)
+
+    project._wait_until_data_rows_are_processed(
+        data_row_ids=[data_row.uid],
+        wait_processing_max_seconds=DATA_ROW_PROCESSING_WAIT_TIMEOUT_SECONDS,
+        sleep_interval=DATA_ROW_PROCESSING_WAIT_SLEEP_INTERNAL_SECONDS)
```
attila-papai, line 126: the below `create_batch` function calls `_wait_until_data_rows_are_processed` internally, so no need for this here

tests/integration/annotation_import/conftest.py
```diff
@@ -493,29 +502,37 @@ def configured_project(client, ontology, rand_gen, image_url):
...
             where=LabelingFrontend.name == "editor"))[0]
     project.setup(editor, ontology)
-    data_row_ids = []
-    data_row_ids.append(dataset.create_data_row(pdf_url).uid)
+    data_row = dataset.create_data_row(pdf_url)
+    data_row_ids = [data_row.uid]
     project._wait_until_data_rows_are_processed(
         data_row_ids=data_row_ids,
         wait_processing_max_seconds=DATA_ROW_PROCESSING_WAIT_TIMEOUT_SECONDS,
         sleep_interval=DATA_ROW_PROCESSING_WAIT_SLEEP_INTERNAL_SECONDS)
```
attila-papai, line 530: the below create_batch function calls _wait_until_data_rows_are_processed internally, so no need for this here

tests/integration/annotation_import/conftest.py
```diff
@@ -493,29 +502,37 @@ def configured_project(client, ontology, rand_gen, image_url):
...
```
attila-papai, line 504: the below `create_batch` function calls `_wait_until_data_rows_are_processed` internally, so no need for this here

tests/integration/annotation_import/conftest.py
```diff
@@ -477,10 +478,18 @@ def func(project):
 
 
 @pytest.fixture
-def configured_project(client, ontology, rand_gen, image_url):
-    project = client.create_project(name=rand_gen(str),
-                                    queue_mode=QueueMode.Dataset)
+def initial_dataset(client, rand_gen):
     dataset = client.create_dataset(name=rand_gen(str))
+    yield dataset
```
attila-papai, line 483: missing `dataset.delete()` after `yield`

vbrodsky: thx

tests/integration/test_batch.py
vbrodsky: I think the reason is to try to do some sort of load testing which I do not think we should do in sdk functionality test, I am discussing further with Richard

tests/integration/test_review.py
vbrodsky: this was discussed with Richard here https://labelbox.slack.com/archives/C02F8TKT7RU/p1689174866073139

tests/integration/test_relationships.py
vbrodsky: this test was designed specifically to test `Relationship` between project and dataset but this does not exist for Batch projects... I think equivalent would be to use projects.datasets.connect(dataset) and this is tested indirectly in tests/integration/test_sorting.py.. but actually let me ask @kkim-labelbox to verify this test can be remove

attila-papai (approved):

PR #1171, opened 2023-07-12
[AL-6294] Support cursor-based pagination for dataset.data_rows()

vbrodsky:
Story: https://labelbox.atlassian.net/browse/AL-6294

Replaces dataset.data_rows() Relation with a method and implement cursor-based pagination

- Why did I not update the existing data_rows() Relationship but created a new method instead?
  - There are specific requirements for how graphql needs to be composed for consumption in Relationship, so I would have to create a Dataset#data_rows_2() node resolver, which I think would be confusing
  - The Relationship framework does not support cursor-based pagination and would have to be extended
    - BUT we are talking of getting rid of this home-grown framework for sdk 4
  - All our queries for cursor-based access use custom methods instead of Relation

And this the work I needed to do:
- make sure my method supports `where` parameter the way Relationship does. I have extracted `where` and order by method into simple helpers in the query module
- extended cursor pagination to allow customize page size

Some of perf test results (I will publish complete results in the story once I finish my tests):

Running on my laptop dataset.data_rows()  against stage / mysql org

Offset pagination (legacy)
page size 100
dataset size: 1 mln
time per query p50/p99 in DataDog: 521/ 1000 ms
db utilization: normal (under 1%)
completion time: ??? still running

Cursor pagination (this PR)
page size 500
dataset size: 1 mln
time per query p50/p99 in DataDog: 16/ 26 ms
db utilization: normal (under 1%)
completion time: 1 hour 2 mins 46 s

### DIFF
--- a/labelbox/schema/dataset.py
+++ b/labelbox/schema/dataset.py
@@ -74,7 +74,7 @@ class Dataset(DbObject, Updateable, Deletable):
         Custom method to paginate data_rows via cursor.
 
         Params:
-            after (str): Cursor (data row id) to start from, if none, will start from the beginning
+            from_cursor (str): Cursor (data row id) to start from, if none, will start from the beginning
             where (dict(str,str)): Filter to apply to data rows. Where value is a data row column name and key is the value to filter on.    
                 example: {'external_id': 'my_external_id'} to get a data row with external_id = 'my_external_id'
 
@@ -87,38 +87,27 @@ class Dataset(DbObject, Updateable, Deletable):
         """
 
         page_size = 500  # hardcode to avoid overloading the server
-        datarow_selections = query.results_query_part(Entity.DataRow)
-
-        empty_string = ''
-
-        where_clause = empty_string
-        where_vars = empty_string
-        where_param = None
-        if where is not None:
-            where_clause = ', $where: DatasetDataRowWhereInput'
-            where_vars = ', where: $where'
-            where_param = query.where_as_dict(Entity.DataRow, where)
+        where_param = query.where_as_dict(Entity.DataRow,
+                                          where) if where is not None else None
 
         template = Template(
-            """query DatasetDataRowsPyApi($$id: ID!, $$from: ID, $$first: Int$where_clause)  {
-                        datasetDataRows(id: $$id, from: $$from, first: $$first$where_vars)
+            """query DatasetDataRowsPyApi($$id: ID!, $$from: ID, $$first: Int, $$where: DatasetDataRowWhereInput)  {
+                        datasetDataRows(id: $$id, from: $$from, first: $$first, where: $$where)
                             {
                                 nodes { $datarow_selections }
                                 pageInfo { hasNextPage startCursor }
                             }
                         }
                     """)
-        query_str = template.substitute(where_clause=where_clause,
-                                        where_vars=where_vars,
-                                        datarow_selections=datarow_selections)
+        query_str = template.substitute(
+            datarow_selections=query.results_query_part(Entity.DataRow))
 
         params = {
             'id': self.uid,
             'from': from_cursor,
             'first': page_size,
+            'where': where_param,
         }
-        if where is not None:
-            params['where'] = where_param
 
         return PaginatedCollection(
             client=self.client,
--- a/labelbox/schema/dataset.py
+++ b/labelbox/schema/dataset.py
@@ -1,4 +1,4 @@
-from typing import Collection, Dict, Generator, List, Optional, Union, Any, TYPE_CHECKING
+from typing import Dict, Generator, List, Optional, Union, Any
 import os
 import json
 import logging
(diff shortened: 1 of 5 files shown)


tests/integration/test_data_rows.py
```diff
@@ -576,12 +576,6 @@ def test_data_row_filtering_sorting(dataset, image_url):
     row2 = row2[0]
     assert row2.external_id == "row2"
 
-    # Test sorting
```
vbrodsky, line 579: order_by not used, removed

labelbox/schema/dataset.py
```diff
@@ -58,12 +60,75 @@ class Dataset(DbObject, Updateable, Deletable):
...
+    def data_rows(
+        self,
+        from_cursor: str = None,
+        where: Comparison = None,
+    ) -> PaginatedCollection:
+        """ 
+        Custom method to paginate data_rows via cursor.
+
+        Params:
+            after (str): Cursor (data row id) to start from, if none, will start from the beginning
```
attila-papai, line 77: typo "after" -> "from_cursor"

labelbox/schema/dataset.py
```diff
@@ -58,12 +60,75 @@ class Dataset(DbObject, Updateable, Deletable):
...
+            where (dict(str,str)): Filter to apply to data rows. Where value is a data row column name and key is the value to filter on.    
+                example: {'external_id': 'my_external_id'} to get a data row with external_id = 'my_external_id'
+
+
+        NOTE: 
+            Order of retrieval is newest data row first.
+            Deleted data rows are not retrieved.
+            Failed data rows are not retrieved.
```
attila-papai, line 85: We are returning both failed and processing data rows too.
Code: https://github.com/Labelbox/intelligence/blob/a688618adaccc8eec391ea479535d94fd15944ce/apps/api/src/catalog/adv/data-rows/data-rows-router.service.ts#L159 (the comment was not removed after all)

labelbox/schema/dataset.py
```diff
@@ -58,12 +60,75 @@ class Dataset(DbObject, Updateable, Deletable):
...
+            Data rows in progress *maybe* retrieved.
+        """
+
+        page_size = 500  # hardcode to avoid overloading the server
```
attila-papai, line 89: should we rather define this in the backend? the users can overwrite this anytime to cause overload

labelbox/schema/dataset.py
```diff
@@ -58,12 +60,75 @@ class Dataset(DbObject, Updateable, Deletable):
...
+
+        template = Template(
+            """query DatasetDataRowsPyApi($$id: ID!, $$from: ID, $$first: Int$where_clause)  {
+                        datasetDataRows(id: $$id, from: $$from, first: $$first$where_vars)
+                            {
+                                nodes { $datarow_selections }
+                                pageInfo { hasNextPage startCursor }
+                            }
+                        }
+                    """)
```
attila-papai, line 110: wouldn't this code read better if you would define all the arguments of `datasetDataRows` and pass `null` to those that are not specified? Instead of concatenating strings based on conditions.

labelbox/schema/dataset.py
```diff
@@ -465,11 +530,11 @@ def data_rows_for_external_id(self,
 
         data_rows = self.data_rows(where=where)
         # Get at most `limit` data_rows.
-        data_rows = list(islice(data_rows, limit))
+        at_most_data_rows = list(islice(data_rows, limit))
```
attila-papai, line 533: can you elaborate why you renamed this?

labelbox/schema/dataset.py
vbrodsky: oh ok that description is for mysql.. I did not want to call out for ADV vs mysql specifically here... @kkim-labelbox pls advice which way I should describe this

labelbox/schema/dataset.py
vbrodsky: I see that... the problem was that currently 100 is hardcoded by our paginator for all other cases... let me look into it

labelbox/schema/dataset.py
vbrodsky: but I have looked at it again and the thing is, we have a const `_PAGE_SIZE` explicitly setting page size in `PaginatedCollection`. So, I guess I could add some param to '`not_set_page_size` and use it for this particular case... I know though we have a separate story to review pagination so perhaps I do it there?

labelbox/schema/dataset.py
vbrodsky: but I have looked at it again and the thing is, we have a const `_PAGE_SIZE` explicitly setting page size in `PaginatedCollection`. So, I guess I could add some param to '`not_set_page_size` and use it for this particular case... I know though we have a separate story to review pagination limits so perhaps I do it there?

kkim-labelbox: Since PAGE_SIZE is already in the SDK, let's not remove it for now, since it'd be a breaking change

attila-papai (approved):
well done!

PR #1173, opened 2023-07-13
Wait for data row media info be populated to fix a flaky test

vbrodsky:
So we have this test `tests/integration/annotation_import/test_conversation_import.py::test_conversation_entity` flaky/almost always failing

I have added our std wait to populate media attributes code and it worked to unflake it.


vbrodsky:
merged in https://github.com/Labelbox/labelbox-python/pull/1166

PR #1174, opened 2023-07-24
[AL-6155] [AL-6724] [AL-6738] Fix some test cases for ADV enabled accounts

attila-papai:
Changes included in this PR:
- marked the "TEXT" attachment type as deprecated, used "RAW_TEXT" instead
- added "is_adv_enabled" fixture to use in some test cases
- added "created_data_rows" property to "Task"


tests/integration/media/bulk_conversation.json
```diff
@@ -1,113 +1,130 @@
...
-                    "userId": "Bot 002",
-                    "name": "Bot"
-                },
-                "align": "left",
-                "canLabel": false
-            }
-        ]
-    }
-]
+[{
```
attila-papai, line 1: this file was reformatted with `yapf`
the only meaningful changes are replacing "TEXT" with "RAW_TEXT"

labelbox/schema/task.py
```diff
@@ -129,6 +129,16 @@ def failed_data_rows(self) -> Optional[Dict[str, Any]]:
         else:
             return None
 
+    @property
+    def created_data_rows(self) -> Optional[Dict[str, Any]]:
```
vbrodsky, line 133: Just for my information, this seems like a general `Task` class yet here we have methods specific to data row tasks like `failed_data_rows` and `created_data_rows` - a consequence of current task implementation?

tests/integration/conftest.py
```diff
@@ -739,3 +739,10 @@ def run_dataset_export_v2_task(cls,
 @pytest.fixture
 def export_v2_test_helpers() -> Type[ExportV2Helpers]:
     return ExportV2Helpers()
+
+
+@pytest.fixture
+def is_adv_enabled(client) -> bool:
```
vbrodsky, line 745: can we make it session scope? PS Not sure whether it is possible, but would save on extra http traffic

labelbox/schema/task.py
attila-papai: probably yes, each async task is represented by this Task class I guess, and some have these properties

tests/integration/conftest.py
attila-papai: good idea, added!

vbrodsky (approved):

PR #1177, opened 2023-07-27
[E2E-752 ] Added negative path tests for test_batch.py

mihhail-m:
TLDR; A little bit updated  `test_batch.py`. 

- Added negative path tests for batch creation.
- Fixed imports order.
- General clean up/refactoring.

### DIFF
--- a/tests/integration/test_batch.py
+++ b/tests/integration/test_batch.py
@@ -13,7 +13,7 @@ def big_dataset(dataset: Dataset):
             "row_data": IMAGE_URL,
             "external_id": EXTERNAL_ID
         },
-    ] * 2)
+    ] * 3)
     task.wait_till_done()
 
     yield dataset
@@ -26,7 +26,7 @@ def small_dataset(dataset: Dataset):
             "row_data": IMAGE_URL,
             "external_id": EXTERNAL_ID
         },
-    ] * 3)
+    ] * 2)
     task.wait_till_done()
 
     yield dataset
@@ -124,7 +124,7 @@ def test_batch_creation_for_data_rows_with_issues(
     ]
     data_rows_to_add = valid_data_rows + invalid_data_rows
 
-    assert len(data_rows_to_add) == 5
+    assert len(data_rows_to_add) == 4
     batch = project.create_batch("batch to test failed data rows",
                                  data_rows_to_add)
     failed_data_row_ids = [x for x in batch.failed_data_row_ids]
--- a/tests/integration/test_batch.py
+++ b/tests/integration/test_batch.py
@@ -1,10 +1,15 @@
 import pytest
 
+from uuid import uuid4
 from labelbox import Dataset, Project
 from labelbox.exceptions import ProcessingWaitTimeout
 
 IMAGE_URL = "https://storage.googleapis.com/diagnostics-demo-data/coco/COCO_train2014_000000000034.jpg"
 EXTERNAL_ID = "my-image"
+
+
+def get_data_row_ids(ds: Dataset):
+    return [dr.uid for dr in list(ds.export_data_rows())]
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


mihhail-m:
@Labelbox/data-io hey, if anyone could give me a review here that would be great.

tests/integration/test_batch.py
```diff
@@ -1,20 +1,26 @@
...
 @pytest.fixture
 def big_dataset(dataset: Dataset):
     task = dataset.create_data_rows([
         {
             "row_data": IMAGE_URL,
-            "external_id": "my-image"
+            "external_id": EXTERNAL_ID
         },
-    ] * 2)
+    ] * 3)
```
attila-papai, line 23: @vbrodsky so this was 250 before (hence big_dataset). I remember you [said](https://labelbox.slack.com/archives/C05J7G8PE4V/p1690224623159979) 250 was giving time out errors, but still, shouldn't we at least raise it to 50? Otherwise this fixture is useless.

attila-papai (approved):
Have you run these tests with both ADV enabled and disabled?

mihhail-m:
@attila-papai yes, I run them against ADV org also, but since the same tests are currently also failing in the SDK pipeline for the ADV it is hard to tell whether my changes are impacted by that. They are passing for non-ADV orgs though.

attila-papai:
> @attila-papai yes, I run them against ADV org also, but since the same tests are currently also failing in the SDK pipeline for the ADV it is hard to tell whether my changes are impacted by that. They are passing for non-ADV orgs though.

okay, our team will take care of the ADV issues

tests/integration/test_batch.py
vbrodsky: @attila-papai let me try to find a slack thread on it... in summary 250+ seems to be a magic number for Richard to catch some bugs so we need to sort it out... @kkim-labelbox perhaps we can add this as a story to me print? (and pull another sdk refactoring)

vbrodsky: 2 or 3 makes no difference for large batch imho

vbrodsky: also see here followup discussion on 250 big dataset rows with Richard if you are so curious https://labelbox.slack.com/archives/C03F33JDRGD/p1690217836296689

mihhail-m: so, what should I change this number to? 50 or 250 ?

mihhail-m: @vbrodsky @attila-papai so, what should be the value change to?

vbrodsky: I recommend keep it to 2. I am working on a story to optimize sdk tests and have in mind of it to either bring back 250 or figure what to do to make sure we test this feature and Richard is happy :)

tests/integration/test_batch.py
```diff
@@ -1,20 +1,26 @@
...
+from uuid import uuid4
 from labelbox import Dataset, Project
+from labelbox.exceptions import ProcessingWaitTimeout, MalformedQueryException, ResourceConflict, LabelboxError
 
 IMAGE_URL = "https://storage.googleapis.com/diagnostics-demo-data/coco/COCO_train2014_000000000034.jpg"
+EXTERNAL_ID = "my-image"
+
+
+def get_data_row_ids(ds: Dataset):
+    return [dr.uid for dr in list(ds.export_data_rows())]
```
yamini-labelbox, line 13: export_data_rows() is legacy exports. Could you use exports_v2 here?

vbrodsky: well we are still need to support exports v1 .. it is not retired yet

mihhail-m: i think it is good to cover legacy functions as long as they are being part of the SDK.

mihhail-m:
Since I don't see any new suggestions for changes and PR was approved I will be moving forward and merging it.

PR #1178, opened 2023-07-28
[AL-6732] fix test by adding sleep

attila-papai:


tomislav-peharda (approved):
hope 5 s will be sufficient

PR #1179, opened 2023-07-28
[SE-2960] Add roles to project members qgl query

jstorts:
[Ticket](https://labelbox.atlassian.net/browse/EF-369)

## Context

A user has requested the roles of project members be returned in the SDK response. This GQL query is already being made by the App. As discussed in the Slack thread, exposing the `accessFrom` field has a risk of introducing latency.

If possible, the updated query should be released for beta testing to further test the performance when used for programmatic access. 

[Slack Thread](https://labelbox.slack.com/archives/C03PTTCEFUJ/p1690372488230829)

## Analysis

[Datadog APM](https://app.datadoghq.com/apm/services/lb-api-public/operations/graphql.execute/resources?env=labelbox-193903-prod-usc1&graphType=flamegraph&resources=qson%3A%28data%3A%28visible%3A%21t%2Chits%3A%28selected%3Atotal%29%2Cerrors%3A%28selected%3Atotal%29%2Clatency%3A%28selected%3Ap50%29%2CtopN%3A%215%2Csearch%3AProjectMemberOverview%29%2Cversion%3A%210%29&shouldShowLegend=true&sort=time&summary=qson%3A%28data%3A%21f%2Cversion%3A%211%29&topGraphs=latency%3Alatency%2Chits%3Aversion_count%2Cerrors%3Aversion_count%2CbreakdownAs%3Apercentage&start=1687976792840&end=1690568792840&paused=false)

The analysis shows that the queries made by both the App and the SDK are somewhat expensive because the resolver has to iterate several collections to compute the result. 

The difference between query performance with the field (in the App) and query performance without the field (in the SDK today) was not large enough to be significant. 

## Description

If merged, this PR adds the `accessFrom` field to the `ProjectMemberOverviewPyApi` query. This field resolves to the role assigned to each member of a project within that project.

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -150,7 +150,7 @@ class Project(DbObject, Updateable, Deletable):
         """
         id_param = "projectId"
         query_str = """query ProjectMemberOverviewPyApi($%s: ID!) {
-             project(where: {id : $%s}) { id members(skip: %%d first: %%d){ id user { %s } role { id name } }
+             project(where: {id : $%s}) { id members(skip: %%d first: %%d){ id user { %s } role { id name } accessFrom }
            }
         }""" % (id_param, id_param, query.results_query_part(Entity.User))
         return PaginatedCollection(self.client, query_str,
@@ -1360,6 +1360,7 @@ class Project(DbObject, Updateable, Deletable):
 class ProjectMember(DbObject):
     user = Relationship.ToOne("User", cache=True)
     role = Relationship.ToOne("Role", cache=True)
+    access_from = Field.String("access_from")
 
 
 class LabelingParameterOverride(DbObject):
--- a/tests/integration/test_user_management.py
+++ b/tests/integration/test_user_management.py
@@ -60,6 +60,7 @@ def test_project_invite(client, organization, project_pack, queries):
                           for proj_role in [project_role_1, project_role_2]])
 
     project_members = project_1.members()
+
     project_member = [
         member for member in project_members
         if member.user().uid == client.get_user().uid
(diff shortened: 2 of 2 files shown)


labelbox/schema/project.py
```diff
@@ -150,7 +150,7 @@ def members(self) -> PaginatedCollection:
         """
         id_param = "projectId"
         query_str = """query ProjectMemberOverviewPyApi($%s: ID!) {
-             project(where: {id : $%s}) { id members(skip: %%d first: %%d){ id user { %s } role { id name } }
+             project(where: {id : $%s}) { id members(skip: %%d first: %%d){ id user { %s } role { id name } accessFrom }
```
msokoloff1, line 153: You need to add the `accessFrom` attribute to ProjectMember. Otherwise it will be ignored.

msokoloff1: I think this is a string. So you can just add Field.String("accessFrom") here https://github.com/Labelbox/labelbox-python/blob/dc44aa8d304afe939a3afa2f41b66fc0268b4c5c/labelbox/schema/project.py#L1363

labelbox/schema/project.py
```diff
@@ -1360,6 +1360,7 @@ def __check_data_rows_have_been_processed(
 class ProjectMember(DbObject):
     user = Relationship.ToOne("User", cache=True)
     role = Relationship.ToOne("Role", cache=True)
+    access_from = Field.String("accessFrom")
```
jstorts, line 1363: I assume this attribute should be snake case instead of `accessFrom`

msokoloff1: oops, yes. Sorry for the wrong info, I haven't contributed to the sdk in a while.

msokoloff1: I think the string arg passed to the Field.String constructor should also be came case.

labelbox/schema/project.py
jstorts: done

labelbox/schema/project.py
jstorts: updated

PR #1184, opened 2023-08-01
[E2E-755] Updated test_feature_schemas.py

mihhail-m:
Updated tests in `test_feature_schemas.py`.

* Added new test to cover more tools and classifications creation/deletion for feature schemas.
* Added new tests.
* Minor code refactoring.

**NB**: Tested both with non-ADV and ADV orgs on `stage`. In both cases tests were passing.


mihhail-m:
@Labelbox/data-io hey, if anyone could take a look and stamp this, that would be great.

PR #1185, opened 2023-08-01
[AL-6294] | Bug Fix | Support for setting a 'from cursor for dataset.data_rows()

vbrodsky:


labelbox/pagination.py
```diff
@@ -134,7 +134,7 @@ class _CursorPagination(_Pagination):
     def __init__(self, cursor_path: List[str], *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.cursor_path = cursor_path
-        self.next_cursor: Optional[Any] = None
+        self.next_cursor: Optional[Any] = kwargs.get('params', {}).get('from')
```
kopreschko, line 137: This change makes the code a bit obscure. To enhance readability and maintainability, what do you think about explicitly adding params to the arguments of the constructor?

kopreschko (commented):
Just added a small observation

labelbox/pagination.py
vbrodsky: I did add it in the past... but the parent init mypy had issues with it... ok let me look at it again

vbrodsky: I had it like you suggested, but had issues similar to https://stackoverflow.com/questions/21764770/typeerror-got-multiple-values-for-argument with mypy.. so I would prob have to 're-add' it back into kwargs before calling the parent constructor. Let me see

kopreschko (approved):

labelbox/pagination.py
```diff
@@ -131,10 +131,10 @@ def get_next_page(self) -> Tuple[Dict[str, Any], bool]:
 
 class _CursorPagination(_Pagination):
 
-    def __init__(self, cursor_path: List[str], *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, cursor_path: List[str], params={}, *args, **kwargs):
+        super().__init__(*args, **dict(kwargs, params=params))
```
kopreschko, line 135: might be shorter to just use ```{**kwargs, **params}```
but I think this is getting weirder, it's probably better to leave it as it was before.

PR #1187, opened 2023-08-01
[AL-5942] Exports v2 integration tests

yamini-labelbox:
https://labelbox.atlassian.net/browse/AL-5942 

- reorganize exports test cases
- add more coverage in terms of validating individual exports attributes

tests/integration/annotation_import/test_data_types.py::test_import_data_types_v2 still has the exports v2 tests for all the datatypes. This test cases does testing for both import and export pipelines. So I let it be here for now. Although it is ideal to have a similar test case in the exports suite. But in view of not duplicating the test cases (which increases the run time), i let it be here. This is the only exports v2 related test case that is not inside the exports_v2 suite. 

- there are a few new ones added in the past week. i will have another PR to move those around once i merge in this one


tests/integration/conftest.py
```diff
@@ -247,6 +248,19 @@ def unique_dataset(client, rand_gen):
     dataset.delete()
 
 
+@pytest.fixture
```
vbrodsky, line 251: we also have a  small_dataset fixture in tests/integration/test_batch.py 
If you have time, can you make sure both this test and that use the same fixture?

tests/integration/export_v2/test__export_slice.py
```diff
@@ -5,7 +5,7 @@
     'Skipping until we have a way to create slices programatically')
 def test_export_v2_slice(client):
     # Since we don't have CRUD for slices, we'll just use the one that's already there
-    SLICE_ID = "clfgqf1c72mk107zx6ypo9bse"
+    SLICE_ID = "clk04g1e4000ryb0rgsvy1dty"
```
vbrodsky, line 8: what did this have to change?

tests/integration/export_v2/test_export_dataset.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+
+
+@pytest.mark.parametrize('data_rows', [5], indirect=True)
```
vbrodsky, line 4: Why does this have to be 5 data rows? Can we reduce to 2?

tests/integration/export_v2/test_export_dataset.py
```diff
@@ -0,0 +1,31 @@
...
+    data_row_ids = [dr.uid for dr in data_rows]
+    params = {"performance_details": False, "label_details": False}
+    task_results = export_v2_test_helpers.run_dataset_export_v2_task(
+        dataset, params=params)
+    assert len(task_results) == len(data_row_ids)
+    assert set([dr['data_row']['id'] for dr in task_results
+               ]) == set(data_row_ids)
+
+
+@pytest.mark.parametrize('data_rows', [5], indirect=True)
```
vbrodsky, line 15: Same, why 5?

tests/integration/export_v2/test_export_project.py
```diff
@@ -0,0 +1,180 @@
+from datetime import datetime, timezone, timedelta
+
+import pytest
+
+from labelbox.schema.media_type import MediaType
+
+IMAGE_URL = "https://storage.googleapis.com/diagnostics-demo-data/coco/COCO_train2014_000000000034.jpg"
```
vbrodsky, line 7: again the same comment, let's use the test artifacts bucket

tests/integration/test_data_row_media_attributes.py
```diff
@@ -1,9 +0,0 @@
-def test_export_empty_media_attributes(client, configured_project_with_label,
```
vbrodsky, line 1: why are these deleted?

tests/integration/conftest.py
```diff
@@ -22,6 +22,7 @@
 from labelbox.schema.user import User
 
 IMG_URL = "https://picsum.photos/200/300.jpg"
+SMALL_DATASET_URL = "https://storage.googleapis.com/diagnostics-demo-data/coco/COCO_train2014_000000000034.jpg"
```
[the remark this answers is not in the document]
yamini-labelbox: what is the name of the bucket? I am kinda confused. Should it be [lb-artifacts-testing] or [labelbox-artifacts]

tests/integration/export_v2/test__export_slice.py
yamini-labelbox: I changed this for testing this on a slice I created locally. It actually doesn't matter. 
This test will always be skipped in the pipeline since we do no have SDK functionality create a slice. This test exists so that one can sanity check the slice export functionality locally.

tests/integration/conftest.py
yamini-labelbox: Good catch. Will do that.

tests/integration/conftest.py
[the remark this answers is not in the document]
yamini-labelbox: nvm got it.

tests/integration/export_v2/test_export_dataset.py
vbrodsky: well as I am working right now to optimize fixtures, it looks like the biggest bottleneck is creation of data rows, because this has to go to es and come back
I can not say that 2 vs 5 will make a world of difference, but generally I think we never need more than 2 data rows in our integration tests to represent the concept of 'many'.

PR #1192, opened 2023-08-03
[PTDT-1552] Add LLM media types

lgluszek:
Jira issue: [Can't fetch a list of projects in SDK when a list contains an LLM project](https://labelbox.atlassian.net/browse/PTDT-1552)

How to test:
1. Add an LLM project
2. Run this code

```
projects = client.get_projects()
for project in projects:
    print(project.name)
```

3. Check if it works

image

### DIFF
--- a/labelbox/schema/media_type.py
+++ b/labelbox/schema/media_type.py
@@ -10,16 +10,16 @@ class MediaType(Enum):
     Html = "HTML"
     Image = "IMAGE"
     Json = "JSON"
+    LLMPromptCreation = "LLM_PROMPT_CREATION"
+    LLMPromptResponseCreation = "LLM_PROMPT_RESPONSE_CREATION"
     Pdf = "PDF"
     Simple_Tile = "TMS_SIMPLE"
     Text = "TEXT"
     Tms_Geo = "TMS_GEO"
     Tms_Simple = "TMS_SIMPLE"
-    Video = "VIDEO"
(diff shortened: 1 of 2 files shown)


labelbox/schema/media_type.py
```diff
@@ -18,6 +18,8 @@ class MediaType(Enum):
     Video = "VIDEO"
     Unknown = "UNKNOWN"
     Unsupported = "UNSUPPORTED"
+    LLMPromptCreation = "LLM_PROMPT_CREATION"
```
kkim-labelbox, line 21: Could you order these alphabetically in the file?

labelbox/schema/media_type.py
```diff
@@ -41,8 +43,10 @@ def is_supported(cls, value):
     @classmethod
     def get_supported_members(cls):
         return [
-            item for item in cls.__members__
-            if item not in ["Unknown", "Unsupported"]
+            item for item in cls.__members__ if item not in [
```
kkim-labelbox, line 46: Do we need to add LLM types as unsupported types here? Shouldn't adding the media types make `get_projects()` work with LLM projects?

lgluszek: It will work without it, but tests started failing.

lgluszek: @pd-lb do you remember which test started failing?

kkim-labelbox (approved):

PR #1197, opened 2023-08-04
[QQC-2355] [WIP] Provide SDK methods to create multiple batches given a dataset

richardsun0713:


labelbox/schema/project.py
```diff
@@ -773,6 +779,158 @@ def create_batch(
...
+                "batchNamePrefix": name_prefix,
+                "dataRowIds": dr_ids,
+                "globalKeys": global_keys,
+                "priority": priority,
+                "consensusSettings": consensus_settings
+            }
+        }
+
+        tasks = self.client.execute(
+            mutation_str, params, timeout=180.0,
```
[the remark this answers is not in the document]
richardsun0713: I will use the default timeout here since the latency characteristics are different

labelbox/client.py
```diff
@@ -1704,3 +1704,27 @@ def unarchive_feature_schema_node(self, ontology_id: str,
...
+                          project(where: {id: $projectId}) {
+                             batches(where: {id: $batchId}) {
+                                nodes {
+                                   %s
+                                }
+                             }
+                        }
+                    }
+                    """ % ("getProjectBatchPyApi",
+                           query.results_query_part(Entity.Batch))
```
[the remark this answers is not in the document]
richardsun0713: I'll leave the decision on this to your team - and follow existing patterns here for now to maintain code consistency

labelbox/schema/create_batches_task.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        Waits for the task to complete.
+
+        Args:
+            timeout_seconds: the number of seconds to wait before timing out
+
+        Returns: None
+        """
+
+        for task in self.tasks:
```
vbrodsky, line 36: [optional] how many tasks we expect here? would running these in parallel be worth doing it?

labelbox/schema/create_batches_task.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def result(self):
+        """
+        Returns the batches created by the task.
+
+        Returns: the list of batches created by the task
+        """
+
+        return [
+            self.client.get_batch(self.project_id, batch_id)
```
vbrodsky, line 65: again the same question: should we run these in parallel? or not worth it?

labelbox/schema/project.py
```diff
@@ -773,6 +779,156 @@ def create_batch(
...
+            "input": {
+                "batchNamePrefix": name_prefix,
+                "dataRowIds": dr_ids,
+                "globalKeys": global_keys,
+                "priority": priority,
+                "consensusSettings": consensus_settings
+            }
+        }
+
+        tasks = self.client.execute(
```
vbrodsky, line 856: what happens if mutation returns an error? I guess `client` will just throw?

labelbox/schema/project.py
```diff
@@ -773,6 +779,156 @@ def create_batch(
             return self._create_batch_sync(name, dr_ids, global_keys, priority,
                                            consensus_settings)
 
+    def create_batches(
+        self,
+        name_prefix: str,
+        data_rows: Optional[List[Union[str, DataRow]]] = None,
```
vbrodsky, line 785: is the number of data rows 'infinite'? i.e. do we know whether it will blow up after certain large number of rows?

labelbox/schema/project.py
```diff
@@ -871,29 +1027,7 @@ def _create_batch_async(self,
...
-            get_batch_str, {
-                "projectId": self.uid,
-                "batchId": batch_id
-            },
-            timeout=180.0,
-            experimental=True)["project"]["batches"]["nodes"][0]
-
-        # TODO async endpoints currently do not provide failed_data_row_ids in response
-        return Entity.Batch(self.client, self.uid, batch)
+        return self.client.get_batch(self.uid, batch_id)
```
vbrodsky, line 1030: nice refactor

tests/integration/conftest.py
```diff
@@ -768,3 +768,37 @@ def is_adv_enabled(client) -> bool:
...
+def big_dataset(dataset: Dataset):
+    task = dataset.create_data_rows([
+        {
+            "row_data": IMAGE_URL,
+            "external_id": EXTERNAL_ID
+        },
+    ] * 3)
+    task.wait_till_done()
+
+    yield dataset
```
vbrodsky, line 787: I think you have just moved this code, but anyway, should now we delete these data rows after `yield`?

tests/integration/test_batches.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+
+from labelbox import Project, Dataset
+
+
+def test_create_batches(project: Project, big_dataset: Dataset):
+    data_rows = [dr.uid for dr in list(big_dataset.export_data_rows())]
```
vbrodsky, line 7: to save on an extra api call, can the `big_dataset` fixture just return those data rows?

tests/integration/test_batches.py
```diff
@@ -0,0 +1,33 @@
...
+    assert task.errors() is None
+    batches = task.result()
+
+    assert len(batches) == 1
+    assert batches[0].name == "test-batch0000"
+    assert batches[0].size == len(data_rows)
+
+
+def test_create_batches_from_dataset(project: Project, big_dataset: Dataset):
+    data_rows = [dr.uid for dr in list(big_dataset.export_data_rows())]
```
vbrodsky, line 20: same comment as above

labelbox/schema/create_batches_task.py
richardsun0713: 10 max at the moment, we can run these in parallel as a follow-up if the limit is increased

labelbox/schema/project.py
richardsun0713: Yes, `client` will throw an error with the message that the backend services passed back

tests/integration/test_batches.py
richardsun0713: Possibly - this fixture is used by other tests as well so it would depend on what those tests need the dataset for

labelbox/client.py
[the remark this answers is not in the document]
kkim-labelbox: @mnoszczak chatted with Richard about this offline. We'll go ahead with GQL APIs for now, since airbnb needs this ASAP, and it's a sizable lift to convert the existing APIs to REST. Orch team will convert these to REST in the future

tests/integration/test_batches.py
richardsun0713: Looks like some tests need more than the data row ids, so inlining wouldn't work for all cases

tests/integration/conftest.py
richardsun0713: They should be deleted as part of the teardown https://github.com/Labelbox/labelbox-python/blob/f0d5fe93e9b3d52efe532408dcbd4499425baf9a/tests/integration/conftest.py#L241

PR #1199, opened 2023-08-08
[AL-6733] Add ADV specific assertion

attila-papai:
An ADV import task won't fail when invalid metadata was given in the payload so a slightly different assertion is needed.


tests/integration/test_data_rows.py
```diff
@@ -475,19 +475,26 @@ def create_data_row(data_rows):
...
+        DataRowMetadataField(schema_id=TEXT_SCHEMA_ID, value="some msg"))
 
     task = dataset.create_data_rows([{
         DataRow.row_data: image_url,
         DataRow.metadata_fields: fields
     }])
     task.wait_till_done()
-    assert task.status == "FAILED"
-    assert len(task.failed_data_rows) > 0
+    if is_adv_enabled:
```
attila-papai, line 490: @kkim-labelbox another difference in ADV
I documented this [here](https://labelbox.atlassian.net/wiki/spaces/AIENG/pages/2240020512/Known+Breaking+Changes+From+Pre-ADV+to+ADV)

tests/integration/test_data_rows.py
```diff
@@ -475,19 +475,26 @@ def create_data_row(data_rows):
...
+        assert task.status == "COMPLETE"
```
shaochen-lb, line 491: The failure says:
```
>       assert task.status == "FAILED"                                                                                                      
E       AssertionError: assert 'IN_PROGRESS' == 'FAILED'                                                                                    
E         - FAILED                                                                                                                          
E         + IN_PROGRESS 
```

attila-papai: where do you see this?

attila-papai: so when you see IN_PROGRESS then it is a service failure

tests/integration/test_data_rows.py
kkim-labelbox: Is this because ADV partially fails now when there's an error with validation of metadata on one data row?

attila-papai: It partially _succeeds_ instead. What can be imported, will be imported. The rest will be reported as failures.

tests/integration/test_data_rows.py
```diff
@@ -475,19 +475,27 @@ def create_data_row(data_rows):
...
         DataRow.metadata_fields: fields
     }])
-    task.wait_till_done()
-    assert task.status == "FAILED"
-    assert len(task.failed_data_rows) > 0
+    task.wait_till_done(timeout_seconds=60)
+    if is_adv_enabled:
+        assert task.status == "COMPLETE"
+        assert len(task.failed_data_rows) == 1
+        print(task.failed_data_rows[0])
```
vbrodsky, line 493: why are you printing stuff?

vbrodsky (approved):

tests/integration/test_data_rows.py
attila-papai: ups

PR #1210, opened 2023-08-15
[AL-6743] skip invalid media type validation when ADV is enabled

attila-papai:
ADV does not validate the passed media type hint when performing async import jobs, so we shouldn't test that either.

### DIFF
--- a/tests/integration/test_data_rows.py
+++ b/tests/integration/test_data_rows.py
@@ -973,7 +973,7 @@ def test_create_conversational_text(dataset, conversational_content):
             data_row.row_data).json() == conversational_content['row_data']
 
 
-def test_invalid_media_type(dataset, conversational_content):
+def test_invalid_media_type(dataset, conversational_content, is_adv_enabled):
     for error_message, invalid_media_type in [[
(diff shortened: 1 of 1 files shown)


tests/integration/test_data_rows.py
```diff
@@ -973,17 +973,21 @@ def test_create_conversational_text(dataset, conversational_content):
...
     ], ["Found invalid media type: 'totallyinvalid'", 'totallyinvalid']]:
         # TODO: What error kind should this be? It looks like for global key we are
         # using malformed query. But for invalid contents in FileUploads we use InvalidQueryError
         with pytest.raises(labelbox.exceptions.InvalidQueryError):
             dataset.create_data_rows_sync([{
-                **conversational_content, 'media_type': invalid_media_type
+                **conversational_content, 'media_type': 'IMAGE'
             }])
 
+        if is_adv_enabled:
```
vbrodsky, line 987: >ADV does not validate the passed media type hint when performing async import jobs

So why not just move this check to the beginning of the function

attila-papai: I still need the first assertion in the `with` block

vbrodsky (approved):

PR #1214, opened 2023-08-17
[AL-6729] assert adv task errors separately

attila-papai:
We can't use a snapshot because the error ADV returns contain a unique error code for easier backtracking.


vbrodsky (approved):
I guess :)

PR #1218, opened 2023-08-22
[QQC-2355] Limit number of data rows to check for processing status at once

richardsun0713:
There is a limit of 100k data rows that can be checked at once for processing status. Split this call to the backend into chunks of 100k

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -1454,12 +1454,35 @@ class Project(DbObject, Updateable, Deletable):
         """ Wait until all the specified data rows are processed"""
         start_time = datetime.now()
 
+        max_data_rows_per_poll = 100_000
+        if data_row_ids is not None:
+            for i in range(0, len(data_row_ids), max_data_rows_per_poll):
+                chunk = data_row_ids[i:i + max_data_rows_per_poll]
+                self._poll_data_row_processing_status(
+                    chunk, [], start_time, wait_processing_max_seconds,
+                    sleep_interval)
+
+        if global_keys is not None:
+            for i in range(0, len(global_keys), max_data_rows_per_poll):
+                chunk = global_keys[i:i + max_data_rows_per_poll]
+                self._poll_data_row_processing_status(
+                    [], chunk, start_time, wait_processing_max_seconds,
+                    sleep_interval)
+
+    def _poll_data_row_processing_status(
+            self,
+            data_row_ids: List[str],
+            global_keys: List[str],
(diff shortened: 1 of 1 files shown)


labelbox/schema/project.py
```diff
@@ -1454,12 +1454,35 @@ def _wait_until_data_rows_are_processed(
         """ Wait until all the specified data rows are processed"""
         start_time = datetime.now()
```
kkim-labelbox, line 1455: Since the start_time is set to the beginning of this function, I think `_poll_data_row_processing_status` may not wait the full `wait_processing_max_seconds` per chunk

labelbox/schema/project.py
```diff
@@ -1454,12 +1454,35 @@ def _wait_until_data_rows_are_processed(
...
 
+        max_data_rows_per_poll = 100_000
+        if data_row_ids is not None:
+            for i in range(0, len(data_row_ids), max_data_rows_per_poll):
```
vbrodsky, line 1459: nice use of step for chunking

labelbox/schema/project.py
```diff
@@ -1454,12 +1454,35 @@ def _wait_until_data_rows_are_processed(
...
+            sleep_interval=30):
+
         while True:
             if (datetime.now() -
                     start_time).total_seconds() >= wait_processing_max_seconds:
                 raise ProcessingWaitTimeout(
-                    "Maximum wait time exceeded while waiting for data rows to be processed. Try creating a batch a bit later"
-                )
+                    "Maximum wait time exceeded while waiting for data rows to be processed. "
+                    "Try creating a batch a bit later")
```
vbrodsky, line 1485: is it supposed to be a multi-line string? does not work? I though smth like
```suggestion
                    raise ProcessingWaitTimeout("""Maximum wait time exceeded while waiting for data rows to be processed.
                    Try creating a batch a bit later""")
```
based on https://stackoverflow.com/questions/51449888/valueerror-compatibility-and-multi-line-string-aesthetics

labelbox/schema/project.py
richardsun0713: My understanding was that the `wait_processing_max_seconds` was to be applied to the entire call for `_wait_until_data_rows_are_processed`, and not the individual chunks. This should preserve original behavior

kkim-labelbox: Ah gotcha makes sense

kkim-labelbox (approved):

PR #1224, opened 2023-08-25
[QQC-2474] Expose quality mode through SDK create_project method

richardsun0713:
* Removes the `auto_audit_percentage` and `auto_audit_number_of_labels` parameters from `labelbox.client.Client.create_project` and replaces them with the `quality_mode` enum

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -31,6 +31,7 @@ from labelbox.schema.model import Model
 from labelbox.schema.model_run import ModelRun
 from labelbox.schema.ontology import Ontology, Tool, Classification, FeatureSchema
 from labelbox.schema.organization import Organization
+from labelbox.schema.quality_mode import QualityMode
 from labelbox.schema.user import User
 from labelbox.schema.project import Project
 from labelbox.schema.role import Role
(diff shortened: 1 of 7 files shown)


labelbox/client.py
```diff
@@ -636,16 +637,33 @@ def create_project(self, **kwargs) -> Project:
             description (str): A short summary for the project
             media_type (MediaType): The type of assets that this project will accept
             queue_mode (Optional[QueueMode]): The queue mode to use
-            auto_audit_percentage (Optional[float]): The percentage of data rows that will require more than 1 label
-            auto_audit_number_of_labels (Optional[float]): Number of labels required for data rows selected for multiple labeling (auto_audit_percentage)
+            quality_mode (Optional[QualityMode]): The quality mode to use (e.g. Benchmark, Consensus). Defaults to
```
vbrodsky, line 640: what happens to existing projects where these attributes are set and, say, we do smth like `client.get_project()`. I suppose it would work? But from the client point of view, is there any migration needs to happen, i.e. if auto_audit_percentage and auto_audit_number_of_labels => set quality mode etcv

labelbox/client.py
```diff
@@ -636,16 +637,33 @@ def create_project(self, **kwargs) -> Project:
...
+                Benchmark
         Returns:
             A new Project object.
         Raises:
             InvalidAttributeError: If the Project type does not contain
                 any of the attribute names given in kwargs.
         """
-        media_type = kwargs.get("media_type")
+
+        auto_audit_percentage = kwargs.get("auto_audit_percentage")
```
vbrodsky, line 649: if these attributes are removed from the model, why are we dealing with them here?

labelbox/client.py
vbrodsky: Also if the quality mode default to Benchmark, shouldn't we set it here?

labelbox/client.py
```diff
@@ -658,20 +676,23 @@ def create_project(self, **kwargs) -> Project:
...
-        if not queue_mode:
-            logger.warning(
-                "Default createProject behavior will soon be adjusted to prefer "
-                "batch projects. Pass in `queue_mode` parameter explicitly to opt-out for the "
-                "time being.")
-        elif queue_mode == QueueMode.Dataset:
-            logger.warning(
-                "QueueMode.Dataset will eventually be deprecated, and is no longer "
-                "recommended for new projects. Prefer QueueMode.Batch instead.")
+        quality_mode = kwargs.get("quality_mode")
```
vbrodsky, line 679: Looks like dupe lines, since we do the same line 684-..

labelbox/client.py
```diff
@@ -658,20 +676,23 @@ def create_project(self, **kwargs) -> Project:
...
+        if not quality_mode:
+            logger.info("Defaulting quality mode to Benchmark.")
+
+        data = kwargs
+        data.pop("quality_mode", None)
+        if quality_mode is None or quality_mode is QualityMode.Benchmark:
+            data["auto_audit_number_of_labels"] = 1
```
vbrodsky, line 686: suggest making these numbers constants

labelbox/client.py
```diff
@@ -658,20 +676,23 @@ def create_project(self, **kwargs) -> Project:
...
+            data["auto_audit_percentage"] = 1
+        else:
+            data["auto_audit_number_of_labels"] = 3
```
vbrodsky, line 689: did you try using the `@property` decorator for auto_audit_percentage and auto_audit_number_of_labels.

labelbox/schema/quality_mode.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class QualityMode(str):
```
vbrodsky, line 4: Did you mean `QualityMode(str, Enum)` ?

labelbox/client.py
richardsun0713: `auto_audit_percentage` and `auto_audit_number_of_labels` are still in the Project model but have no meaning anymore - it has been replaced with equivalents at the batch level (specified during batch creation). The migration has already been performed on the backend, this change simply removes the old obsolete data from the client model

labelbox/client.py
richardsun0713: This is here to help present a more readable message that can inform users how to migrate to the new usage model

labelbox/client.py
richardsun0713: These aren't duplicate lines; one exists to inform the user that a default has been selected, and the other exists to translate the given quality_mode to something our backend can interpret

vbrodsky (approved):

PR #1225, opened 2023-08-25
Vb/bulk delete datarow id sdk 33

vbrodsky:
[EXPERIMENTAL]

Story: https://labelbox.atlassian.net/browse/SDK-33

The original intent for this story was to reduce friction in bulk deleting data rows and allow using both data row ids and global keys (currently a user passes an array of whole `DataRow`s plus there is a 4K limitation on total number of data rows)

As I started working, I have discovered that we have a different mutation (not the one used in bulk_delete) that allows us to delete data rows by ids of global keysn and at bulk numbers too

The most important point to review in this PR is user-facing method signature and packaging since, in the future, we will rewrite the internal implementation

- Even though we already have a `DataRow.bulk_delete(data_rows)`, I thought adding more parameters might be confusing for a caller and will make implementation convoluted since this method uses a generic way to bulk_delete any entity and the entity itself must be present (not the ids or global keys)
- I have implement, instead, in the `client` which has it's own drawbacks. Our client has been polluted with many methods and I feel like the only reason we stick it there because we always need access to the client and we have no easy way to inject it.

### DIFF
--- a/labelbox/schema/es_filters.py
+++ b/labelbox/schema/es_filters.py
@@ -1,0 +1,27 @@
+import sys
+
+from datetime import datetime, timezone
+from typing import Collection, Dict, Tuple, List, Optional
+from labelbox.typing_imports import Literal
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+SEARCH_LIMIT_PER_EXPORT_V2 = 2_000
+ISO_8061_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
+
+
+def build_id_filters(
+        ids: list,
+        type_name: str,
+        search_where_limit: int = SEARCH_LIMIT_PER_EXPORT_V2) -> dict:
+    if not isinstance(ids, list):
+        raise ValueError(f"{type_name} filter expects a list.")
+    if len(ids) == 0:
+        raise ValueError(f"{type_name} filter expects a non-empty list.")
+    if len(ids) > search_where_limit:
+        raise ValueError(
+            f"{type_name} filter only supports a max of {search_where_limit} items."
+        )
+    return {"ids": ids, "operator": "is", "type": type_name}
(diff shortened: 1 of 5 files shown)


labelbox/schema/es_filters.py
```diff
@@ -0,0 +1,27 @@
...
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+SEARCH_LIMIT_PER_EXPORT_V2 = 2_000
+ISO_8061_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
+
+
+def build_id_filters(
```
vbrodsky, line 15: I have refactored this because I thought it would be useful for my story. As we have changed the direction from using the `DeleteDataRowsByQuery` mutation to `DeleteDataRows`, we do not need to use this code. Still I have decided to keep the refactor since it is good and has prior unit tests.

labelbox/schema/es_filters.py
```diff
@@ -0,0 +1,16 @@
+SEARCH_LIMIT = 2_000
```
vbrodsky, line 1: just refactored existing code to a sep file for reuse

vbrodsky:
re: `data_rows: Union[List[str], List[DataRow]]`, this might be the core of discussion we should focus on @mnoszczak and everyone.

I personally feel that most of the time we use `Union` in Python typing it is a red flag and I would not use it most type. If agree a type hint in python is the same as a type in a strongly typed language what does a type `Union` stand for? A type that could be any of the 2 types? 

Well the thought is to make it easier for a user.. so they can do things like

vbrodsky:
@mnoszczak this pattern

vbrodsky:
Also for `def bulk_delete(data_rows: Optional[List[Union[str, DataRow]]] = None, global_keys: Optional[List[str]]` we need to figure out how to instantiate a client for the case when data row ids or global keys

PR #1230, opened 2023-09-01
DEVOPS-1626 - Remove overriding of permissions

lhriley:
These permissions seem to be interfering with the ability to push artifacts.


lhriley:
success: https://github.com/Labelbox/labelbox-python/actions/runs/6044731260/job/16403738870

PR #1239, opened 2023-09-12
Vb/test annotation upload by global keys

vbrodsky:
Story: https://labelbox.atlassian.net/browse/SDK-237

This PR adds sdk integration test for predictions upload with a data row global key instead of a uuid

Additional improvements:
- found fixed a test case where we stilled used ndjson lib instead of our parser
- for `configured_project`, our popular fixture that creates a data row for each annotation, added automatically-generated global key


tests/data/serialization/ndjson/test_video.py
```diff
@@ -12,7 +12,6 @@
 from labelbox import parser
 
 from labelbox.data.serialization.ndjson.converter import NDJsonConverter
-from labelbox.schema.annotation_import import MALPredictionImport
```
vbrodsky, line 15: not used

PR #1244, opened 2023-09-27
Fix bulk_export() not exporting global_key(s)

yamini-labelbox:
https://labelbox.atlassian.net/browse/SDK-198

SDK tests will fail for this PR until the corresponding API [change](https://github.com/Labelbox/intelligence/pull/17545) is merged.


kkim-labelbox:
@yamini-labelbox do we need test coverage on this?

PR #1254, opened 2023-10-04
Update the exporter video test to check for frame rate and count

vbrodsky:


yamini-labelbox (approved):
LGTM. Approving. Looks like the SDK tests will pass only after merging the monorepo change.

PR #1262, opened 2023-10-17
[SDK-239] Fix dataset types

mnoszczak:
Makes the types more correct

### DIFF
--- a/labelbox/schema/dataset.py
+++ b/labelbox/schema/dataset.py
@@ -68,8 +68,8 @@ class Dataset(DbObject, Updateable, Deletable):
 
     def data_rows(
         self,
-        from_cursor: str = None,
-        where: Comparison = None,
+        from_cursor: Optional[str] = None,
+        where: Optional[Comparison] = None,
     ) -> PaginatedCollection:
         """ 
         Custom method to paginate data_rows via cursor.


apollonin (approved):

kopreschko (approved):
lgtm

vbrodsky (commented):
Not sure we need Optional here... please see this too https://stackoverflow.com/questions/69440494/python-3-10-optionaltype-or-type-none#:~:text=Optional%20was%20invented%20to%20indicate,gets%20rid%20of%20an%20identifier.

vbrodsky (commented):
Also here we have a default None, so Optional is not required:
https://docs.python.org/3/library/typing.html#typing.Optional

labelbox/schema/dataset.py
```diff
@@ -68,8 +68,8 @@ class Dataset(DbObject, Updateable, Deletable):
 
     def data_rows(
         self,
-        from_cursor: str = None,
-        where: Comparison = None,
+        from_cursor: Optional[str] = None,
```
vbrodsky, line 71: this is tricky, but correct change I believe, based on https://docs.python.org/3/library/typing.html#typing.Optional

vbrodsky (approved):

PR #1271, opened 2023-10-25
Handle 409 exception from ADV when there is a global Key conflict

yamini-labelbox:
https://labelbox.atlassian.net/browse/SDK-276

Currently, globalKey conflicts are treated as internal server errors because of which the query is retried.
This change handles the [HttpStatus.CONFLICT](https://github.com/Labelbox/intelligence/blob/develop/packages/datavault/adv/src/main/kotlin/com/labelbox/adv/rest/ErrorController.kt#L73-L75) as MalformedQueryException on the SDK side.

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -215,6 +215,7 @@ class Client:
                 "Failed to parse response as JSON: %s" % response.text)
 
         errors = r_json.get("errors", [])
+        print(errors)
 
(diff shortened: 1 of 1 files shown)


yamini-labelbox:
closing this as we are handling this on the api side https://github.com/Labelbox/intelligence/pull/17841

PR #1277, opened 2023-11-07
add support for image bytes as a source for raster video mask import

apollonin:
Make support for video mask to be provided as a byte string in addition to URL

### DIFF
--- a/labelbox/data/annotation_types/video.py
+++ b/labelbox/data/annotation_types/video.py
@@ -1,7 +1,7 @@
 from enum import Enum
 from typing import List, Optional, Tuple
 
-from pydantic import BaseModel, validator
+from pydantic import BaseModel, validator, root_validator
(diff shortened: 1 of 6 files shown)


labelbox/data/annotation_types/video.py
```diff
@@ -81,14 +81,24 @@ class DICOMObjectAnnotation(VideoObjectAnnotation):
...
-        extra (Dict[str, Any])        
+        extra (Dict[str, Any])
     """
     group_key: GroupKey
 
 
 class MaskFrame(_CamelCaseMixin, BaseModel):
     index: int
-    instance_uri: str
+    instance_uri: Optional[str] = None
```
vbrodsky, line 91: `content: Union[str,byte]` wdyt?
and you could have helper methods like
```
def instance_uri(self):
  # if self.content is str, return it else return None etc.
```

vbrodsky (approved):

PR #1281, opened 2023-11-14
[SDK-400] Validate url, secret, and topic list on webhook creation.

sfendell-labelbox:
Check that these fields are non-empty on creation to match frontend behavior.


labelbox/schema/webhook.py
```diff
@@ -77,9 +77,15 @@ def create(client, topics, url, secret, project) -> "Webhook":
             ValueError: If the topic is not one of Topic or status is not one of Status
 
         Information on configuring your server can be found here (this is where the url points to and the secret is set).
-                        https://docs.labelbox.com/en/configure-editor/webhooks-setup#setup-steps
+                        https://docs.labelbox.com/reference/webhook
```
sfendell-labelbox, line 80: A little out of scope, but this docs link is broken.

vbrodsky: nice!

labelbox/schema/webhook.py
```diff
@@ -77,9 +77,15 @@ def create(client, topics, url, secret, project) -> "Webhook":
...
 
         """
+        if not secret:
+            raise TypeError(f"Secret must be a non-empty string.")
```
vbrodsky, line 84: I think `ValueError` is more appropriate here

sfendell-labelbox: Thanks, fixed.

labelbox/schema/webhook.py
```diff
@@ -77,9 +77,15 @@ def create(client, topics, url, secret, project) -> "Webhook":
             ValueError: If the topic is not one of Topic or status is not one of Status
 
         Information on configuring your server can be found here (this is where the url points to and the secret is set).
-                        https://docs.labelbox.com/en/configure-editor/webhooks-setup#setup-steps
+                        https://docs.labelbox.com/reference/webhook
 
         """
+        if not secret:
+            raise ValueError(f"Secret must be a non-empty string.")
```
kopreschko, line 84: There's no need for an f-string here, as there are no variables/expressions in the string. This goes for all the messages below.

tests/integration/test_webhook.py
```diff
@@ -40,3 +40,39 @@ def test_webhook_create_update(project, rand_gen):
         "Topics must be List[Webhook.Topic]. Found `invalid..`"
 
     webhook.delete()
+
+
```
kopreschko, line 44: These, in my opinion, should be considered unit-tests.

kopreschko (commented):
Nice work in both the API and SDK by adding tests for these validations! Just some minor observations, the rest looks good to me.

labelbox/schema/webhook.py
sfendell-labelbox: Fixed.

tests/integration/test_webhook.py
sfendell-labelbox: There's a lot of tests in `integration` that could be better considered a unit test, but this needs a configured project and client which only exist in integration fixtures, and pulling them out would be too much of a pain for this I think.

kopreschko: `client` and `project` could be mocks, the exceptions are thrown before these are used.

sfendell-labelbox: Good point. I moved them to unit.

vbrodsky: Second Klaus comments, after that can approve

sfendell-labelbox: It should be fixed now and checks are passing, can I get an approval?

kopreschko (approved):
nice work Sam!

PR #1284, opened 2023-11-15
Add global key support for Project move_data_rows_to_task_queue

vbrodsky:
Story: https://labelbox.atlassian.net/browse/SDK-414
API PR: https://github.com/Labelbox/intelligence/pull/18036

- We are implementing the `Identifiable` hierarchy to ensure a consistent approach in handling data row IDs and global keys.
- To support both legacy string inputs and the new `Identifiable` objects, we are utilizing the @overload decorator. 
- Our plan is to apply this same methodology to other methods that require the management of both data row IDs and global keys.


labelbox/schema/project.py
```diff
@@ -643,7 +639,7 @@ def labeler_performance(self) -> PaginatedCollection:
         def create_labeler_performance(client, result):
             result["user"] = Entity.User(client, result["user"])
             # python isoformat doesn't accept Z as utc timezone
-            result["lastActivityTime"] = datetime.fromisoformat(
+            result["lastActivityTime"] = utils.format_iso_from_string(
```
vbrodsky, line 642: we have gotten rid of this monkeypatch sometime ago

labelbox/schema/identifiable.py
```diff
@@ -0,0 +1,63 @@
+from abc import ABC
+from enum import Enum
+from typing import List, Union
+from typing_extensions import TypeAlias
```
kopreschko, line 4: TypeAlias seems to be available since 3.10

vbrodsky: yes, but I am using `typing_extensions` that back-ported it. However `TypeAlias` is not strictly necessary, it is just for improved typing. Should I remove it?

labelbox/schema/identifiable.py
```diff
@@ -0,0 +1,63 @@
...
+    GlobalKey = "GLOBAL_KEY"
+
+
+class Identifiable(ABC):
+
+    def __init__(self, keys: Union[str, List[str]]):
+        self._keys = keys
+        if isinstance(keys, str):
+            self.keys = [keys]
+        self._id_type = IdType.DataRowId
```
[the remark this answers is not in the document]
vbrodsky: I am setting it now in derived classes, but why should it be in the constructor? Actually I  think it should be a const

labelbox/schema/identifiable.py
```diff
@@ -0,0 +1,63 @@
...
+        self._keys = keys
+        if isinstance(keys, str):
+            self.keys = [keys]
+
+    @property
+    def id_type(self):
+        return self._id_type
+
+    def __eq__(self, other):
+        return other.keys == self.keys
```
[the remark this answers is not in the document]
vbrodsky: I wanted to keep it as a method in case we want to amend the `keys` getter in the future and it will then keep the semantics for `__eq__`

labelbox/schema/identifiable.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self, keys: Union[str, List[str]]):
+        super().__init__(keys)
+        self._id_type = IdType.GlobalKey
+
+
+DefaultIdentifiable: TypeAlias = UniqueIds
+DataRowIdentifiers: TypeAlias = Union[UniqueIds, GlobalKeys]
+
+
+def strings_to_identifiable(keys: Union[str, List[str]]) -> DefaultIdentifiable:
```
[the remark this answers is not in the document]
vbrodsky: ok that is cool

labelbox/schema/identifiable.py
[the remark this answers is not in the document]
kopreschko: I can think of at least two reasons:
- the inclusion of `id_type` in the constructor of `Identifiable` explicitly communicates the necessity of specifying the type when constructing an Identifiable object
- `Identifiable` should refrain from making assumptions about the default type. Assuming a default type without explicit assignment by derived classes could lead to potential issues, especially when new derived classes are introduced in the future.

kopreschko (approved):
lgtm

labelbox/schema/identifiable.py
```diff
@@ -0,0 +1,65 @@
...
+    @classmethod
+    def strings_to_identifiable(cls, keys: Union[str, List[str]]):
+        return cls(keys)
+
+    def __init__(self, keys: Union[str, List[str]]):
+        super().__init__(keys, IdType.GlobalKey)
+
+
+DefaultIdentifiable = UniqueIds
+DataRowIdentifiers = Union[UniqueIds, GlobalKeys]
```
kopreschko, line 65: missing a newline at the end of the file

whistler: From a user point of view, it does not seem clear that UniqueIds is DataRowIds, since global keys are also unique and they are also ids. What do you think of creating a sub-class / alias for `UniqueIds` called `DataRowIds`

vbrodsky: Hi @whistler I am trying to avoid 'over-classing' but I see your point. Would something like this help
`DataRowIds = UniqueIds` i.e. kind of class alias
PS We will also document this in a number of places

adrian-chang:
LGTM, i'll wait on the engineers to give the final approval

labelbox/schema/identifiable.py
```diff
@@ -0,0 +1,65 @@
+from abc import ABC, abstractmethod
+from enum import Enum
+from typing import List, Union
+
+
+class IdType(str, Enum):
+    DataRowId = "UID"
```
sqlboy, line 7: in ADV these are ID and GKEY, these are used by the census / bucket scanning APIS.

labelbox/schema/identifiable.py
```diff
@@ -0,0 +1,65 @@
...
+    GlobalKey = "GLOBAL_KEY"
+
+
+class Identifiable(ABC):
```
sqlboy, line 11: Confused why this is a wrapper around an array, from initial designs this is basically a wrapper around a single id.

labelbox/schema/identifiable.py
sqlboy: Confused why this code even exists really.

whistler: Yes @vbrodsky, that seems good to me.

labelbox/schema/identifiable.py
sqlboy: Maybe its a matter of renaming it to Identifiables

sqlboy: The idea with Identifiable ABC is that its an interface that lets code know the object is an entity with a unique Id and it provides some function for code to fetch the id structure.  This is more like a utility class, but since its a list you should implement __iter__ and other functions so it can act like a collection.

labelbox/schema/identifiable.py
vbrodsky: ok I can change it, but in a follow up PR because this will require the api update too

labelbox/schema/identifiable.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @keys.setter
+    def keys(self, keys):
+        self._keys = keys
+        if isinstance(keys, str):
+            self._keys = [keys]
+
+    @classmethod
+    @abstractmethod
+    def strings_to_identifiable(cls, keys: Union[str, List[str]]):
```
sqlboy, line 31: Here there is no reason to duplicate the argument type being passed in.  (strings_). The function params already says its strings.  Many libs use like of() or from(), like:

```
keys  = GlobalKeys.of(["some key"])
```

For example in spanner library, its ```Key.of("value")```

labelbox/schema/identifiable.py
vbrodsky: Implemented as an iter, that is a good idea thank you @sqlboy

labelbox/schema/identifiables.py
```diff
@@ -0,0 +1,33 @@
+from enum import Enum
```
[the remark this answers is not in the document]
vbrodsky: of course

sqlboy (approved):

Issue #1261, opened 2023-10-17
Python 3.10 / 3.11 official support


SimonSchick:
According to https://github.com/Labelbox/labelbox-python/releases/tag/v.3.53.0 3.10 is officially supported, but your `README` doesn't reflect that, you might want to update.

Further I'd like to enquire if this package is generally compatible with python 3.11 or not.

adam-homeboost:
Same question for python 3.12

paultancre:
We will update README, appreciate the feedback @SimonSchick.
For clarity requirements will stipulate the following versions : 3.7, 3.8, 3.9 and 3.10

PR #1285, opened 2023-11-17
Update README.md

paultancre:
Addind 3.10 in the requirement list
to resolve this : https://github.com/Labelbox/labelbox-python/issues/1261


paulnoirel (approved):
of for 3.10
We should remove 3.7 since it's not supported anymore https://devguide.python.org/versions/

PR #1286, opened 2023-11-17
Bytes array, multiple annotations per frame SN-84,SN-85

ovalle15:
Updates to use bytes array for masks (video, image), and add examples of multiple notations per frame (video)


examples/annotation_import/image.ipynb
```diff
@@ -430,23 +432,26 @@
     {
       "metadata": {},
       "source": [
-        "# Identifying what values in the numpy array correspond to the mask annotation\n",
-        "color = (255, 255, 255)\n",
-        "mask_data = lb_types.MaskData(url=\"https://storage.googleapis.com/labelbox-datasets/image_sample_data/raster_seg.png\")\n",
+        "### Raster Segmentation (Byte string array)\n",
+        "url = \"https://storage.googleapis.com/labelbox-datasets/image_sample_data/raster_seg.png\"\n",
+        "# url = \"https://storage.googleapis.com/labelbox-datasets/image_sample_data/color_mask.png\"\n",
```
apollonin, line 437: do we want to leave the comments here?

apollonin (approved):
This looks good to me, assuming we now "force" users to use im_bytes as a default input instead of url

ovalle15:
@apollonin  Manu suggested we use image bytes as default.

examples/annotation_import/image.ipynb
ovalle15: Does it make sense?  or not really and I should remove it?

apollonin: I mean looks like you have 
```
url = "https://storage.googleapis.com/labelbox-datasets/image_sample_data/color_mask.png"
# url = "https://storage.googleapis.com/labelbox-datasets/image_sample_data/color_mask.png"
```

^ these two lines in the notebook and I don't see why we need the seconds line. could be just a leftover?

ovalle15: AH - will remove that

ovalle15: Done

PR #1294, opened 2023-11-28
[SN-82] removed relationship reference in text and conversational notebook

Gabefire:
https://labelbox.atlassian.net/browse/SN-82


ovalle15:
Hi before making a PR, always run your notebooks locally to make sure they they work. Convo notebook is broken.

Gabefire:
i find the mistake I had annotations for the NDJSON replaced with python annotations by mistake I ran both text and conversational through locally

PR #1295, opened 2023-11-29
[QQC-2944] Remove deprecated project.datasets and dataset.projects methods

richardsun0713:


labelbox/schema/dataset.py
```diff
@@ -56,12 +55,6 @@ class Dataset(DbObject, Updateable, Deletable):
     row_count = Field.Int("row_count")
 
     # Relationships
-    projects = Relationship.ToMany(
```
vbrodsky, line 59: Can you do me a favor, so that I do not forget, add an item to a future release in CHANGELOG.

labelbox/schema/project.py
```diff
@@ -86,12 +85,6 @@ class Project(DbObject, Updateable, Deletable):
     media_type = Field.Enum(MediaType, "media_type", "allowedMediaType")
 
     # Relationships
-    datasets = Relationship.ToMany(
```
vbrodsky, line 89: same comment as in the above

labelbox/schema/dataset.py
vbrodsky: Also, just checking: would older sdk version still work, i.e. does the backend still support this api?

tests/integration/test_project_data_rows.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+import uuid
+
+from labelbox import Project, Dataset
+
+
+def test_batches(project: Project, dataset: Dataset, image_url):
```
vbrodsky, line 7: Appreciate you adding a test, but do we need it? We already have tests/integration/test_batch.py

labelbox/schema/dataset.py
richardsun0713: Backend has not changed, so it will return an empty list, just like before.

tests/integration/test_project_data_rows.py
richardsun0713: Will remove

ovalle15:
I checked the few notebooks that should run (basics.ipynb and datasets.ipynb) and looks good. 
Note that the other notebooks in this PR that are not actively maintained - so there are many more issues (unrelated to this PR) that need to be resolved for them to work as expected.

ovalle15 (approved):

vbrodsky (approved):

PR #1296, opened 2023-11-29
[AL-7380] Run foundry app

whistler:
Add support for running foundry apps


tests/integration/test_foundry.py
```diff
@@ -0,0 +1,92 @@
...
+    data_rows = lb.GlobalKeys([data_row.global_key])
+    task = foundry_client.run_app(
+        model_run_name=f"test-app-with-global-key-{random_str}",
+        data_rows=data_rows,
+        app_id=app.id)
+    task.wait_till_done()
+    assert task.status == 'COMPLETE'
+
+
+def test_run_foundry_app_with_non_existant_data_rows(foundry_client, data_row,
```
richardsun0713, line 83: nit: `existent`

labelbox/schema/foundry.py
```diff
@@ -0,0 +1,160 @@
...
+                    {field_names_str}
+                }}
+            }}
+        """
+        params = {"id": id}
+
+        try:
+            response = self.client.execute(query_str, params)
+        except exceptions.InvalidQueryError as e:
+            raise exceptions.ResourceNotFoundError(App, params)
```
richardsun0713, line 85: how come we need to handle this differently compared to the other schema entities?

tests/integration/test_foundry.py
whistler: Fixed

labelbox/schema/foundry.py
whistler: We do not want to expose App as a public DbObject yet. This is added just to support run_foundry_app.

richardsun0713 (approved):

labelbox/client.py
```diff
@@ -1786,3 +1789,17 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
             experimental=True)["project"]["batches"]["nodes"][0]
 
         return Entity.Batch(self, project_id, batch)
+
+    def run_foundry_app(self, model_run_name: str, data_rows: Union[DataRowIds,
```
vbrodsky, line 1793:
```suggestion
    def run_foundry_app(self, model_run_name: str, data_rows: DataRowIdentifiers,
```

labelbox/schema/foundry.py
```diff
@@ -0,0 +1,160 @@
+from datetime import datetime
```
vbrodsky, line 1: perhaps make foundry a folder and each class in a separate file?

labelbox/schema/foundry.py
```diff
@@ -0,0 +1,160 @@
...
+class App(_CamelCaseMixin, BaseModel):
+    id: Optional[str]
+    model_id: str
+    name: str
+    description: str
+    inference_params: Dict[str, Any]
+    class_to_schema_id: Dict[str, str]
+    created_by: Optional[str] = None
+
+    @classmethod
```
vbrodsky, line 19: I understand we have `_NoCoercionMixin` which you wrote yourself :). Curious why we are not using it here

labelbox/schema/foundry.py
```diff
@@ -0,0 +1,160 @@
...
+APP_FIELD_NAMES = list(App.schema()['properties'].keys())
+MODEL_FIELD_NAMES = list(Model.schema()['properties'].keys())
+
+
+class FoundryClient:
+
+    def __init__(self, client):
+        self.client = client
+
+    def _create_app(self, app: App) -> App:
```
vbrodsky, line 42: is this only used in test? do not see any other usage

labelbox/schema/foundry.py
```diff
@@ -0,0 +1,160 @@
...
+                    description
+                    inferenceParams
+                    classToSchemaId
+                }
+            }
+            """
+
+        params = app.dict(by_alias=True, exclude={"id"})
+
+        response = self.client.execute(query_str, params)
```
vbrodsky, line 67: and if there are errors?

labelbox/schema/foundry.py
```diff
@@ -0,0 +1,160 @@
...
+                    name
+                    modelId
+                    description
+                    inferenceParams
+                    classToSchemaId
+                }
+            }
+            """
+
+        params = app.dict(by_alias=True, exclude={"id"})
```
vbrodsky, line 65: any reason not to use `APP_FIELD_NAMES` like we do below?

labelbox/schema/foundry.py
```diff
@@ -0,0 +1,160 @@
...
+                    model {{
+                        {field_names_str}
+                    }}
+                }}
+            }}
+        """
+        params = {"id": id}
+
+        response = self.client.execute(query_str, params)
+        return Model(**response["modelFoundryModel"]["model"])
```
vbrodsky, line 114: and if there was an error?

labelbox/schema/foundry.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def run_app(self, model_run_name: str,
+                data_rows: Union[DataRowIds, GlobalKeys], app_id: str) -> Task:
+        app = self._get_app(app_id)
+        model = self._get_model(app.model_id)
+
+        if isinstance(data_rows, DataRowIds):
+            data_rows_query = {
+                "type": "data_row_id",
```
vbrodsky, line 123: suggestion: if you could change the type name, you could then use `IdType` and make this code more flexible. IdType(s) names are the same across API and ADV btw. Perhaps it would be good to stick to the same naming conventions?

vbrodsky (commented):
Lots of could stuff. My comments are mostly questions / non-blocking improvements.

The two main questions:
- error handling  / reporting for our mutation calls
- perhaps using the `IdType` class to make your code a little cleaner and more flexible

labelbox/client.py
whistler: I didn't use `DataRowIdentifiers` to make it clearer in docs and IDE code reference how this function can be used. i.e. if I see DataRowIdentifers in docs, I have to first find what it is, I will see" `Union[UniqueIds, GlobalKeys]`, its not clear here that `UniqueIds` means DataRowIds. Also, now I need to know about symbols two level-deep, first `DataRowIdentifers` and then one of `UniqueIds` or `GlobalKeys`. I agree with your suggestion for internal code, but IMO, for a public-facing method, the current annotation is clearer.

labelbox/schema/foundry.py
whistler: The `ResourceNotFoundError` expects this method.

labelbox/schema/foundry.py
whistler: good idea, I'll update.

labelbox/schema/foundry.py
whistler: This is the query format used in a large part of api codebase and not easy to change.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+# Version 3.57.0 (Not released yet)
```
vbrodsky, line 3: We are releasing 3.57.0 today. Please remove this entry

whistler: removed

vbrodsky (approved):

PR #1299, opened 2023-11-30
[AL-7381] Introduce send to annotate support

richardsun0713:


labelbox/client.py
```diff
@@ -1786,3 +1789,104 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
...
 
         return Entity.Batch(self, project_id, batch)
+
+    def send_to_annotate_from_catalog(
+            self,
+            batch_name: str,
+            data_rows: DataRowIdentifiers,
+            source_model_run_id: Optional[str],
+            source_project_id: Optional[str],
+            destination_project: str,
```
kkim-labelbox, line 1799: Should this be `destination_project_id`?

labelbox/client.py
```diff
@@ -1786,3 +1789,104 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
...
+            task_queue_id: Optional[str],
+            exclude_data_rows_in_project: bool = False,
```
kkim-labelbox, line 1801: An idea: what if we encapsulated all of the "configuration" type parameters in an object? If we decide to introduce a new configuration in the future as a required param, we'd have to rearrange the order of parameters in a breaking manner. Instead, what if we had a SendToAnnotateParams class as a TypeDict (or pydantic model) similar to export v2 params and filters? The benefit of this would be:
1. Less parameters in the function signature
2. Can reuse the same config object for both send_to_annotate functions

labelbox/client.py
```diff
@@ -1786,3 +1789,104 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
...
+            ontology_mapping: Dict[str, str] = None,
+            override_existing_annotations_rule:
+        ConflictResolutionStrategy = ConflictResolutionStrategy.KeepExisting,
+            priority: int = 5):
```
kkim-labelbox, line 1805: Let's add docstrings to both send_to_annotate methods with descriptions for what all of the parameters do

labelbox/schema/model_run.py
```diff
@@ -565,6 +568,86 @@ def export_v2(
...
+            priority: int = 5) -> Task:
+
+        mutation_str = """mutation SendToAnnotateFromMeaPyApi($input: SendToAnnotateFromMeaInput!) {
+                            sendToAnnotateFromMea(input: $input) {
+                              taskId
+                            }
+                          }
+        """
+
+        destination_task_queue = {
```
kkim-labelbox, line 590: Any way we can share some of the common input processing and query formatting code between the two methods?

tests/integration/test_send_to_annotate.py
```diff
@@ -0,0 +1,49 @@
...
+def test_send_to_annotate_include_annotations(
+        client, configured_batch_project_with_label, project_with_ontology):
+    [source_project, _, data_row, _] = configured_batch_project_with_label
+    [destination_project, _] = project_with_ontology
+
+    try:
+        queues = destination_project.task_queues()
+        initial_labeling_task = next(
+            q for q in queues if q.name == "Initial labeling task")
+        print(initial_labeling_task)
```
kkim-labelbox, line 16: Were the print statements left here on purpose?

labelbox/client.py
```diff
@@ -1786,3 +1789,104 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
...
+        annotations_input = {
+            "projectId":
+                source_project_id,
+            "featureSchemaIdsMapping":
+                ontology_mapping if ontology_mapping else {},
+        } if source_project_id else None
+
+        predictions_input = {
+            "featureSchemaIdsMapping":
+                ontology_mapping if ontology_mapping else {},
```
kozikkamil, line 1830: nit: did we also consider having two separate mappings, one for annotations and one for predictions? Seems like it could be a little bit cleaner

tests/integration/test_send_to_annotate.py
richardsun0713: No, I need to come back and clean them up. I wanted to first make sure the tests were passing on all python versions

labelbox/schema/model_run.py
richardsun0713: Possibly - we just need to find the right place to put these utility methods

labelbox/client.py
```diff
@@ -1786,3 +1789,104 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
...
+            "modelRunId":
+                source_model_run_id,
+            "minConfidence":
+                0,
+            "maxConfidence":
+                1
+        } if source_model_run_id else None
+
+        if isinstance(data_rows, UniqueIds):
```
vbrodsky, line 1839: I believe I've come across this code snippet in a previous PR for Model Foundry. Can we create a method and reuse it in both places?

labelbox/client.py
```diff
@@ -1786,3 +1790,156 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
             experimental=True)["project"]["batches"]["nodes"][0]
 
         return Entity.Batch(self, project_id, batch)
+
+    def send_to_annotate_from_catalog(self, destination_project_id: str,
+                                      task_queue_id: Optional[str],
+                                      batch_name: str,
+                                      data_rows: DataRowIdentifiers,
```
whistler, line 1797: Something we discussed earlier with @karenkyang that it may not be clear to the user that `UniqueIds` is `DataRowIds`. For running the foundry app, I have annotated data_rows as: `Union[DataRowIds, GlobalKeys]`. https://github.com/Labelbox/labelbox-python/pull/1296/files#diff-3755e76da3ad93f8739b50227489ef9f27db2432abd67ca08999f57293ca2794R1793

Lets decide on how we want to annotate data_rows and make it consistent.

labelbox/client.py
```diff
@@ -1,51 +1,53 @@
...
 import mimetypes
 import os
-import time
+import random
 import sys
+import time
 import urllib.parse
+from collections import defaultdict
+from datetime import datetime, timezone
+from typing import Any, List, Dict, Union, Optional
```
kkim-labelbox, line 12: nit: how come the ordering of the imports changed so much?

labelbox/schema/send_to_annotate_params.py
```diff
@@ -0,0 +1,59 @@
...
+    from typing_extensions import TypedDict
+
+
+class SendToAnnotateFromCatalogParams(TypedDict):
+    """
+    Extra parameters for sending data rows to a project through catalog. At least one of source_model_run_id or
+    source_project_id must be provided.
+
+    :param source_model_run_id: Optional[str] - The model run to use for predictions. Defaults to None.
+    :param model_run_ontology_mapping: Optional[Dict[str, str]] - A mapping of ontology ids to ontology ids. Defaults to
```
kkim-labelbox, line 19: Should we rename these to
`predictions_ontology_mapping` and `annotations_ontology_mapping`, to stay consistent with `predictions` vs `annotations` terminology?

kkim-labelbox: Also I think `A mapping of feature schema ids to feature schema ids` is more accurate

labelbox/schema/model_run.py
```diff
@@ -565,6 +569,75 @@ def export_v2(
...
+                        predictions_input,
+                    "conflictLabelsResolutionStrategy":
+                        override_existing_annotations_rule,
+                    "searchQuery": [data_rows_query],
+                    "sourceModelRunId":
+                        self.uid
+                }
+            })['sendToAnnotateFromMea']
+
+        return Entity.Task.get_task(self.client, res['taskId'])
```
kkim-labelbox, line 639: Do Task.result and Task.errors work, or any tweaks needed there to surface results/errors?

labelbox/client.py
```diff
@@ -1790,6 +1792,163 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
...
+                        None,
+                    "limit":
+                        None
+                }
+            })['sendToAnnotateFromCatalog']
+
+        return Entity.Task.get_task(self, res['taskId'])
+
+    @staticmethod
+    def build_annotations_input(project_ontology_mapping: Optional[Dict[str,
```
kkim-labelbox, line 1887: nit: should we consider a `send_to_annotate_params.py` file to put all of the static methods?

labelbox/schema/model_run.py
richardsun0713: `result` and `result_url` are not well supported with bulk-ops tasks at the moment. I will add support for surfacing errors though

labelbox/client.py
```diff
@@ -1790,6 +1793,141 @@ def get_batch(self, project_id: str, batch_id: str) -> Entity.Batch:
...
+                        None,
+                    "limit":
+                        None
+                }
+            })['sendToAnnotateFromCatalog']
+
+        return Entity.Task.get_task(self, res['taskId'])
+
+    @staticmethod
+    def build_catalog_query(data_rows: Union[DataRowIds, GlobalKeys]):
```
kkim-labelbox, line 1903: Let's move this also to `send_to_annotate_params.py` and rename it to `build_data_rows_query`?

kkim-labelbox (commented):
Left 1 more comment

labelbox/client.py
richardsun0713: It's used by `labelbox.schema.foundry.foundry_client.FoundryClient.run_app` as well which is why I left it here

kkim-labelbox (approved):

PR #1304, opened 2023-12-01
Fix doc installation packages

vbrodsky:


vbrodsky:
Doc update, no need to wait for tests to finish

PR #1313, opened 2023-12-03
Adjusted Task.wait_until_done to include the check frequency as an argument

Gabefire:
I am running into an issue on hitting my api rate limit when using the task wait_until_done method. I'm messing with multi threading on the dataset.create_data_rows method to see if I can get datarows imported through the SDK faster and more reliable. What I found is spacing out each of my requests to wait for the previous datarows to be completed leads to less errors and more consistent results. The wait until done method works for this purpose, but when I try to use more then a few threads the check frequency being set to every two seconds hits the api rate limit. Having the check_frequency set up as an argument defaulted to two seconds (previous value) then passing ten seconds fixes my issue and still maintains the original purpose of this method. I think it is such a minor change which does not change the actual functionality of the method it could be useful in production for others who run into my situation especially customers in the free tier/ pay as you go who have a lower api rate limit.


Gabefire:
Hi @vbrodsky I can do that! I just pushed an addition that would raise a valueError if you passed a check frequency of less than 2 seconds. Let me know if that works I also updated the argument string to show the minimal value.

vbrodsky (approved):
approved for sdk code (not notebooks)

Gabefire:
Ignore my recent history tried to remove SDK commits from branch was slightly behind but it looks like you approved my commits

PR #1319, opened 2023-12-07
SN-87 (Include additional annotations for LLM notebook)

ovalle15:
1. Included additional supported annotations 
2. Changed the folder structure
3. Improve formatting 
4. added annotation types

### DIFF
--- a/examples/annotation_import/conversational_LLM_data_generation.ipynb
+++ b/examples/annotation_import/conversational_LLM_data_generation.ipynb
@@ -213,10 +213,10 @@
       "metadata": {},
       "source": [
         "# Enter the project id\n",
-        "project_id = \"clpvnyd2303pl07xu5rtofz7p\"\n",
+        "project_id = \"\"\n",
         "\n",
         "# Select one of the global keys from the data rows generated\n",
-        "global_key = \"clpvnydjb03pp07xuhyqi9bgk\""
+        "global_key = \"\""
       ],
       "cell_type": "code",
       "outputs": [],


Gabefire (changes requested):
Hey the conversational_LLM_predicitions .ipyn is messed up for that json URL you have to delete out the ModelOutput key when you go to upload the file to a dataset. It honestly needs to give a warning but if you don't in the UI it will show a "can not load asset" Something like this work:
```
res = requests.get("https://storage.googleapis.com/labelbox-datasets/conversational-sample-data/pairwise_shopping_2.json")
my_dict = res.json()
del my_dict["modelOutputs"]
json_file = json.dumps(my_dict)

with open("sample.json", "w") as outfile:
    outfile.write(json_file)

global_key = "pairwise_shooping_asset2"
convo_data =  {
        "row_data": "./sample.json",
        "global_key": global_key
}
# Create a dataset
dataset = client.create_dataset(name="pairwise_prediction_demo")
# Create a datarows
task = dataset.create_data_rows([convo_data])
print("Errors:",task.errors)
print("Failed data rows:", task.failed_data_rows)
```
I think you would use the model outputs key for the predictions list

ovalle15:
Hi @Gabefire , I do not see any errors. Can you please point where in the code you are getting an error?

Gabefire:
The SDK actually does not get an error (even though it should). if you go into the UI you will see this for the data row 
image

ovalle15:
@Gabefire that is an issue with LB, can you try to reprocess please

Gabefire:
Scratch that idea it is working now that I checked on it later, and it finally processed. (that might be important to note though because I had to that happen before when I was uploading stuff to MLSE and I had an extra key included) Ill push approve now

Gabefire (changes requested):
@ovalle15 Hey on conversational_LLM_data-generation! I got it to run but can you remove your projet ID and Datarow global ID I think you forgot to remove them from when you were testing. Also on the checklist in the JSON Ontology you went response_a response_c and response_d. I know it is minor but do you think response_a and then response_b would be better?

PR #1322, opened 2023-12-09
[AL-7548] Improve test coverage for send to annotate

richardsun0713:


tests/integration/annotation_import/conftest.py
```diff
@@ -579,6 +579,21 @@ def configured_project(client, initial_dataset, ontology, rand_gen, image_url):
...
+def project_with_ontology(client, configured_project, ontology, rand_gen):
+    project = client.create_project(name=rand_gen(str),
+                                    queue_mode=QueueMode.Batch,
+                                    media_type=MediaType.Image)
+    editor = list(
+        client.get_labeling_frontends(
+            where=LabelingFrontend.name == "editor"))[0]
+    project.setup(editor, ontology)
+
+    yield [project, ontology]
```
vbrodsky, line 592:
```suggestion
    yield project, ontology # do not think need to wrap it in an array, i.e leave it as a tuple
```

tests/integration/annotation_import/test_send_to_annotate_mea.py
```diff
@@ -13,15 +14,38 @@ def test_send_to_annotate_from_model(client, configured_project,
...
     task = model_run.send_to_annotate_from_model(
         destination_project_id=destination_project.uid,
         batch_name="batch",
         data_rows=UniqueIds(data_row_ids),
-        task_queue_id=initial_labeling_task.uid,
-        params={})
+        task_queue_id=initial_review_task.uid,
+        params={
+            "predictions_ontology_mapping":
+                ontology_mapping,
```
vbrodsky, line 45: so this is basically {id1: id1, id2: id2... } right? Kind of weird mapping to itself. Should be test a case when we have like {id1: id2, id3: id4} ?

tests/integration/annotation_import/test_send_to_annotate_mea.py
```diff
@@ -32,3 +56,7 @@ def test_send_to_annotate_from_model(client, configured_project,
     destination_data_rows = list(destination_batches[0].export_data_rows())
     assert len(destination_data_rows) == len(data_row_ids)
     assert all([dr.uid in data_row_ids for dr in destination_data_rows])
+
+    # Since data rows were added to a review queue, predictions should be imported into the project as labels
+    destination_project_labels = (list(destination_project.labels()))
+    assert len(destination_project_labels) == len(data_row_ids)
```
vbrodsky, line 62: Possible also to compare some label attributes? Also to make sure len(...) is not 0?

tests/integration/annotation_import/test_send_to_annotate_mea.py
richardsun0713: This is one of the valid use-cases. I can add additional tests as a follow-up

vbrodsky (approved):

PR #1339, opened 2023-12-18
[SN-92] create a foundry + send to annotate workflow example

Gabefire:
Finished notebook on foundry SDK basic methods. I made a new folder on this since no folder seemed to match foundry. Also, not sure if you want me to do other examples besides just images i.e. make a notebook on text models


examples/foundry/image_classification.ipynb
```diff
@@ -0,0 +1,352 @@
+{
```
ovalle15, line 1: <p>Could we add some brief text that explains what is the point of this step ? </p><p>Something like: This step is meant to generate annotations that can later be reused as pre-labels in a project</p><p>and link docs  perhaps ? https://docs.labelbox.com/docs/model-runs</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/1339/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

ovalle15, line 1: <p>I think we need to explain this a bit better:</p><p>Something like: Mapping a model's ontology to a project's ontology is currently not supported through the SDK, however, to showcase how to send foundry predictions to a project, we are going to generate the mapping of the foundry app ontology to the project ontology through the UI.</p><p> </p><p> </p><ol><li>Navigate to your dataset you created for your model run</li><li>Select&nbsp;<strong>Select all</strong>&nbsp;in the top right corner</li><li>Select&nbsp;<strong>Manage selection</strong>&nbsp;&gt;&nbsp;<strong>Send to Annotate</strong></li><li>Specify the project we created from the project dropdown menu</li><li>Selecting a workflow step is not required since we are not sending annotations to a project using this UI modal.</li><li>Mark&nbsp;<strong>Include model predictions</strong>&nbsp;then scroll down and select&nbsp;<strong>Map</strong></li><li>Select the incoming ontology and matching ontology feature for both Computer and Phone</li><li>Once both features are mapped press the&nbsp;<strong>Copy ontology mapping as JSON</strong>&nbsp;in the top right corner</li><li>Do not save this configuration, since we are not sending predictions to a project using this UI modal. We will be sending predictions in the following steps using the SDK. </li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/1339/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

ovalle15, line 1: <p>Should the title use "annotations" or "predictions" ?</p><ol><li>Could we format parameters like this --&gt; <code>predictions_ontology_mapping</code> please?</li><li><code>predictions_ontology_mapping </code> for this one, I think we need to clarify, something like, "A dictionary containing the mapping of the model's ontology feature schema ids to the project's ontology feature schema ids.</li><li><code>override_existing_annotations_rule</code> for this one we are missing one resolution</li><li class="ql-indent-1"><code>OverrideWithAnnotations = "OVERRIDE_WITH_ANNOTATIONS" </code> Do we know what this is ? and how to use it?</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/1339/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

aovallelb:
I think we should only include the classification example - I think the object detection and text entities is just redundant

Gabefire:
@aovallelb I think maybe the same thing but the object detection one instead. That one shows off the foundry. It picks up 20-30 objects for the cost of being slightly more complicated, which might be a good thing since you have to map your ontology. It also uses an image we use across multiple notebooks, which is a nice touch. But do you think it should be in a new or existing folder? Maybe extras?

aovallelb:
> @aovallelb I think maybe the same thing but the object detection one instead. That one shows off the foundry. It picks up 20-30 objects for the cost of being slightly more complicated, which might be a good thing since you have to map your ontology. It also uses an image we use across multiple notebooks, which is a nice touch. But do you think it should be in a new or existing folder? Maybe extras?

@Gabefire sounds good; I ll review that one. I think it is ok to leave that single file in the foundry folder for now, since, hopefully, there will be more methodologies in the near future that we could create notebooks for, wdyt ?

aovallelb:
@Gabefire feel free to incorporate the changes I suggested on the object classification file only.

examples/foundry/image_classification.ipynb
Gabefire: <p>Hey! Override with annoations is meant for when you want to override the annotations on a project from one project to another. That is why I did not include since it would not be application for that method. I believe it does the exact same as OverrideWithPredictions when it is with a model. Though it lets me pass in either or and gives the same result. Im sending in some commits now from my changes so far</p>

Gabefire:
@aovallelb Hey I made the changes for the conflictresolution I believe the Overridewithannotations is for if you go from one project to another with the other send_to_annotate method. But when I used it, it would work but do the same thing as overwithpredictions. I have it included in the notebook. Also, just a heads up, there is a bug with the run_foundry_app method. You have to have a description in your foundry app. I actually fixed this on another. PR, and they approved it, but when you go to test, you will likely need to include a description

examples/foundry/object_detection.ipynb
```diff
@@ -0,0 +1,376 @@
...
+        }
+      ],
+      "execution_count": null
+    },
+    {
+      "metadata": {},
+      "source": [
+        "## Step 6: Map ontology through the UI\n",
+        "\n",
+        "Something like: Mapping a model's ontology to a project's ontology is currently not supported through the SDK, however, to showcase how to send foundry predictions to a project, we are going to generate the mapping of the foundry app ontology to the project ontology through the UI.\n",
```
ovalle15, line 262: Could you remove "Something like:" -  Once removed this all looks good and you can merge!

ovalle15 (approved):

PR #1343, opened 2023-12-21
[SN-100] Update notebook for Project move_data_rows_to_task_queue

Gabefire:
The task wanted this added to the basic/data_row.ipynb notebook but this is a project method. I put it in the project notebook, and it seems to fit better. I had to modify it to import labels since it does not work unless data rows are labeled but this showed off a few of our other project methods in the process. Let me know if this will work!

### DIFF
--- a/examples/basics/projects.ipynb
+++ b/examples/basics/projects.ipynb
@@ -221,7 +221,9 @@
         "     name=\"label_import_job\"+str(uuid.uuid4()),\n",
         "     labels=labels)\n",
         "\n",
-        "upload_job.wait_until_done()"
+        "upload_job.wait_until_done()\n",
+        "\n",
+        "print(f\"Errors: {upload_job.errors}\")"
       ],
       "cell_type": "code",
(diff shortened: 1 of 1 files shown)


examples/basics/projects.ipynb
```diff
@@ -48,7 +48,7 @@
     {
```
ovalle15, line 48: <p>Can we add some text here about why are we creating ontology and labels, and re-format some of the code blocks like we discussed?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Labelbox/labelbox-python/pull/1343/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

Gabefire: Hey! I made these changes. Let me know what you think.

ovalle15 (approved):

PR #1347, opened 2023-12-22
[SDK-503] Prevent users from uploading video annotations over the API limit (5000).

sfendell-labelbox:


pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
-addopts = -s -vv --reruns 5 --reruns-delay 10 --durations=20
+addopts = -s -vv --reruns-delay 10 --durations=20
```
vbrodsky, line 2: coincidental?

tests/integration/annotation_import/test_bulk_import_request.py
```diff
@@ -370,3 +370,69 @@ def test_nested_video_object_annotations(client,
...
+                                               ))
+                     ])
+
+    configured_project_with_one_data_row.update(media_type=MediaType.Video)
+    configured_project_with_one_data_row.create_batch(
+        rand_gen(str),
+        data_row_uids,  # sample of data row objects
+        5  # priority between 1(Highest) - 5(lowest)
+    )
+    labels = [create_label() for _ in range(5000)]
```
vbrodsky, line 398: Are you considering running a full upload of 5,000 annotations for SDK integration testing? This could significantly impact performance and might even result in flakiness. I recall we previously reduced some tests due to similar concerns. Would you consider creating a method or perhaps patching the LABEL_LIMIT to a smaller number, like 2, specifically for integration tests to mitigate these issues?

vbrodsky (changes requested):

vbrodsky (approved):

PR #1349, opened 2024-01-02
[SDK-0] Fix class annotations

mnoszczak:


labelbox/data/serialization/ndjson/converter.py
```diff
@@ -40,8 +42,34 @@ def serialize(
...
+                    source_uuid = annotation.value.source._uuid
+                    target_uuid = annotation.value.target._uuid
+                    new_source_uuid = str(uuid.uuid4())
+                    new_target_uuid = str(uuid.uuid4())
+                    annotation_uuid_lookup[source_uuid] = new_source_uuid
+                    annotation_uuid_lookup[target_uuid] = new_target_uuid
+                    annotation.value.source._uuid = new_source_uuid
+                    annotation.value.target._uuid = new_target_uuid
+
+            for annotation in label.annotations:
```
sfendell-labelbox, line 64: A small optimization: Can you filter out these annotations in the first pass into a separate array, then just iterate over that array? Something like:

```
annotations_to_fix = []
for annotation in label.annotations:
  if isinstance:
      ....this stuff...
  else:
    annotations_to_fix.append(annotation)

for annotation in annotation_to_fix:
  ....
```
It avoids repeating the condition logic and avoids iterating over the entire array again for a tiny cost in memory.

labelbox/data/serialization/ndjson/converter.py
```diff
@@ -40,8 +42,45 @@ def serialize(
         Returns:
             A generator for accessing the ndjson representation of the data
         """
+        used_annotation_uuids = set()
+        for label in labels:
+            annotation_uuid_lookup = {}
+            # UUIDs are private properties used to enhance UX when defining relationships.
```
vbrodsky, line 48: wonderful clear comments

labelbox/data/serialization/ndjson/converter.py
```diff
@@ -40,8 +42,45 @@ def serialize(
...
+                    source_uuid = annotation.value.source._uuid
+                    target_uuid = annotation.value.target._uuid
+
+                    if (len(
+                            used_annotation_uuids.intersection(
+                                {source_uuid, target_uuid})) > 0):
+                        new_source_uuid = uuid.uuid4()
+                        new_target_uuid = uuid.uuid4()
+
+                        annotation_uuid_lookup[source_uuid] = new_source_uuid
```
vbrodsky, line 64: perhaps, rename `annotation_uuid_lookup` to `annotation_uuid_to_generated_uuid_lookup` ? Not sure, just trying to give this important variable even more clarity...

labelbox/data/serialization/ndjson/converter.py
```diff
@@ -40,8 +42,45 @@ def serialize(
...
+                    annotation._uuid = annotation_uuid_lookup.get(
+                        annotation._uuid, None)
+
         for example in NDLabel.from_common(labels):
-            res = example.dict(by_alias=True)
+            annotation_uuid = getattr(example, "uuid", None)
+
+            annotation_uuid_is_none = annotation_uuid == "None"
+            if not annotation_uuid_is_none:
+                used_annotation_uuids.add(annotation_uuid)
```
vbrodsky, line 80: I must be missing something here regarding `used_annotation_uuids`. We update them here, line 80. However we check them for intersection BEFORE, at line 59... wouldn't it be always empty there?

vbrodsky (approved):

PR #1355, opened 2024-01-05
Update PyPi Publish action to use trusted publisher setup

vbrodsky:
Username / password not returns http 403


abacchilb (approved):
On zoom w V- looks good

PR #1358, opened 2024-01-08
Pno/qqc 3095 project get overview

paulnoirel:
`project.get_overview()` provides a shortcut for users to retrieve the number of data rows and their current status. It also provides the number of issues, if requested, like in the _Overview tab_ of a project.

Examples:

```project.get_overview()```
Result:

```
{'labeled': 4124,
 'in_review': {'Initial review task': 0, 'PolyPerception review': 0, 'all': 0},
 'in_rework': 0,
 'done': 4124,
 'to_label': 0,
 'all_data_rows': 4124}
```

```project.get_overview(with_review=True)```
Result:

```
{'labeled': 4124,
 'in_review': {'Initial review task': 0, 'PolyPerception review': 0, 'all': 0},
 'in_rework': 0,
 'done': 4124,
 'issues': 0,
 'to_label': 0,
 'all_data_rows': 4124}
```

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -1655,7 +1655,24 @@ class Project(DbObject, Updateable, Deletable):
         return response["queryAllDataRowsHaveBeenProcessed"][
             "allDataRowsHaveBeenProcessed"]
 
-    def get_overview(self, with_issues=False):
+    def get_overview(self, with_issues: bool = False) -> Dict:
+        """Return 
+
+                Args:
+                    with_issues: (optional) boolean to include issues in the overview
+                Returns:
+                    Dictionary with the following structure:
+                    {
+                        'Labeled': (int) Number of labeled data rows,
+                        'InReview': (dict) Name of the in review queue and associated number of data rows
+                        'InRework': (int) Number of data rows in rework,
+                        'Done': (int) Number of completed data rows,
+                        'ToLabel': (int) Number of data rows to label,
+                        'AllDataRows': (int) Total number of data rows,
+                        'Issues': (optional) Number of issues
+                    }
+                """
+
         query = """query GetWorkstreamStateCounts($projectId: ID!) {
             project(where: { id: $projectId }) {      
             workstreamStateCounts {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: labelbox/schema/project.py [hunk 1]
```diff
@@ -1655,6 +1655,73 @@ def __check_data_rows_have_been_processed(
         return response["queryAllDataRowsHaveBeenProcessed"][
             "allDataRowsHaveBeenProcessed"]
 
+    def get_overview(self, with_issues: bool = False) -> Dict:
```
vbrodsky, line 1658: the return should be an object of a class, a python `@dataclass`or similar... since you are doing fair number of conversions, a pydantic model might be a good match.

labelbox/schema/project.py
```diff
@@ -1655,6 +1655,73 @@ def __check_data_rows_have_been_processed(
...
+            issues {
+                totalCount
+            }
+            completedDataRowCount
+            }
+        }
+        """
+
+        # Must use experimental to access "issues"
+        result = self.client.execute(query, {"projectId": self.uid},
```
vbrodsky, line 1697: why is `experimental=True`?

labelbox/schema/project.py
```diff
@@ -1655,6 +1655,73 @@ def __check_data_rows_have_been_processed(
...
+                                     experimental=True)["project"]
+
+        overview = {
+            utils.snake_case(st["state"]): st["count"]
+            for st in result["workstreamStateCounts"]
```
vbrodsky, line 1702: Is `workstreamStateCounts` always guaranteed to be there? The same question for other keys

tests/integration/test_task_queue.py
```diff
@@ -11,6 +11,9 @@ def test_get_task_queue(project: Project):
         tq for tq in task_queues if tq.queue_type == "MANUAL_REVIEW_QUEUE")
     assert review_queue
 
+    overview = project.get_overview(with_issues=True)
+    assert len(overview.keys()) == 7
+
```
vbrodsky, line 16: we should be able to validate all attribute values too

vbrodsky (changes requested):

PR #1363, opened 2024-01-17
Syncing with original labelbox-python

mihhail-m:


labelbox/client.py
```diff
@@ -117,6 +117,8 @@ def __init__(self,
             'X-Python-Version': f"{python_version_info()}",
         }
         self._data_row_metadata_ontology = None
+        self._session = requests.Session()
```
vbrodsky, line 120: why not do create `Session` in the constructor?

labelbox/client.py
```diff
@@ -184,20 +186,15 @@ def convert_value(value):
...
-                'headers': self.headers,
-                'timeout': timeout
-            }
-            if files:
-                request.update({'files': files})
-                request['headers'] = {
-                    'Authorization': self.headers['Authorization']
-                }
+        request = requests.Request(method="POST", url=endpoint, data=data)
+        prepare_req = request.prepare()
```
vbrodsky, line 190: what is the advantage of using a prepared request here as compared to how it was done before?

vbrodsky: Also I have found this comment [here](https://requests.readthedocs.io/en/latest/user/advanced/#prepared-requests):
>However, the above code will lose some of the advantages of having a Requests [Session](https://requests.readthedocs.io/en/latest/api/#requests.Session) object. In particular, [Session](https://requests.readthedocs.io/en/latest/api/#requests.Session)-level state such as cookies will not get applied to your request. To get a [PreparedRequest](https://requests.readthedocs.io/en/latest/api/#requests.PreparedRequest) with that state applied, replace the call to [Request.prepare()](https://requests.readthedocs.io/en/latest/api/#requests.Request.prepare) with a call to [Session.prepare_request()](https://requests.readthedocs.io/en/latest/api/#requests.Session.prepare_request), like this:

labelbox/client.py
```diff
@@ -184,20 +186,15 @@ def convert_value(value):
...
+        prepare_req.headers = self.headers
 
-            response = requests.post(**request)
+        if files:
+            prepare_req.headers["files"] = files
+
+        try:
+            response = self._session.send(prepare_req, timeout=timeout, stream=True)
```
vbrodsky, line 197: curious why `stream=true` here

vbrodsky (commented):
It's great to see that we are adopting the approach of utilizing HTTP connection pooling. This method should theoretically enhance our system's efficiency. Could you provide any data or metrics regarding the performance improvements we can expect from this implementation?

mihhail-m:
@vbrodsky my bad this code wasn't meant for `labelbox-python`. I was trying to sync my fork[1] that i am using in order to run the automated tests.

> why not do create Session in the constructor?

Not sure I understand. Isn't `__init__` method is the constructor in the python?

Regarding metrics. This is the only metrics[2] I got when I was trying to implement the `Session` object approach. Basically, this approach helped me resolving issue I was having when running automated tests in the pipeline and hitting some weird api exceptions errors[3]. I haven't seen this error since. 

> what is the advantage of using a prepared request here as compared to how it was done before?

Not exactly sure. But the idea is that it allows to add some options before sending an actual request. Make it more flexible in the long run, allegedly.

> curious why stream=true here

There is an issue with the the imports. For some reasons imports serialized differently with `Session` object i think, i thought that would help. I believe, it didn't. Haven't had much time to look into that since.

1. https://github.com/mihhail-m/labelbox-python-mm-fork
2. https://docs.google.com/spreadsheets/d/1pgmfLR6HvtARklMvL0p47QM-SjAb8GqhrKvDcD6fnHw/edit?usp=sharing
3. https://labelbox.slack.com/archives/C03PTTCEFUJ/p1700481117551419

PR #1379, opened 2024-01-23
[AL-7665] Renamed GQL mutation name

dubininsergey:


labelbox/schema/foundry/foundry_client.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, client):
     def _create_app(self, app: App) -> App:
         field_names_str = "\n".join(APP_FIELD_NAMES)
         query_str = f"""
-            mutation CreateDataRowAttachmentPyApi(
+            mutation CreateFoundryApp(
```
kkim-labelbox, line 17: Should add `PyApi` to the end so we can track it as an SDK invoked GQL request

`CreateFoundryAppPyApi`

dubininsergey: thanks for the tip, Kevin!

PR #1380, opened 2024-01-23
[QQC-2962] Add a method to get label count for project

mkozik-lb:
Adds a new method `get_label_count` to the Project class, which allows to effectively find how many labels are in the project. Our customers would do this before:

```python
labels_from_labelbox = sum([1 for _ in target_project.labels()])
```

### DIFF
--- a/labelbox/schema/project.py
+++ b/labelbox/schema/project.py
@@ -1149,6 +1149,20 @@ class Project(DbObject, Updateable, Deletable):
 
         return mode
 
+    def get_label_count(self) -> int:
+        """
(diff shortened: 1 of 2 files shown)


vbrodsky:
> Adds a new method `get_label_count` to the Project class, which allows to effectively find how many labels are in the project. Our customers would do this before:
> 
> ```python
> labels_from_labelbox = sum([1 for _ in target_project.labels()])
> ```
I do not find the `labels()` method

vbrodsky (approved):

PR #1387, opened 2024-01-25
[SDK-546] Must pass model run id in order for model slices to work correctly

vbrodsky:
I want to bring to attention the recent updates made to the SDK's ModelSlice methods, specifically `get_data_row_identifiers` and `get_data_row_ids`. We've modified these methods to now require a model run id as a parameter. It's important to note that these changes are **NOT backward-compatible**.

Prior to the last SDK release (version 3.61.0), these methods were non-functional. After the release of 3.61.0, they have not been operating correctly. This update, involving the inclusion of a model run id, is essentially a fix to ensure these methods work as intended.

While this change does break backward compatibility, it is necessary to correct the functionality that was previously either broken or incorrect.

Required API PR: https://github.com/Labelbox/intelligence/pull/19099

### DIFF
--- a/labelbox/schema/slice.py
+++ b/labelbox/schema/slice.py
@@ -239,9 +239,10 @@ class ModelSlice(Slice):
     @classmethod
     def query_str(cls):
         query_str = """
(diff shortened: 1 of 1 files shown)


sfendell-labelbox:
Change is LGTM, but do we have/need test coverage for this?

kopreschko (approved):
lgtm

PR #1399, opened 2024-01-31
[AL-7738] Support custom metrics for predictions

kozikkamil:
Adds "customMetrics" field definition for ndjson and "custom_metrics" for class object for annotations.
It's meant to be used at the prediction level in when adding predictions to MEA.

Should not be merged without respective changes deployed to rest of tech stack.

### DIFF
--- a/labelbox/data/mixins.py
+++ b/labelbox/data/mixins.py
@@ -55,8 +55,8 @@ class CustomMetricsMixin(BaseModel):
     def dict(self, *args, **kwargs):
         res = super().dict(*args, **kwargs)
 
-        if "customMetrics" in res and res["customMetrics"] is None:
-            res.pop("customMetrics")
+        if "custom_metrics" in res and res["custom_metrics"] is None:
+            res.pop("custom_metrics")
 
         return res
 
--- a/labelbox/data/mixins.py
+++ b/labelbox/data/mixins.py
@@ -1,32 +1,70 @@
-from typing import Optional
+from typing import Optional, List
 
 from pydantic import BaseModel, validator
 
-from labelbox.exceptions import ConfidenceNotSupportedException
+from labelbox.exceptions import ConfidenceNotSupportedException, CustomMetricsNotSupportedException
 
 
 class ConfidenceMixin(BaseModel):
     confidence: Optional[float] = None
 
-    @validator('confidence')
+    @validator("confidence")
     def confidence_valid_float(cls, value):
         if value is None:
             return value
         if not isinstance(value, (int, float)) or not 0 <= value <= 1:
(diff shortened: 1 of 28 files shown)


labelbox/data/mixins.py
```diff
@@ -1,32 +1,70 @@
...
+                "Confidence is not supported for this annotation type yet"
+            )
+        return super().__new__(cls)
+
+
+class CustomMetric(BaseModel):
+    name: str
+    value: float
+
+    @validator("name")
```
vbrodsky, line 39: yep this is a good catch... pydantic will coerce to str any other possible type... I know in pydantic2 there is a config to make it more strict.  I did not see anything like this in pydantic 1. Good catch

labelbox/data/mixins.py
```diff
@@ -1,32 +1,70 @@
...
+        return value
+
+
+class CustomMetricsMixin(BaseModel):
+    custom_metrics: Optional[List[CustomMetric]] = None
+
+    def dict(self, *args, **kwargs):
+        res = super().dict(*args, **kwargs)
+
+        if "customMetrics" in res and res["customMetrics"] is None:
```
vbrodsky, line 58: so, to be clear, we do not want to serlaize custom metrics?

tests/integration/annotation_import/test_mea_prediction_import.py
```diff
@@ -185,6 +186,29 @@ def test_create_from_local_file(tmp_path, model_run_with_data_rows,
     assert annotation_import.state == AnnotationImportState.FINISHED
     annotation_import_test_helpers.download_and_assert_status(
         annotation_import.status_file_url)
+    
+
+def test_predictions_with_custom_metrics(model_run,
```
vbrodsky, line 191: is this an additional test or is it something specific to the metrics use case?

vbrodsky (commented):
Solid work, can approve when the time comes, left a couple of question for my own curiosity

labelbox/data/mixins.py
kozikkamil: Could you elaborate a bit please? This particular line is just making sure we don't pass None, but rather remove the key entirely in that case

tests/integration/annotation_import/test_mea_prediction_import.py
kozikkamil: It checks whether a mea import with custom metrics on prediction level succeeds

labelbox/data/mixins.py
vbrodsky: Sorry IS NONE... missed it

vbrodsky (approved):

PR #1402, opened 2024-02-01
Fixed label py to fix issues with frame annotations

Gabefire:
Framed-based annotations for classification and radio on videos are currently not serializing correctly when all the answers for each frame are broken. The underlying reason seems to be that we use code from non-frame-based annotations to generate our pydantic labels. For frame-based, you have to consider all answers spread out for the entire video timeline before you can accurately make labels. With previous implementation, it is making the label one at a time. This fix makes the frame-based annotation for each answer on a label. There still needs to be a fix to group those answers into their feature categories. I attempted to do this, but it was getting very complicated. That fix will likely require a lot more work so I made this a draft pull and putting a ticket in.

### DIFF
--- a/labelbox/data/serialization/ndjson/label.py
+++ b/labelbox/data/serialization/ndjson/label.py
@@ -190,8 +190,11 @@ class NDLabel(BaseModel):
         ])
         if len(sorted_frame_segment_indices) == 0:
             # Group segment by consecutive frames, since `segment_index` is not present
-            return cls._get_consecutive_frames(
-                sorted([annotation.frame for annotation in annotation_group]))
+            return [
+                cls._get_consecutive_frames(
+                    sorted(
+                        [annotation.frame for annotation in annotation_group]))
+            ]
         elif len(sorted_frame_segment_indices) == len(annotation_group):
             # Group segment by segment_index
             last_segment_id = 0
@@ -204,8 +207,10 @@ class NDLabel(BaseModel):
                 segment_groups[segment_index].append(frame)
                 last_segment_id = segment_index
             frame_ranges = []
+
+            # Group consecutive frames of segment_index
             for group in segment_groups.values():
-                frame_ranges.append((group[0], group[-1]))
+                frame_ranges.append(cls._get_consecutive_frames(sorted(group)))
             return frame_ranges
         else:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


vbrodsky:
@Gabefire , thank you for taking the lead on this. After partially reviewing and testing the code you worked on, I can see it's functioning well. These issues often demand more thought than actual coding, and your proactive approach is greatly appreciated. Reflecting on this, there are a few considerations we need to address to identify the most effective and consistent solution:

Currently, for Python objects, we're operating under the assumption that 'everything has a segment'. This means that frames without an explicit segment are automatically grouped into a single segment, which has led to this bug.
Considering a fix, we're faced with two potential directions:
- Align our code's behavior more closely with ndjson, which, from my understanding, is the route you're currently exploring.
- Alternatively, adjust our code to treat each frame as its own distinct segment and address any resulting bugs from this change.
Choosing the best path forward requires a deeper understanding of why our object serialization behaves differently from ndjson serialization processes. Do you have any insights here?

This is possible 1-2 points for me to get to the bottom of it, and prob total of 3 if I had to code it. I will ask @adrian-chang to prioritize it

Gabefire:
@vbrodsky, I appreciate you taking the time to look at this draft! I think point 2 would be a less time-consuming solution, but the only problem with just doing point 2 is you miss out on the opportunity of having a nice segment line inside the UI with consecutive frames with the same answer link. You also won't be able to specify custom segments like if I wanted to split up a segment line of the same answers into different pieces. Below is what I mean by the linked answers:
segment
Treating each frame as a segment also changes the export, where you will see an answer for each frame. Instead of groups of consecutive frames with answers.  This could mean that these mal import exports could behave differently then if they were hand labeled as likely when you hand label you would have a segment line of consecutive answers. I think this might be the reason why the object serializes differently. You can see from the prior code I think someone recognized this as they tried grouping segments together. We could do a sort combo of point 1 and 2. right now my code just groups the answers of each features together with consecutive frames shown. I tired to go further and then group the frames with the same feature together but that was not as clean... So we could treat each separate label answer as a there own distinct segment combo and put them all on the same line

labelbox/data/serialization/ndjson/label.py
```diff
@@ -227,17 +232,38 @@ def _create_video_annotations(
                 yield NDObject.from_common(annotation=annot, data=label.data)
 
         for annotation_group in video_annotations.values():
-            segment_frame_ranges = cls._get_segment_frame_ranges(
-                annotation_group)
             if isinstance(annotation_group[0], VideoClassificationAnnotation):
-                annotation = annotation_group[0]
```
Gabefire, line 233: @vbrodsky This line here is were the bulk of the problem comes from on prior solution it is take the first label of the group with its answer and then adding the frames to that. That is why it shows one answer for the entire video

vbrodsky: Thanks, @Gabefire, for another insightful observation!

I'd like to propose an alternative method to address this issue: by assigning a new segment to each frame that lacks one, in addition to implementing a fix for the issue you are pointing out above. It's important to note that our current approach groups all frames without a specified segment into a single, default segment.

This comment is meant to provide a comprehensive view of the situation. I've initiated a separate story and have already informed @adrian-chang about it. This approach, while potentially beneficial, would require careful consideration and likely some time to resolve the questions raised in my comments on the previous PR.

Gabefire: Hey Val! Sounds good! Let me know if you need anything. It was fun looking through this!

Gabefire:
Closing for now

PR #1407, opened 2024-02-06
Increment sdk version and update CHANGELOG

kozikkamil:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+# Version 3.62.0 (2024-02-06)
+## Added 
+* Support custom_metrics at the annotation level for MEA imports
```
vbrodsky, line 5: Assuming a method is affected, or a class, can we follow the format in 3.61.2

vbrodsky (commented):
Left a question, also is there a new class being added? If yes, I recommend to also update https://github.com/Labelbox/labelbox-python/blob/develop/docs/source/index.rst  in this PR

PR #1408, opened 2024-02-06
[AL-0] Fix S2A validation tests

whistler:
Fix invalid feature schema mappings that broke these tests after adding validation for the mapping in https://labelbox.atlassian.net/browse/AL-7547

### DIFF
--- a/tests/integration/annotation_import/test_send_to_annotate_mea.py
+++ b/tests/integration/annotation_import/test_send_to_annotate_mea.py
@@ -6,13 +6,15 @@ from labelbox.schema.conflict_resolution_strategy import ConflictResolutionStrat
 
 def test_send_to_annotate_from_model(client, configured_project,
                                      model_run_predictions,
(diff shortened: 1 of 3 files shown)


tests/integration/annotation_import/test_send_to_annotate_mea.py
```diff
@@ -6,13 +6,15 @@
...
                                      model_run_predictions,
-                                     model_run_with_data_rows,
-                                     project_with_ontology):
+                                     model_run_with_data_rows, project):
     model_run = model_run_with_data_rows
     data_row_ids = [p['dataRow']['id'] for p in model_run_predictions]
     assert len(data_row_ids) > 0
 
-    destination_project, _ = project_with_ontology
+    destination_project = project
```
vbrodsky, line 14: does it make any sense to setup a fixture to wrap these 4 lines? I think it is the same code as below

vbrodsky (approved):

PR #1412, opened 2024-02-08
Vb/support pydantic2 plt 145

vbrodsky:
Story: https://labelbox.atlassian.net/browse/PLT-145

Please review for background and implementation approach: https://labelbox.atlassian.net/wiki/spaces/~63d753c8724a5c79c7b69a36/pages/2369257507/SDK+Pydantic2+Pydantic1+Compatibility

This PR introduces several crucial updates aimed at enhancing our sdk compatibility with Pydantic2. The changes are as follows:
- **Implementation of a Pydantic Compatibility Module**: This is the centerpiece of the PR. The module is designed to facilitate our future usage of Pydantic v1 up to  transition to Pydantic v2 or some other DTO implementation
- **Updates to All Files Utilizing Pydantic**: These changes are captured in the commit **2611b389a56c2a88eae8c488db6ecbc2f8e96407**. Given the comprehensive nature of our test automation tools, I believe manual review of each file for Pydantic-related updates is unnecessary. Our automated tests are robust enough to ensure that the updates meet our quality standards without requiring manual inspection of every change.
- **Removal of Pydantic v2 Restriction**: We have eliminated the restriction on Pydantic v2 from both requirements.txt and setup.py.

### DIFF
--- a/labelbox/pydantic_compat.py
+++ b/labelbox/pydantic_compat.py
@@ -8,14 +8,13 @@ def pydantic_import(class_name, sub_module_path: Optional[str] = None):
     # Get the version of pydantic
     pydantic_version = pkg_resources.get_distribution("pydantic").version
 
-    # Check if the version is 1
-    if pydantic_version.startswith("1"):
-        pydantic_v1_module_name = "pydantic" if sub_module_path is None else f"pydantic.{sub_module_path}"
-    else:  # use pydantic 2 v1 thunk
-        pydantic_v1_module_name = "pydantic.v1" if sub_module_path is None else f"pydantic.{sub_module_path}"
+    # Determine the module name based on the version
+    module_name = "pydantic" if pydantic_version.startswith(
+        "1") else "pydantic.v1"
+    module_name = f"{module_name}.{sub_module_path}" if sub_module_path else module_name
 
-    klass = getattr(importlib.import_module(pydantic_v1_module_name),
-                    class_name)
+    # Import the class from the module
+    klass = getattr(importlib.import_module(module_name), class_name)
 
     return klass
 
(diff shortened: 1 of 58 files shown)


labelbox/pydantic_compat.py
```diff
@@ -0,0 +1,35 @@
...
+    import pkg_resources
+
+    # Get the version of pydantic
+    pydantic_version = pkg_resources.get_distribution("pydantic").version
+
+    # Check if the version is 1
+    if pydantic_version.startswith("1"):
+        pydantic_v1_module_name = "pydantic" if sub_module_path is None else f"pydantic.{sub_module_path}"
+    else:  # use pydantic 2 v1 thunk
+        pydantic_v1_module_name = "pydantic.v1" if sub_module_path is None else f"pydantic.{sub_module_path}"
```
mrobers1982, line 15: I would assume we want `pydantic.v1` in both cases.  If so, why not?

vbrodsky: Well what if users are still using pydantic v1? Currently we need to support this case so we will do `from pydantic` instead of `from pydantic.v1`

mrobers1982: I thought that case is captured by the `if` condition?

I'm referring to the `f"pydantic.{sub_module_path}"` if the `else` condition which assumes the usage of Pydantic 2.x.

vbrodsky: Good point

labelbox/pydantic_compat.py
```diff
@@ -8,14 +8,13 @@ def pydantic_import(class_name, sub_module_path: Optional[str] = None):
...
-        pydantic_v1_module_name = "pydantic" if sub_module_path is None else f"pydantic.{sub_module_path}"
-    else:  # use pydantic 2 v1 thunk
-        pydantic_v1_module_name = "pydantic.v1" if sub_module_path is None else f"pydantic.{sub_module_path}"
-
-    klass = getattr(importlib.import_module(pydantic_v1_module_name),
-                    class_name)
+    # Determine the module name based on the version
+    module_name = "pydantic" if pydantic_version.startswith(
+        "1") else "pydantic.v1"
+    module_name = f"{module_name}.{sub_module_path}" if sub_module_path else module_name
```
mrobers1982, line 14: 👍

mrobers1982 (approved):
Good start!

sfendell-labelbox (approved):

PR #1413, opened 2024-02-09
[PLT-58] Vb/fix all model runs export timeout plt 58

vbrodsky:
Story Overview: [PLT-58](https://labelbox.atlassian.net/browse/PLT-58)

Objective: The primary aim of this story was to diagnose and resolve issues related to export timeouts when dealing with a dataset comprising 7 million rows, specifically when using the options all_project and all_labels.

Root Cause: We identified that the direct cause of the timeout was the Elasticsearch (ES) query that retrieves 7 million data row IDs. This query is executed within the synchronous API responsible for initiating the mutation process. Due to time constraints and potential risks, we opted against optimizing the API at this juncture.

Current Solution: This Pull Request (PR) tackles the issue by extending the HTTP read timeout, extending task wait and adding error handling for network timeouts. These adjustments collectively facilitate the fetching of the large dataset.

Additional Requirements: It's important to note that implementing this PR successfully also hinges on specific infrastructure changes. These changes have been outlined and are currently under discussion [here](https://labelbox.slack.com/archives/CLLA9ATM3/p1707423276607449).

Next Steps: Aside from the solutions proposed in this PR, I am also exploring a workaround. This workaround would leverage our existing codebase to meet our objectives without necessitating further updates to the API.

[PLT-58]: https://labelbox.atlassian.net/browse/PLT-58?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/labelbox/client.py
+++ b/labelbox/client.py
@@ -217,7 +217,9 @@ class Client:
             elif response.status_code == 502:
                 error_502 = '502 Bad Gateway'
                 raise labelbox.exceptions.InternalServerError(error_502)
-
+            elif response.status_code == 504:
+                error_504 = 'Gateway Timeout'
+                raise labelbox.exceptions.InternalServerError(error_504)
             raise labelbox.exceptions.LabelboxError(
                 "Failed to parse response as JSON: %s" % response.text)
 
--- a/labelbox/client_constants.py
+++ b/labelbox/client_constants.py
@@ -1,3 +1,2 @@
-# type: ignore
 class ClientConstants:
     EXPORT_SUBMISSION_TIMEOUT = 60 * 10  # seconds
--- a/labelbox/client_constants.py
+++ b/labelbox/client_constants.py
@@ -1,0 +1,3 @@
+# type: ignore
+class ClientConstants:
+    EXPORT_SUBMISSION_TIMEOUT = 60 * 10  # seconds
--- a/labelbox/schema/data_row.py
+++ b/labelbox/schema/data_row.py
@@ -1,6 +1,8 @@
 import logging
 from typing import TYPE_CHECKING, Collection, Dict, List, Optional, Union
 import json
+import pdb
+from labelbox.client_constants import ClientConstants
 from labelbox.exceptions import ResourceNotFoundError
 
 from labelbox.orm import query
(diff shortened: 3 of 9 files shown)


itssujan:
Can you point me to the code having the ES query? are we not paging and why would the query timeout in that case? may be we are doing something wrong thr?

mrobers1982:
I think the PR is premature as we still need to identify the cause of the timeout - it also sounds like we will not be updating the timeouts for any of the proxies - see discussion mentioned in the top-level PR message.

vbrodsky:
> I think the PR is premature as we still need to identify the cause of the timeout ...

@mrobers1982 I disagree. We know the cause of the timeout (perhaps I did not do a good job explaining as how I determined this to be the cause - will do it now below). While we can not directly confirm the cause via, say DD traces (we can also talk as to why) or logs or any other sort of direct observations there is a pretty strong causal link:
- I have done a number of experiments with various export options for this 7 mlns dataset data rows and the only one the api times out for is with the option `"all_model_runs": True`.
- I have examined the code that would be executed for this case and here is the main and only difference for this option and other code execution path: https://github.com/Labelbox/intelligence/blob/develop/apps/api/src/exports-v2/resolvers/export-data.resolver.ts#L241-L248
- If you follow this thread, you will see this method executes pages search of all data rows for the data set: https://github.com/Labelbox/intelligence/blob/develop/apps/api/src/exports-v2/resolvers/export-data.resolver.ts#L369

mrobers1982:
> > I think the PR is premature as we still need to identify the cause of the timeout ...
> 
> @mrobers1982 I disagree. We know the cause of the timeout (perhaps I did not do a good job explaining as how I determined this to be the cause - will do it now below). While we can not directly confirm the cause via, say DD traces (we can also talk as to why) or logs or any other sort of direct observations there is a pretty strong causal link:
> 
> * I have done a number of experiments with various export options for this 7 mlns dataset data rows and the only one the api times out for is with the option `"all_model_runs": True`.
> * I have examined the code that would be executed for this case and here is the main and only difference for this option and other code execution path: https://github.com/Labelbox/intelligence/blob/develop/apps/api/src/exports-v2/resolvers/export-data.resolver.ts#L241-L248
> * If you follow this thread, you will see this method executes pages search of all data rows for the data set: https://github.com/Labelbox/intelligence/blob/develop/apps/api/src/exports-v2/resolvers/export-data.resolver.ts#L369

I was referring to the 120s Gateway timeout that was brought up in the Slack discussion, do we have an idea what component in our infra is responsible for the timeout?

vbrodsky:
Will close this PR, see https://labelbox.atlassian.net/browse/PLT-58?focusedCommentId=181951 for more

PR #1429, opened 2024-02-21
[PLT-71] Relationship imports

mnoszczak:
Latest CR changes broke the fix (while attempting to make it slightly more performant), added the test case which surfaces this issue, reverted the latest commit from previous PR

### DIFF
--- a/labelbox/data/serialization/ndjson/converter.py
+++ b/labelbox/data/serialization/ndjson/converter.py
@@ -106,7 +106,6 @@ class NDJsonConverter:
                     if not isinstance(annotation, RelationshipAnnotation):
                         uuid_safe_annotations.append(annotation)
             label.annotations = uuid_safe_annotations
-            print(label.annotations)
             for example in NDLabel.from_common([label]):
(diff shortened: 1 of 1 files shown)


labelbox/data/serialization/ndjson/converter.py
```diff
@@ -74,39 +74,39 @@ def serialize(
...
-                        used_uuids.add(annotation._uuid)
+                    if annotation._uuid in used_uuids:
+                        annotation._uuid = uuid.uuid4()
+                    used_uuids.add(annotation._uuid)
                     uuid_safe_annotations.append(annotation)
+                else:
+                    if not isinstance(annotation, RelationshipAnnotation):
+                        uuid_safe_annotations.append(annotation)
             label.annotations = uuid_safe_annotations
+            print(label.annotations)
```
mihhail-m, line 109: this perhaps not needed here. Otherwise going to clutter terminal stdout

vbrodsky (approved):

PR #1432, opened 2024-02-22
[PLT-319] Add Catalog class, export_v2 method 

mnoszczak:
Allows to make exports from catalog without the need to create a slice prior to that. Support for streamable and non-streamable.
```
import labelbox as lb
from labelbox.schema.catalog import Catalog
"""# Replace with your API Key"""

API_KEY = ""
ENDPOINT = "https://app.lb-stage.xyz/api/_gql"

client = lb.Client(API_KEY, endpoint=ENDPOINT)
catalog = client.get_catalog()
task = catalog.export_v2(filters={
    "last_activity_at": ["2024-02-01 00:00:00", "2050-01-01 00:00:00"],
})
```

### DIFF
--- a/labelbox/schema/catalog.py
+++ b/labelbox/schema/catalog.py
@@ -29,7 +29,8 @@ class Catalog:
         
         >>>     import labelbox as lb
         >>>     client = lb.Client(<API_KEY>)
-        >>>     task = Catalog.export_v2(
+        >>>     catalog = client.get_catalog()
+        >>>     task = catalog.export_v2(
         >>>         filters={
         >>>             "last_activity_at": ["2000-01-01 00:00:00", "2050-01-01 00:00:00"],
         >>>             "label_created_at": ["2000-01-01 00:00:00", "2050-01-01 00:00:00"],


labelbox/schema/catalog.py
```diff
@@ -0,0 +1,169 @@
...
+from labelbox.schema.export_task import ExportTask
+from labelbox.schema.task import Task
+
+from typing import TYPE_CHECKING
+import sys
+if TYPE_CHECKING or 'pytest' in sys.modules:
+    from labelbox import Client
+
+
+class Catalog:
```
vbrodsky, line 17: You need to add this class here labelbox/__init__.py

labelbox/schema/catalog.py
```diff
@@ -0,0 +1,169 @@
...
+        >>>         filters={
+        >>>             "last_activity_at": ["2000-01-01 00:00:00", "2050-01-01 00:00:00"],
+        >>>             "label_created_at": ["2000-01-01 00:00:00", "2050-01-01 00:00:00"],
+        >>>         },
+        >>>         params={
+        >>>             "performance_details": False,
+        >>>             "label_details": True
+        >>>         })
+        >>>     task.wait_till_done()
+        >>>     task.result
```
vbrodsky, line 76: Will `task.result` work here? For example in my scripts I do smth like
```
    print("Started data extraction at ", datetime.now())
    data = list(
        map(lambda x: json.loads(x.json_str),
            export_task.get_stream()))
    return data
```

tests/integration/export/legacy/test_export_catalog.py
```diff
@@ -0,0 +1,19 @@
+import pytest
```
vbrodsky, line 1: why not adding a test for streamable export?

labelbox/client.py
```diff
@@ -117,6 +119,7 @@ def __init__(self,
             'X-Python-Version': f"{python_version_info()}",
         }
         self._data_row_metadata_ontology = None
+        Client.instance = self
```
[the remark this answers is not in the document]
mrobers1982: Can we not do this?

Is there a reason to not create an instance of `Catalog` and pass in an instance of a `Client`.

vbrodsky (approved):

PR #1433, opened 2024-02-22
DEVOPS-2168 - Securing GHA Publish to pypi

abacchilb:
Based on security best practice - I split up the build and publish process so that the `      id-token: write` is available in the minimum amount of steps. 
Also added environment so the publish to pypi portion so we can add further security later. Removed twine install because it was no longer needed.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -4,39 +4,46 @@ name: Publish Python Package
 
 on:
   release:
-    types: [created, updated]
+    types: [created]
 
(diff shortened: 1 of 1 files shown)


abacchilb:
@vbrodsky would need to update the pypi environment with 'publish'
image

vbrodsky (approved):

abacchilb:
https://www.loom.com/share/f9d6bc380c5a4c65a1d1a7d677575430

PR #1437, opened 2024-02-26
Add example for tags management

paulnoirel:
Add an example showing how to create a tag, add it to a project and list the tags of a project.


paultancre (approved):
LGTM

PR #1438, opened 2024-02-27
Update notebook for Custom metrics at the feature-level in Model

paulnoirel:
This is an updated version of the Custom Metrics demo based on the notebook used by Kamil K.


ovalle15 (approved):
I have verified  my recent updates work , will merge to develop once tests complete

PR #1439, opened 2024-02-28
[PLT-130] Validate "external_id", "global_key", "row_data" fields in row update

sfendell-labelbox:
None of these fields should be set to an empty value, and at least 1 is needed for an update.

### DIFF
--- a/labelbox/schema/data_row.py
+++ b/labelbox/schema/data_row.py
@@ -1,7 +1,6 @@
 import logging
-from typing import TYPE_CHECKING, Collection, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, List, Optional, Union
 import json
-from labelbox.exceptions import ResourceNotFoundError
 
 from labelbox.orm import query
 from labelbox.orm.db_object import DbObject, Updateable, BulkDeletable, experimental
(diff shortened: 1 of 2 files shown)


labelbox/schema/data_row.py
```diff
@@ -75,6 +73,8 @@ def update(self, **kwargs):
         # Convert row data to string if it is an object
         # All other updates pass through
         row_data = kwargs.get("row_data")
+        if not row_data and not kwargs.get("external_id"):
```
mrobers1982, line 76: The validation in ADV enforces that if `row_data` or `global_key` or `external_id` is set, it must not be an empty string.

What you have above is likely too strict, e.g. a user should be able to only update the `global_key`.

sfendell-labelbox: Thanks, fixed

labelbox/schema/data_row.py
```diff
@@ -75,6 +73,9 @@ def update(self, **kwargs):
         # Convert row data to string if it is an object
         # All other updates pass through
         row_data = kwargs.get("row_data")
+        if not row_data and not kwargs.get("external_id") and not kwargs.get(
```
mrobers1982, line 76: I do not think you implemented the logic I mentioned here - https://github.com/Labelbox/labelbox-python/pull/1439#discussion_r1505196585.

What happens if `row_data` is set to a non-empty value and `global_key` is set to an empty value?

You might need to differentiate a field being unset from it being set but to an empty value.

mrobers1982 (approved):

PR #1444, opened 2024-02-29
few fixes to custom basics

ovalle15:
-  updated the iuo in the methods name, since this is a reserved name. 
-  removed hardcoded variables


mina-ebeid (approved):
Looks good! Notebook is working fine. Will continue to look into the methods used to validate

PR #1447, opened 2024-03-01
Release v.3.64.0

sfendell-labelbox:
# Version 3.64.0 (2024-02-29)

## Added
* Catalog schema class. Catalog exports can now be made without creating a slice first

## Removed
* Review related WebhookDataSource topics


vbrodsky (changes requested):
Looks like there were lots of notebook updates and I think this sdk release is mainly to address it. Can we updated the CHANGELOG

sfendell-labelbox:
> Looks like there were lots of notebook updates and I think this sdk release is mainly to address it. Can we updated the CHANGELOG

Done.

vbrodsky (approved):

PR #1455, opened 2024-03-06
[SN-50] Create demo on how to export predictions from a model-run and import them as pre-labels in a project

Gabefire:
Took advantage of a method that came with Foundry `send_to_annotate_from_model`.


Gabefire:
FYI @ovalle15 the project notebook was included because it had meta data. I had another PR merged that removed it so Ill rebase this branch to get that notebook of this PR

PR #1460, opened 2024-03-07
[PLT-344] Add dataset.upsert_data_rows method

attila-papai:
This PR adds support for upserting data rows to a dataset, meaning the provided data row specifications either create new data rows or update existing ones. 

Please see the added unit tests for example usages of this new functionality.


vbrodsky (commented):
I have a question regarding Task/returns.

A few months ago, when we were discussing upserts, we talked about returning results, specifically successes and failures (and warnings? partial successes?). I think there was an issue with how create_data_rows currently works, but I don't remember all the details. Oh yeah, one of them was like how many data rows we returns errors? I.e. is there a limit etc Perhaps Matt C. might recall better. Is this issue within the scope of the current implementation to review task return to make it support 'infinite' number of rows?

labelbox/schema/dataset.py
```diff
@@ -750,3 +755,35 @@ def _export(
         res = res[mutation_name]
         task_id = res["taskId"]
         return Task.get_task(self.client, task_id)
+
+    def upsert_data_rows(self, items: list[DataRowUpsertItem]) -> "Task":
```
mrobers1982, line 759: For now, let's also enforce this limit - https://github.com/Labelbox/labelbox-python/blob/develop/labelbox/schema/dataset.py#L34.

This matches the expectation for imports, in the future I can update the backend to remove this limitation.

labelbox/schema/dataset.py
```diff
@@ -750,3 +755,35 @@ def _export(
...
+        chunk_size = 3
+        chunks = [
+            items[i:i + chunk_size] for i in range(0, len(items), chunk_size)
+        ]
+        manifest = {"source": "SDK", "item_count": 0, "chunk_uris": []}
+        for chunk in chunks:
+            manifest["chunk_uris"].append(self._create_descriptor_file(chunk))
```
mrobers1982, line 766: If we re-use this method we need to expose setting the file name; otherwise I believe it will just overwrite the file if there are multiple chunks.

labelbox/schema/dataset.py
```diff
@@ -750,3 +755,35 @@ def _export(
...
+        dataset_param = "datasetId"
+        manifest_uri_param = "manifestUri"
+        query_str = """mutation UpsertDataRowsPyApi($%s: ID!, $%s: String!){
+                upsertDataRows(data:{datasetId: $%s, manifestUri: $%s}
+                ){ id createdAt updatedAt name status completionPercentage result errors type metadata } } """ % (
+            dataset_param, manifest_uri_param, dataset_param,
+            manifest_uri_param)
+
+        res = self.client.execute(query_str, {
+            dataset_param: self.uid,
```
mrobers1982, line 783: We should validate the dataset id set on each specification (if any) matches what is set here.

labelbox/schema/data_row.py
```diff
@@ -17,6 +19,35 @@
...
+class ResolvableId(BaseModel):
+    """
+    The ResolvableId class is a unique ID abstraction that allows us to reference
+    a DataRow by either a Global key or CUID
+    """
+    type: KeyType = KeyType.GKEY
+    value: str
+
+
+class DataRowSpec(BaseModel):
```
mrobers1982, line 40: This will ultimately need to support metadata, attachments, embeddings, .etc, essentially all the fields supported by import.

labelbox/schema/data_row.py
```diff
@@ -17,6 +19,35 @@
...
+    row_data: Union[str, dict]
+    external_id: Optional[str]
+    global_key: Optional[str]
+
+
+class DataRowUpsertItem(BaseModel):
+    id: ResolvableId
```
mrobers1982, line 47: We might consider abstracting this concept from the user and map the user-facing entity to this type which the backend expects.

One example is the user can optionally provide a key; otherwise we assume it will be auto-generated (on create).

labelbox/schema/dataset.py
attila-papai: it won't overwrite it, an UUID is appended to the filename as a prefix (in api)

labelbox/schema/dataset.py
```diff
@@ -750,3 +750,50 @@ def _export(
         res = res[mutation_name]
         task_id = res["taskId"]
         return Task.get_task(self.client, task_id)
+
+    def upsert_data_rows(self, items) -> "Task":
```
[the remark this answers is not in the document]
attila-papai: We make chunks of these items and upload them to LB as separate JSON files. Then we create a manifest file referencing all the chunk files and also upload the manifest file to LB. Then we call backend with the manifest URL only. The Data Ingestion system will then parallelize the import as much as possible.

labelbox/schema/dataset.py
```diff
@@ -32,6 +32,7 @@
 logger = logging.getLogger(__name__)
 
 MAX_DATAROW_PER_API_OPERATION = 150_000
+UPSERT_CHUNK_SIZE = 10_000
```
mrobers1982: Perhaps we should mangle this value by prefixing with `__` - in general we should do better in this project when defining "private" variables.

labelbox/schema/data_row.py
```diff
@@ -17,6 +21,79 @@
...
+    name: Optional[str]
+
+
+class DataRowAttachmentSpec(BaseModel):
+    type: AttachmentType
+    value: str
+    name: Optional[str]
+
+
+class DataRowSpec(BaseModel):
```
mrobers1982: I'm not sure we should include these models here - they are only used for `upsert` and have very generic names which might lead to confusion on when to use and why and IMO we do not want to set any new standards as part of this change.

We can think on this more but we most likely want to get buy-in from code-owners and/or MLSE.

labelbox/schema/dataset.py
```diff
@@ -293,7 +294,10 @@ def create_data_rows(self, items) -> "Task":
         task._user = user
         return task
 
-    def _create_descriptor_file(self, items, max_attachments_per_data_row=None):
+    def _create_descriptor_file(self,
+                                items,
+                                max_attachments_per_data_row=None,
+                                is_upsert=False):
```
mrobers1982, line 308: I understand you wanted to keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), but the code is arguably more complex as a result.

labelbox/schema/dataset.py
```diff
@@ -749,3 +765,94 @@ def _export(
...
+        >>>         ),
+        >>>     ])
+        >>>     task.wait_till_done()
+        """
+        if len(specs) > MAX_DATAROW_PER_API_OPERATION:
+            raise MalformedQueryException(
+                f"Cannot upsert more than {MAX_DATAROW_PER_API_OPERATION} DataRows per function call."
+            )
+
+        class ManifestFile:
```
mrobers1982: I understand functions within functions is a pattern already used in this file but I would avoid this pattern unless there's not viable alternative - there's a performance penalty, not to mention it makes the code more difficult to read.

labelbox/schema/dataset.py
```diff
@@ -749,3 +765,94 @@ def _export(
...
+
+        items = _convert_specs_to_upsert_items(specs)
+        chunks = [
+            items[i:i + UPSERT_CHUNK_SIZE]
+            for i in range(0, len(items), UPSERT_CHUNK_SIZE)
+        ]
+        manifest = ManifestFile()
+        for chunk in chunks:
+            manifest.chunk_uris.append(
+                self._create_descriptor_file(chunk, is_upsert=True))
```
mrobers1982: I understand this is an optimization but did you consider uploading the chunks in parallel?

labelbox/schema/data_row.py
attila-papai: in the follow up refactor ticket, I was gonna reuse this for the create methods
IMO we better start using types to make this SDK library easier to use. Without it, clients need to either read the docs or the underlying code itself, especially when we use `*args` and `**kwargs` that makes it even more complicated to understand what variables are supported.

labelbox/schema/dataset.py
attila-papai: I will address this in the follow-up work.

labelbox/schema/dataset.py
attila-papai: I didn't want to use a class in the first place. Sync up with @vbrodsky 
I'm gonna remove it for now.

labelbox/schema/data_row.py
mrobers1982: Can you provide more details re: what you plan to refactor?

I'm worried the scope is too large - you mention refactoring `create_data_rows` but you also need to be mindful that we do not break existing implementations as that would seemingly imply a major upgrade to the SDK.

Personally, I believe we should aim our sights lower with a minimal integration, minimizing changes to existing patterns.

Per above, if you want to continue down this path we likely need to get buy-in from the SDK team as they would likley own this work.

labelbox/schema/dataset.py
mrobers1982: My comment was not so much in the class as the functions within functions, personally I would extract those to avoid the performance penalty as I do not think you need the closure.

labelbox/schema/asset_attachment.py
```diff
@@ -61,7 +61,7 @@ def validate_attachment_value(cls, attachment_value: str) -> None:
 
     @classmethod
     def validate_attachment_type(cls, attachment_type: str) -> None:
-        valid_types = set(cls.AttachmentType.__members__)
+        valid_types = set(AttachmentType.__members__)
```
vbrodsky, line 64: why not make a method on AttachmentType `valid_attachment` ?

labelbox/schema/dataset.py
```diff
@@ -47,6 +47,8 @@ class Dataset(DbObject, Updateable, Deletable):
         created_by (Relationship): `ToOne` relationship to User
         organization (Relationship): `ToOne` relationship to Organization
     """
+    __upsert_chunk_size: Final = 10_000
```
vbrodsky, line 57: how was this number chosen?

labelbox/schema/data_row.py
sfendell-labelbox: I'd love to have types in all these methods, but I think Matt's right; it's not trivial to add typing to these methods without breaking backward compatibility, and until we do that, we'd have 2 similar methods taking in arguments in different formats.

Since we can't add typing to other methods immediately, we should aim for consistency for now. If we do upgrade to typed arguments, we can do it in all the relevant methods at once, in a single PR, and announce it in patch notes for the relevant SDK version.

Could you file a ticket (if we don't have one already) for adding types here and put it in backlog so this doesn't get lost?

labelbox/schema/dataset.py
sfendell-labelbox: I agree solely on readability concerns, no need to nest the function instead of making it a private func if it doesn't use anything from the outer scope.

Or even just not having a function for this at alll; the function is defined inline right before it's used for the only time so it's not like it's encapsulating anything or reducing line count.

labelbox/schema/dataset.py
mrobers1982: nit: In Python, constants are typically written in all capital letters.

labelbox/schema/dataset.py
```diff
@@ -749,3 +773,106 @@ def _export(
...
+        >>>             "global_key": "global_key1",
+        >>>             "external_id": "ex_id1",
+        >>>             "attachments": [
+        >>>                 {"type": AttachmentType.RAW_TEXT, "name": "att1", "value": "test1"}
+        >>>             ],
+        >>>             "metadata": [
+        >>>                 {"name": "tag", "value": "tag value"},
+        >>>             ]
+        >>>         },
+        >>>         # update existing data row by global key
```
mrobers1982: While it does not look like the code supports this case, I would not give an end-user multiple ways of doing the same thing, it will only create confusion.

labelbox/schema/dataset.py
```diff
@@ -749,3 +773,106 @@ def _export(
...
+        >>>         },
+        >>>     ])
+        >>>     task.wait_till_done()
+        """
+        if len(items) > MAX_DATAROW_PER_API_OPERATION:
+            raise MalformedQueryException(
+                f"Cannot upsert more than {MAX_DATAROW_PER_API_OPERATION} DataRows per function call."
+            )
+
+        def _convert_items_to_upsert_format(_items):
```
mrobers1982: There was an ask to remove the nested functions, can you take care of that too?

labelbox/schema/dataset.py
```diff
@@ -749,3 +773,106 @@ def _export(
...
+        specs = _convert_items_to_upsert_format(items)
+        chunks = [
+            specs[i:i + self.__upsert_chunk_size]
+            for i in range(0, len(specs), self.__upsert_chunk_size)
+        ]
+
+        def _upload_chunk(_chunk):
+            return self._create_descriptor_file(_chunk, is_upsert=True)
+
+        file_upload_thread_count = 20
```
mrobers1982: Personally, I would expose this value in the method signature and provide a reasonable default.

labelbox/schema/dataset.py
```diff
@@ -749,3 +773,106 @@ def _export(
...
+        def _convert_items_to_upsert_format(_items):
+            _upsert_items: List[DataRowUpsertItem] = []
+            for item in _items:
+                # enforce current dataset's id for all specs
+                item['dataset_id'] = self.uid
+                if 'key' not in item:
+                    key = {'type': 'AUTO', 'value': ''}
+                elif isinstance(item['key'], UniqueId):
+                    key = {'type': 'ID', 'value': item['key'].key}
+                    del item['key']
```
mrobers1982: nit: When setting the value in the line above, you can also use `pop` to retrieve the value and delete the key from the dictionary.

mrobers1982 (changes requested):
A couple minor things, but almost there - thanks for taking care of this!

mrobers1982 (approved):
Let's ultimately wait to merge until we have an approval from @sfendell-labelbox or @vbrodsky.

mrobers1982:
Let's potentially hold off on merging this change if product wants us to move the implementation from `Dataset` to `Client`.

https://labelbox.atlassian.net/browse/PLT-111

mrobers1982:
> Let's potentially hold off on merging this change if product wants us to move the implementation from `Dataset` to `Client`.
> 
> https://labelbox.atlassian.net/browse/PLT-111

~~Actually, let's assume we are moving to the `Client` and make these changes.~~

Product wants to get feedback from stakeholders before we make this change.

mrobers1982:
> > Let's potentially hold off on merging this change if product wants us to move the implementation from `Dataset` to `Client`.
> > https://labelbox.atlassian.net/browse/PLT-111
> 
> ~Actually, let's assume we are moving to the `Client` and make these changes.~
> 
> Product wants to get feedback from stakeholders before we make this change.

Product wants to leave as-is for now - https://labelbox.atlassian.net/browse/PLT-111?focusedCommentId=192239.

We can go ahead and merge if everyone is happy with the current implementation.

vbrodsky (approved):

PR #1474, opened 2024-03-11
[SN-127] included the get_catalog() method to the export notebook

ovalle15:
- Removed some of the redundant commented-out code
- Updated some missing details 
- Added get_catalog() method


mina-ebeid:
Is the project_id filter in the params section supposed to only export from catalog the datarows that are in the project? I tried using it and it kept returning all datarows in my catalog

mina-ebeid (approved):
Other than project_ids, this looks like its good to go!

PR #1476, opened 2024-03-12
[WIP] [PLT-0] Update runner to run unit independently from integration 

adrian-chang:
Split unit tests from integration

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -69,8 +69,7 @@ jobs:
         run: |
           pip install mypy==1.9.0
           # TODO: consider installing requests typing
-          mypy --install-types
-          mypy -p labelbox --pretty --show-error-codes
+          mypy -p labelbox --pretty --show-error-codes --non-interactive --install-types
       - name: Install package and test dependencies
         run: |
           pip install tox==4.14.1 pytest-snapshot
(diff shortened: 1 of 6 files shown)


tests/unit/test_unit_query.py
```diff
@@ -20,6 +20,7 @@ def test_query_subquery():
 
 
 def test_query_where():
+    print('foo')
```
sn0v, line 23: let's lose this?

.github/workflows/python-package.yml
```diff
@@ -76,22 +77,40 @@ jobs:
...
         env:
           # make sure to tell tox to use these environs in tox.ini
           LABELBOX_TEST_API_KEY_PROD: ${{ secrets[matrix.prod-key] }}
-
           LABELBOX_TEST_API_KEY_STAGING: ${{ secrets[matrix.staging-key] }}
-
           DA_GCP_LABELBOX_API_KEY: ${{ secrets[matrix.da-test-key] }}
         run: |
-          tox -e py -- -n 10 -svv --reruns 5 --reruns-delay 8
+          tox -e py -- -n 10 -svv --reruns 5 --reruns-delay 8 --testmon tests/unit
```
sn0v, line 111: not too familiar with tox but shouldn't `tests/unit` be something else here?

PR #1479, opened 2024-03-12
[PLT-0] Speed up tests, Prep for Python 3.11, 3.12, Deprecate Python 3.7

adrian-chang:
* Prepare for 3.11, 3.12 Python Support (EG I am planning getting off setuptools if possible and move to either Rye / PDM / Poetry (slight preference for Rye)
* Prepare to eliminate Python 3.7 Support
* Speed up tests by trying to cache them with pycov-testmon and Github actions cache
* Consider adding (in the future) py-cov for SDK test coverage

This really should not go probably until I change the package manager


labelbox/schema/id_type.py
```diff
@@ -1,7 +1,7 @@
-from enum import Enum
+from strenum import StrEnum
 
 
-class IdType(str, Enum):
+class IdType(StrEnum):
```
adrian-chang, line 4: https://blog.pecar.me/python-enum -> see this for anyone with why about this for 3.11 support

setup.py
```diff
@@ -33,9 +33,11 @@
...
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11',
+        'Programming Language :: Python :: 3.12',
     ],
     python_requires='>=3.7',
```
sfendell-labelbox, line 42: Are we dropping 3.7? If so, update this too please

adrian-chang: Ugh maybe I should just upgrade the package manager now as then I wouldn't have to do this, but scope is then getting too large, I'll update this

sfendell-labelbox (approved):

PR #1487, opened 2024-03-15
[SN-89] Remove/ update integration notebooks

Gabefire:


Gabefire:
removed examples/integrations/tlt/labelbox_upload.ipynb since the urls are broken and the dataset it needed was 6 gb. Also it just goes over a basic demo of labelbox which we have plenty of

Gabefire:
Removed

Gabefire:
Made the decision to delete the databricks integrations they require Java runtime to work which makes them tricky, labelsparks is currently out of date and labelsparks has notebook examples of its own. If you disagree let me know.

Gabefire:
removed detectron notebooks since they depending on coco serializer to work and we have no plans of fixing its many issues. Can add back later if/ when coco stuff gets fixed

Gabefire:
Improved huggingface.ipynb to make it cleaner and added some mulitthreading to help with the image downloads. Honestly not sure if this one works or not it seems it does now I need you to check this one

Gabefire:
Once this is improved we need to find the broken links if they were used at all. https://docs.labelbox.com/page/tutorials and we need to update the readme

tomislav-peharda (approved):

PR #1488, opened 2024-03-15
[PLT-431] add update method to attachments

attila-papai:
Add the ability to update data row attachments.
Also introduced the `attachment_name` attribute.

### DIFF
--- a/labelbox/schema/asset_attachment.py
+++ b/labelbox/schema/asset_attachment.py
@@ -12,6 +12,7 @@ class AssetAttachment(DbObject):
     Attributes:
         attachment_type (str): IMAGE, VIDEO, IMAGE_OVERLAY, HTML, RAW_TEXT, TEXT_URL, or PDF_URL. TEXT attachment type is deprecated.
         attachment_value (str): URL to an external file or a string of text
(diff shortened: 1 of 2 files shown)


mrobers1982 (approved):

labelbox/schema/asset_attachment.py
```diff
@@ -65,3 +66,29 @@ def delete(self) -> None:
                     id}
             }"""
         self.client.execute(query_str, {"attachment_id": self.uid})
+
+    def update(self,
+               name: Optional[str] = None,
+               type: Optional[str] = None,
+               value: Optional[str] = None):
+        """Updates an attachment on the data row."""
```
vbrodsky, line 74: I belive this goes into our online documentation on [readthedoc](https://readthedocs.org/projects/labelbox-python/) so we should not refer here to any internal gql stuff, just what the user should see

BTW we use sphinx to generate documentation and here is the markup https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html there is also a structured way to call out args etc some people have been doing

This is optional, we do not enforce this at the moment

vbrodsky (approved):
Berry good :)

PR #1505, opened 2024-03-26
[PLT-0] fix pypath / use of pytest conventions around fixtures

adrian-chang:
* Delete hacky use of pythonpath
* move fixtures to confpy files (fixture files)
* Move files around to data or integration (eg pip install labelbox or pip install labelbox[python])


.github/workflows/python-package.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           if [[ "${{github.base_ref}}" == "master" || "${{github.ref}}" == "refs/heads/master" ]]; then
             echo "LABELBOX_TEST_ENVIRON=prod" >> $GITHUB_ENV
           else
-            echo "LABELBOX_TEST_ENVIRON=staging" >> $GITHUB_ENV
+            echo "LABELBOX_TEST_ENVIRON=prod" >> $GITHUB_ENV
```
vbrodsky: why this change? I.e. if the branch is not master, shouldn't we assume staging?

adrian-chang: Can you explain the hard dependency of a branch against staging? 

EG SDK (branch) -> lb-api (STAGING) 

I don't see a particular reason that it needs to be synchronized, with exception of there's possibly a feature only in lb-api (STAGING) that the SDK (branch) needs to work against. Even then, why not just deploy lb-api first to prod first?

Our customers, also have 0 context of any lb-api (staging), and shouldn't, where in fact we almost expect / want them to be aware of this fact if they contribute via a branch. (not great)

The SDK also should be tested against prod in a branch because ultimately that's what customers are going to experience immediately after merging to develop. 

Allowing someone to merge into develop but fail, because something is in lb-api staging and not on prod, is not a good thing. (Nothing should be merged into develop without the lb-api changes in production first, where the existing workflow allows for that to happen because of the above tests).

If there's a strong desire to be able to run a branch against staging, we can have that run as a manual workflow, but having the default SDK (branch) -> lb-api (staging) be the final check before merging into develop is insufficent

vbrodsky (approved):

PR #1508, opened 2024-03-26
[PLT-383] Fix SDK integration tests.

sfendell-labelbox:
Video and text tests are failing because expected values do not match existing output, so I fixed those expected values. 

test_import_data_types_v2 relies on ordering of exported data, which exporter makes no promises about. This fix is required to merge https://github.com/Labelbox/python-monorepo/pull/1789.


adrian-chang (approved):

midhun-pm (approved):

sfendell-labelbox:
Video and text tests are failing locally but succeeding on staging...Not sure what's going on there, but I know it's unrelated to my exporter change so I'm just getting rid of that. Removing the dependency on export ordering is still necessary though

PR #1511, opened 2024-03-27
[PLT-0] Use fixture in a test instead of a global var

vbrodsky:


tomislav-peharda (approved):

adrian-chang:
@vbrodsky closing this until you are ready (multiple months old not sure if it's relevant)

PR #1513, opened 2024-03-28
[PLT-0] Prep Image for CI Tests

adrian-chang:
* Prep image to run CI tests


.github/workflows/release-container.yaml
```diff
@@ -40,4 +40,4 @@ jobs:
 
         tags: |
           ${{ env.CONTAINER_IMAGE }}:latest
-          ${{ env.CONTAINER_IMAGE }}:${{ github.event.release.tag_name }}
+          ${{ env.CONTAINER_IMAGE }}:${{ github.event.release.tag_name }}
```
sfendell-labelbox: Nit: Add the newline back to EOF here so there's no unnecessary diff

.github/workflows/testing-container.yaml
```diff
@@ -41,4 +41,4 @@ jobs:
 
         tags: |
           ${{ env.CONTAINER_IMAGE }}:develop
-          ${{ env.CONTAINER_IMAGE }}:${{ github.sha }}
+          ${{ env.CONTAINER_IMAGE }}:${{ github.sha }}
```
sfendell-labelbox: Nit: Add the newline back to EOF here so there's no unnecessary diff

Dockerfile
```diff
@@ -1,6 +1,8 @@
-FROM python:3.9-slim-bullseye
+#https://stackoverflow.com/questions/77364550/attributeerror-module-pkgutil-has-no-attribute-impimporter-did-you-mean
+#https://github.com/pyproj4/pyproj/issues/1321
+FROM python:3.8-slim-bullseye
```
sfendell-labelbox, line 3: Could you add a TODO + jira ticket here to upgrade this from python 3.8? It doesn't have to be a near term goal, but 3.8 will be EOL soon and I think we should track this

sfendell-labelbox (approved):

PR #1515, opened 2024-03-28
[PLT-X] Make export_v2 methods use streamable backend

mnoszczak:
Prepares export_v2 methods to use streamable backend under the hood

### DIFF
--- a/labelbox/schema/catalog.py
+++ b/labelbox/schema/catalog.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 from labelbox.orm.db_object import experimental
 from labelbox.schema.export_filters import CatalogExportFilters, build_filters
 
@@ -94,7 +94,7 @@ class Catalog:
                 filters: Union[CatalogExportFilters, Dict[str, List[str]],
                                None] = None,
                 params: Optional[CatalogExportParams] = None,
-                streamable: bool = False) -> tuple[Task, bool]:
+                streamable: bool = False) -> Tuple[Task, bool]:
 
         _params = params or CatalogExportParams({
             "attachments": False,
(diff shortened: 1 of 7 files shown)


labelbox/schema/catalog.py
```diff
@@ -23,7 +23,7 @@ def export_v2(
         task_name: Optional[str] = None,
         filters: Union[CatalogExportFilters, Dict[str, List[str]], None] = None,
         params: Optional[CatalogExportParams] = None,
-    ) -> Task:
+    ) -> Union[Task, ExportTask]:
```
midhun-pm, line 26: In this PR we return a union or a tuple in several places, is it possible to avoid that ? Here for instance I can't tell the difference between `Task` and `ExportTask`. Also we end up not using one of the two return values in other places.

mnoszczak: This PR modifies `ExportTask` so that it is a superset of `Task` and behaves similarly. We cannot know what is being returned until we toggle the flag for all customers. We'll be able to narrow down the type in future version (once flag is enabled for everyone).

labelbox/schema/export_task.py
```diff
@@ -467,7 +467,8 @@ class ExportTask:
     class ExportTaskException(Exception):
         """Raised when the task is not ready yet."""
 
-    def __init__(self, task: Task) -> None:
+    def __init__(self, task: Task, is_export_v2: bool = False) -> None:
+        self._is_export_v2 = is_export_v2
```
vbrodsky, line 471: I assume we need to explicitly pass it in because of the api feature flag that might change streamable to export_v2?

vbrodsky (approved):

labelbox/schema/export_task.py
mnoszczak: This value comes down from API though. The flow is as follows: 
1. New SDK version: Hey API I'm able to handle streamable backend
2. API: Ok let me check feature flag if I can serve you streamable export
3. SDK Receives streamable response: uses ExportTask internal (with is_export_v2=True) and tweaks its public API so it behaves just like Task class (result_url/errors_url)

PR #1520, opened 2024-03-29
[PLT-0] move to manual release

adrian-chang:
* Run prod tests before prod
* Update readme noting unstable (use tags)
* Remove main


sfendell-labelbox (approved):

Dockerfile
```diff
@@ -19,4 +19,5 @@ RUN pip install -r requirements.txt
 COPY . /usr/src/
 
 RUN python setup.py install
-CMD tox -e py -- tests/integration tests/data
\ No newline at end of file
+RUN tox -e no
```
vbrodsky, line 22: what does this line do

adrian-chang: Caches the tox requirements in the docker image so I can speed up tests, i'll tag you on the iac-codefresh item

PR #1526, opened 2024-04-02
[PLT-51] Add error url for import and upsert task.

sfendell-labelbox:
Leaving error url empty is confusing to users. Since we already print errors to the results file (and can't remove this behavior without breaking backwards compatibility), it can be reused for errors.


vbrodsky (approved):
good stuff

PR #1527, opened 2024-04-03
Vb/remove label data plt 37

vbrodsky:
Story: https://labelbox.atlassian.net/browse/PLT-37

**Label Validation Update:**

Modified the Label pydantic validator to automatically convert a dictionary into a GenericDataRowData object for user convenience. For the time being, usage of other *RowData classes is allowed, but a deprecation warning will be issued.

**Other Improvements:**

- Enum Support for Python 3.8: Implemented a workaround to support string enums in Python 3.8, as it does not support StrEnum. Notably, the Foo(string, Enum) approach is deprecated in Python 3.9.
- Integration Test Client Fix: Resolved an issue where the IntegrationTestClient was using an incorrect API key or failing to find the correct one.
- Test Logic Refactoring: Consolidated some repetitive test logic into shared fixtures for efficiency.

### DIFF
--- a/labelbox/data/annotation_types/data/generic_data_row_data.py
+++ b/labelbox/data/annotation_types/data/generic_data_row_data.py
@@ -1,0 +1,28 @@
+from typing import Callable, Literal, Optional
+
+from labelbox import pydantic_compat
+from labelbox.data.annotation_types.data.base_data import BaseData
+from labelbox.utils import _NoCoercionMixin
+
+
+class GenericDataRowData(BaseData, _NoCoercionMixin):
+    """Generic data row data
+    """
+    url: Optional[str] = None
+    class_name: Literal["GenericDataRowData"] = "GenericDataRowData"
+
+    def create_url(self, signer: Callable[[bytes], str]) -> None:
+        return None
+
+    @pydantic_compat.root_validator(pre=True)
+    def validate_one_datarow_key_present(cls, data):
+        keys = ['external_id', 'global_key', 'uid']
+        count = 0
+        for key in keys:
+            if data.get(key):
+                count += 1
+        if count < 1:
+            raise ValueError(f"Exactly one of {keys} must be present.")
+        if count > 1:
+            raise ValueError(f"Only one of {keys} can be present.")
+        return data
--- a/labelbox/data/annotation_types/data/generic_data_row_data.py
+++ b/labelbox/data/annotation_types/data/generic_data_row_data.py
@@ -11,8 +11,8 @@ class GenericDataRowData(BaseData, _NoCoercionMixin):
     url: Optional[str] = None
     class_name: Literal["GenericDataRowData"] = "GenericDataRowData"
 
-    def create_url(self, signer: Callable[[bytes], str]) -> None:
-        return None
+    def create_url(self, signer: Callable[[bytes], str]) -> Optional[str]:
+        return self.url
 
     @pydantic_compat.root_validator(pre=True)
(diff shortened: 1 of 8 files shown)


labelbox/data/annotation_types/data/generic_data_row_data.py
```diff
@@ -0,0 +1,28 @@
...
+from labelbox.utils import _NoCoercionMixin
+
+
+class GenericDataRowData(BaseData, _NoCoercionMixin):
+    """Generic data row data. This is replacing all other DataType passed into Label
+    """
+    url: Optional[str] = None
+    class_name: Literal["GenericDataRowData"] = "GenericDataRowData"
+
+    def create_url(self, signer: Callable[[bytes], str]) -> None:
```
vbrodsky: For backward compatibility, does not look like it is used

labelbox/data/annotation_types/data/generic_data_row_data.py
```diff
@@ -0,0 +1,28 @@
...
+from labelbox.utils import _NoCoercionMixin
+
+
+class GenericDataRowData(BaseData, _NoCoercionMixin):
+    """Generic data row data. This is replacing all other DataType passed into Label
+    """
+    url: Optional[str] = None
+    class_name: Literal["GenericDataRowData"] = "GenericDataRowData"
+
+    def create_url(self, signer: Callable[[bytes], str]) -> Optional[str]:
```
vbrodsky, line 14: For backward compatibility with legacy classes, currently not used

labelbox/data/annotation_types/data/generic_data_row_data.py
```diff
@@ -0,0 +1,28 @@
...
+        return self.url
+
+    @pydantic_compat.root_validator(pre=True)
+    def validate_one_datarow_key_present(cls, data):
+        keys = ['external_id', 'global_key', 'uid']
+        count = 0
```
sfendell-labelbox: `count = sum([key in data for key in keys])`

tests/data/annotation_import/test_generic_data_types.py
```diff
@@ -0,0 +1,336 @@
...
+    assert parsed_t.hour is not None
+    assert parsed_t.minute is not None
+    assert parsed_t.second is not None
+
+
+def to_pascal_case(name: str) -> str:
+    return "".join([word.capitalize() for word in name.split("_")])
+
+
+def set_project_media_type_from_data_type(project, data_type_class):
```
sfendell-labelbox: These exist in test_data_types already, could we move them to helpers and deduplicate?

tests/data/annotation_import/test_generic_data_types.py
```diff
@@ -0,0 +1,336 @@
...
+    label_import.wait_until_done()
+
+    assert label_import.state == AnnotationImportState.FINISHED
+    assert len(label_import.errors) == 0
+
+    # TODO need to migrate project to the new BATCH mode and change this code
+    # to be similar to tests/integration/test_task_queue.py
+
+    result = export_v2_test_helpers.run_project_export_v2_task(project)
+    find_data_row = lambda dr: dr['data_row']['id'] == data_row.uid
```
sfendell-labelbox: Can we avoid redefining this filter function? We have 3 copies of it between here and test_data_type. Something in helpers like:

```
def find_data_row_filter(data_row):
    return lambda dr: dr['data_row']['id'] == data_row.uid
```

sfendell-labelbox (approved):

PR #1543, opened 2024-04-05
[PLT-582] add example and documentation for custom embeddings

mrobers1982:
https://labelbox.atlassian.net/browse/PLT-582

I chose to not update the existing file - https://github.com/Labelbox/labelbox-python/blob/develop/examples/basics/custom_embeddings.ipynb to not break the existing public-facing documentation.

We can remove the above file once `advlib` is deprecated and the public-facing documentation is updated.

https://docs.labelbox.com/reference/custom-embeddings

### DIFF
--- a/examples/basics/custom_embeddings_sdk.ipynb
+++ b/examples/basics/custom_embeddings_sdk.ipynb
@@ -1,0 +1,269 @@
+{
+ "nbformat": 4,
+ "nbformat_minor": 0,
+ "metadata": {},
+ "cells": [
+  {
+   "metadata": {},
+   "source": [
+    "<td>\n",
+    "   <a target=\"_blank\" href=\"https://labelbox.com\" ><img src=\"https://labelbox.com/blog/content/images/2021/02/logo-v4.svg\" width=256/></a>\n",
+    "</td>"
+   ],
+   "cell_type": "markdown"
+  },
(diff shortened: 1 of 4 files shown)


mrobers1982:
~~I likely need to add documentation here as well - https://labelbox-python.readthedocs.io/en/latest/.~~

This should be done now, let me know if it's incorrect.

examples/basics/custom_embeddings_sdk.ipynb
```diff
@@ -0,0 +1,269 @@
+{
```
midhun-pm, line 1: It would have been nice if the example helped motivate the need for custom embeddings.

mrobers1982: Is this not sufficient (see top-most paragraph)?

https://github.com/Labelbox/labelbox-python/blob/mrobers1982/PLT_582/examples/basics/custom_embeddings_sdk.ipynb

mrobers1982: There is also customer-facing documentation on our website - https://docs.labelbox.com/docs/embeddings.

It ultimately links here, not sure we need to duplicate everywhere.

sfendell-labelbox (approved):

PR #1549, opened 2024-04-08
remove dateutil max version constraint

vincent-grosbois:
remove dateutil max version constraint

otherwise labelbox will conflict with other packages now that python-dateutil is widespread

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ pydantic>=1.8
 pygeotile
 pyproj
 pytest-xdist
-python-dateutil>=2.8.2,<2.9.0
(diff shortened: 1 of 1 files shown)


adrian-chang:
Hey there!

Appreciate the contribution. 

I think we actually found that this dependency, is not in use anymore within the Labelbox package (+ also we made some changes to not need to use requirements.txt anymore). Feel free though to follow up with any comments if you disagree, otherwise I'll plan on closing the PR soon.

adrian-chang:
Closing per my note above. Feel free to follow up if there's an issue.

PR #1552, opened 2024-04-08
[PLT-0] Override RTD build commands

adrian-chang:
* https://docs.readthedocs.io/en/stable/config-file/v2.html#build-commands
* Override the build commands section to match building docs locally


vbrodsky (commented):
Can you add some more info, i.e. why did we remove python version and why are we using latest rye instead of a specific version?

adrian-chang:
Sure that's done, updated the summary

vbrodsky (approved):

PR #1584, opened 2024-04-30
[PLT-817] Move to buffered reader

adrian-chang:
* Buffer the file locally instead of trying to parse chunks in flight, then pass row by row to the convertor

### DIFF
--- a/libs/labelbox/src/labelbox/schema/export_task.py
+++ b/libs/labelbox/src/labelbox/schema/export_task.py
@@ -415,18 +415,18 @@ class _MultiGCSFileReader(_Reader):  # pylint: disable=too-few-public-methods
 
 
 @dataclass
-class BufferedJSONConverterOutput:
+class BufferedJsonConverterOutput:
     """Output with the JSON object"""
     json: Any
 
 
-class _BufferedJSONConverter(Converter[BufferedJSONConverterOutput]):
+class _BufferedJsonConverter(Converter[BufferedJsonConverterOutput]):
     """Converts JSON data in a buffered manner
     """
     def convert(
         self, input_args: Converter.ConverterInputArgs
-    ) -> Iterator[BufferedJSONConverterOutput]:
-        yield BufferedJSONConverterOutput(json=json.loads(input_args.raw_data))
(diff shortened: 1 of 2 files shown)


libs/labelbox/src/labelbox/schema/export_task.py
```diff
@@ -106,7 +108,14 @@ class JsonConverterOutput:
 
 
 class JsonConverter(Converter[JsonConverterOutput]):  # pylint: disable=too-few-public-methods
-    """Converts JSON data."""
+    """Converts JSON data.
+    
+    Deprecated: This converter is deprecated and will be removed in a future release.
+    """
```
adrian-chang, line 116: The _find_json_object_offsets method leads to coding errors (implemented incorrectly), trying to parse chunks for json will lead to errors, buffer is guaranteed to be accurate but possibly slower, but at least simpler!

libs/labelbox/src/labelbox/schema/export_task.py
```diff
@@ -397,6 +414,52 @@ def read(self) -> Iterator[Tuple[_MetadataFileInfo, str]]:
             result = self._retrieval_strategy.get_next_chunk()
 
 
+@dataclass
+class BufferedJSONConvertorOutput:
```
mnoszczak: typo?

libs/labelbox/src/labelbox/schema/export_task.py
```diff
@@ -397,6 +414,52 @@ def read(self) -> Iterator[Tuple[_MetadataFileInfo, str]]:
...
+
+        with tempfile.NamedTemporaryFile(mode='w+') as temp_file:
+            result = self._retrieval_strategy.get_next_chunk()
+            while result:
+                file_info, raw_data = result
+                temp_file.seek(file_info.offsets.start)
+                temp_file.write(raw_data)
+                result = self._retrieval_strategy.get_next_chunk()
+            # reset the line to starting to read
+            temp_file.seek(0)
```
midhun-pm: I think it's better to close the file before we start reading from it again (so that any buffers are flushed)

midhun-pm (approved):

PR #1589, opened 2024-05-03
Vb/chat eval ontology

vbrodsky:
Story: https://labelbox.atlassian.net/browse/PLT-774

This PR adds support for creating ontologies for multi-modal chat evaluation.

I have introduced distinct methods for both create_ontology and create_monotony_from_feature_schemas. These methods are designed to shield users from the complexities of the implementation details.

### DIFF
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -1228,6 +1228,33 @@ class Client:
         return Entity.Ontology(self, res['upsertOntology'])
 
     def create_model_chat_evaluation_ontology(self, name, normalized):
+        """
+        Creates a model chat evalutation ontology from normalized data
+            >>> normalized = {"tools" : [{'tool': 'model-output-single-selection', 'name': 'model output single selection', 'color': '#ff0000',}, 
+                                         {'tool': 'model-output-multi-selection', 'name': 'model output multi selection', 'color': '#00ff00',}, 
+                                         {'tool': 'model-output-ranking', 'name': 'model output multi ranking', 'color': '#0000ff',}]
+                             }
+            >>> ontology = client.create_ontology("ontology-name", normalized)
+
+        Or use the ontology builder
+            >>> ontology_builder = OntologyBuilder(tools=[
+                Tool(tool=Tool.Type.MODEL_OUTPUT_SINGLE_SELECTION,
+                    name="model output single selection"),
+                Tool(tool=Tool.Type.MODEL_OUTPUT_MULTI_SELECTION,
+                    name="model output multi selection"),
+                Tool(tool=Tool.Type.MODEL_OUTPUT_MULTI_RANKING,
+                    name="model output multi ranking"),
+            ],)
+
+            >>> ontology = client.create_model_chat_evaluation_ontology("Multi-chat ontology", ontology_builder.asdict())
+
+        Args:
+            name (str): Name of the ontology
+            normalized (dict): A normalized ontology payload. See above for details.
+        Returns:
+            The created Ontology
+        """
+
         return self.create_ontology(
             name=name,
             normalized=normalized,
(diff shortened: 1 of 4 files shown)


libs/labelbox/src/labelbox/client.py
```diff
@@ -1166,7 +1172,11 @@ def get_unused_feature_schemas(self, after: str = None) -> List[str]:
                 "Failed to get unused feature schemas, message: " +
                 str(response.json()['message']))
 
-    def create_ontology(self, name, normalized, media_type=None) -> Ontology:
+    def create_ontology(self,
```
sfendell-labelbox, line 1175: Needs a docstring for editor_task_type.

libs/labelbox/src/labelbox/client.py
```diff
@@ -1204,9 +1221,65 @@ def create_ontology(self, name, normalized, media_type=None) -> Ontology:
...
+        """
+
+        return self.create_ontology(
+            name=name,
+            normalized=normalized,
+            media_type=MediaType.Conversational,
+            editor_task_type=EditorTaskType.ModelChatEvaluation)
+
+    def create_model_chat_evaluation_ontology_from_feature_schemas(
+            self, name, feature_schema_ids):
```
sfendell-labelbox, line 1265: Can this function have its inputs + return types typed? I know we have a bunch of untyped methods but I think we should try to keep all new functions typed (admittedly we don't have a style guide to refer to or anything for this, I'd just really prefer it).

libs/labelbox/tests/integration/test_model_chat_evaluation_ontology.py
```diff
@@ -0,0 +1,86 @@
...
+            assert tool.schema_id
+            assert tool.feature_schema_id
+
+    finally:
+        if ontology:
+            client.delete_unused_ontology(ontology.uid)
+
+
+def test_ontology_create_feature_schema(client, rand_gen):
+    try:
```
sfendell-labelbox, line 32: Can any of this be pulled out into a fixture? I think that's a better place to create/cleanup this kind of test data instead of this try/finally structure.

libs/labelbox/src/labelbox/client.py
```diff
@@ -938,10 +939,12 @@ def rootSchemaPayloadToFeatureSchema(client, payload):
...
-    def create_ontology_from_feature_schemas(self,
-                                             name,
-                                             feature_schema_ids,
-                                             media_type=None) -> Ontology:
+    def create_ontology_from_feature_schemas(
+            self,
+            name,
+            feature_schema_ids,
+            media_type: MediaType = None,
+            editor_task_type: EditorTaskType = None) -> Ontology:
```
sfendell-labelbox, line 947: Docstring here too.

libs/labelbox/src/labelbox/client.py
vbrodsky: I will add thank you!

libs/labelbox/src/labelbox/client.py
vbrodsky: yes

libs/labelbox/tests/integration/test_model_chat_evaluation_ontology.py
vbrodsky: not sure because the test itself is testing creation of ontology, but possibly in a follow up PRs where I add project and linking ontology to project

libs/labelbox/src/labelbox/client.py
Tim-Kerr: @vbrodsky We'd really like to hide this `editor_task_type` from the customer. One idea we had was to add a new `media_type` called `"model-chat-evaluation"` which is a special media type that maps to `MediaType.Conversational` and an editor task type with value `model-chat-evaluation`. What does you think?

cc @lgluszek

vbrodsky:
Will start a new PR that includes project and editor setup

PR #1590, opened 2024-05-03
[DOCS-450] upsert and attachment new methods

ovalle15:
- I removed redundancies 
- I removed methods that are not efficient
- I added new methods to attachments and create data rows


adrian-chang (approved):

mina-ebeid (approved):
This looks good to go!

PR #1591, opened 2024-05-03
[PLT-773] Allow users to create configs and associate them with projects through SDK.

sfendell-labelbox:
Users can create and delete model configs (tying inference params to a model id), and associate model configs with projects.


libs/labelbox/src/labelbox/client.py
```diff
@@ -590,6 +590,55 @@ def _create(self, db_object_type, data):
...
+        query = """mutation DeleteModelConfigPyApi($id ID!)  {
+                    deleteModelConfig(input: {id: $id}) {
+                        id
+                    }
+                }"""
+        params = {
+            "id": id
+        }
+        result = self.execute(query, params)
+        return result['deleteModelConfig']['success']
```
midhun-pm, line 647: Should we handle errors ?

tomislav-peharda: I think they are handled inside `execute`

tomislav-peharda:
not my domain, but interfaces seem fine.

libs/labelbox/src/labelbox/client.py
```diff
@@ -590,6 +590,55 @@ def _create(self, db_object_type, data):
...
+                        name
+                    }
+                }"""
+        params = {
+            "modelId": model_id,
+            "inferenceParams": inference_params,
+            "name": name
+        }
+        result = self.execute(query, params)
+        return result['createModelConfig']['modelConfigId']
```
attila-papai: you're not seem to be selecting "modelConfigId" in the `createModelConfig` response, so I suppose we get a KeyError here?
Also, I believe it should be simply `id`, and since that's all you want to return here, you can also simplify the selection of response fields to `id` only.

libs/labelbox/src/labelbox/client.py
attila-papai: similar to above, `success` is not being selected as a field in the response of `deleteModelConfig` but `id`, which is not even returned

sfendell-labelbox: This one returns a bool, not an ID.

libs/labelbox/src/labelbox/client.py
sfendell-labelbox: I think I need the query key here (`createModelConfig`), at least every other example we have of indexing into a client execute result uses it (I'm having trouble actually calling these queries right now). Agreed about the ID field, that's fixed.

libs/labelbox/src/labelbox/client.py
sfendell-labelbox: And yeah, execute retries errors considered retriable, and errors besides those should be raised to the client.

sfendell-labelbox: Sorry, I misunderstood your comment here, fixing these queries now

libs/labelbox/src/labelbox/schema/model_config.py
```diff
@@ -0,0 +1,15 @@
+from labelbox.orm.db_object import DbObject
+from labelbox.orm.model import Field
+
+
+class ModelConfig(DbObject):
+    """ A ModelConfig represents a set of inference params configured for a model
+
+    Attributes: 
+        inference_params (JSON): Dict of inference params
+        model_id (str): ID of the model to configure
```
vbrodsky, line 10: Add name to Attributes?

libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -1226,6 +1249,50 @@ def get_queue_mode(self) -> "QueueMode":
...
+                }"""
+
+        params = {
+            "projectId": self.uid,
+            "modelConfigId": model_config_id,
+        }
+        result = self.client.execute(query, params)
+        return result["createProjectModelConfig"]["projectModelConfigId"]
+
+    def remove_project_model_config(self, project_model_config_id: str) -> bool:
```
vbrodsky: why not `delete_...` as consistent with other method names?

libs/labelbox/src/labelbox/schema/project_model_config.py
```diff
@@ -0,0 +1,34 @@
...
+class ProjectModelConfig(DbObject):
+    """ A ProjectModelConfig represents an association between a project and a single model config.
+
+    Attributes: 
+        project_id (str): ID of project to associate
+        model_config_id (str): ID of the model configuration
+        model_config (ModelConfig): Configuration for model
+    """
+
+    project_id = Field.String("project_id", "projectId")
```
vbrodsky, line 14: BTW second param is only needed if, camelCased, it differs from the first - do not need to update this PR, FYI

libs/labelbox/tests/integration/test_model_config.py
```diff
@@ -0,0 +1,12 @@
...
+def test_create_model_config(client, valid_model_id):
+    model_config = client.create_model_config("model_config", valid_model_id, {"param": "value"})
+    assert model_config.inference_params["param"] == "value"
+    assert model_config.name == "model_config"
+    assert model_config.model_id == valid_model_id
+
+
+def test_delete_model_config(client, valid_model_id):
+    model_config_id = client.create_model_config("model_config", valid_model_id, {"param": "value"})
+    assert(client.delete_model_config(model_config_id.uid))
```
vbrodsky, line 12: is there a way to retrieve it back and make sure it does not exist

libs/labelbox/tests/integration/test_project_model_config.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+def test_add_single_model_config(configured_project, model_config):
+    project_model_config_id = configured_project.add_model_config(model_config.uid)
+
+    assert set(config.uid for config in configured_project.project_model_configs()) == set([project_model_config_id])
+
+    configured_project.remove_project_model_config(project_model_config_id)
```
vbrodsky: assertion is missing

libs/labelbox/tests/integration/test_project_model_config.py
```diff
@@ -0,0 +1,26 @@
...
+    assert set(config.uid for config in configured_project.project_model_configs()) == expected_model_configs
+
+    [configured_project.remove_project_model_config(project_model_config_id) for project_model_config_id in expected_model_configs]
+
+
+def test_remove_model_config_from_project(configured_project, model_config):
+    assert configured_project.remove_project_model_config(configured_project.add_model_config(model_config.uid))
+
+def test_delete_project_model_config(configured_project, model_config):
+    assert configured_project.remove_project_model_config(configured_project.add_model_config(model_config.uid))
```
vbrodsky: would be nice to retrieve back configs to make sure they have been indeed removed

vbrodsky (approved):

libs/labelbox/tests/integration/test_project_model_config.py
sfendell-labelbox: Sure, done.

libs/labelbox/tests/integration/test_model_config.py
sfendell-labelbox: I'll add one.

libs/labelbox/tests/integration/test_project_model_config.py
sfendell-labelbox: users can see project model configs are gone with the project.project_model_configs command. I added get_model_config to the client

libs/labelbox/src/labelbox/schema/project.py
sfendell-labelbox: Fixed.

libs/labelbox/src/labelbox/schema/model_config.py
sfendell-labelbox: Done.

libs/labelbox/tests/integration/test_project_model_config.py
sfendell-labelbox: Actually not adding this in this PR, but will do in a followup.

vbrodsky:
The only failing tests are for model_run upsert_data_rows, cause not related to this code, see https://labelbox.slack.com/archives/C040H094D4Y/p1715210472609159?thread_ts=1715185323.737509&cid=C040H094D4Y

libs/labelbox/tests/integration/test_model_config.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+def test_create_model_config(client, valid_model_id):
+    model_config = client.create_model_config("model_config", valid_model_id, {"param": "value"})
```
attila-papai, line 4: I don't see a cleanup of this model config, we should make sure it's deleted after the test case runs

libs/labelbox/tests/integration/test_model_config.py
```diff
@@ -0,0 +1,12 @@
...
+    assert model_config.inference_params["param"] == "value"
+    assert model_config.name == "model_config"
+    assert model_config.model_id == valid_model_id
+
+
+def test_delete_model_config(client, valid_model_id):
+    model_config_id = client.create_model_config("model_config", valid_model_id, {"param": "value"})
```
attila-papai, line 11: it should be rather called `model_config`

libs/labelbox/tests/integration/test_project_model_config.py
```diff
@@ -0,0 +1,25 @@
...
+def test_add_single_model_config(configured_project, model_config):
+    project_model_config_id = configured_project.add_model_config(model_config.uid)
+
+    assert set(config.uid for config in configured_project.project_model_configs()) == set([project_model_config_id])
+
+    assert configured_project.delete_project_model_config(project_model_config_id)
+
+
+def test_add_multiple_model_config(client, rand_gen, configured_project, model_config, valid_model_id):
+    second_model_config = client.create_model_config(rand_gen(str), valid_model_id, {"param": "value"})
```
attila-papai, line 12: `second_model_config` is not cleaned up

PR #1592, opened 2024-05-07
Gu/fix test v2

Gabefire:


libs/labelbox/src/labelbox/data/serialization/coco/annotation.py
```diff
@@ -26,21 +26,12 @@ def rle_decoding(rle_arr: List[int], w: int, h: int) -> np.ndarray:
     return mask.reshape((w, h)).T
 
 
-def get_annotation_lookup(
-    annotations: List[Union[ClassificationAnnotation, ObjectAnnotation,
-                            VideoMaskAnnotation, ScalarMetric,
-                            ConfusionMatrixMetric, RelationshipAnnotation]]
-) -> defaultdict[Any, list]:
-    """Get annotations from Label.annotations objects
```
mihhail-m, line 34: leaving a comment for a function is probably not a bad idea

Gabefire: Added!

libs/labelbox/src/labelbox/data/serialization/coco/annotation.py
```diff
@@ -26,11 +26,7 @@ def rle_decoding(rle_arr: List[int], w: int, h: int) -> np.ndarray:
     return mask.reshape((w, h)).T
 
 
-def get_annotation_lookup(
-    annotations: List[Union[ClassificationAnnotation, ObjectAnnotation,
-                            VideoMaskAnnotation, ScalarMetric,
-                            ConfusionMatrixMetric, RelationshipAnnotation]]
-) -> defaultdict[Any, list]:
```
attila-papai, line 33: why did you remove the types?

libs/labelbox/src/labelbox/data/serialization/coco/annotation.py
```diff
@@ -39,8 +35,8 @@ def get_annotation_lookup(
     annotation_lookup = defaultdict(list)
     for annotation in annotations:
         annotation_lookup[getattr(annotation, 'image_id', None) or
-                          getattr(annotation, 'name', None)].append(annotation)
-    return annotation_lookup
+                          getattr(annotation, 'name')].append(annotation)
+    return annotation_lookup
```
attila-papai: what does removing `None` exactly fix and how?

libs/labelbox/src/labelbox/data/serialization/coco/annotation.py
Gabefire: It is a revert from a pr I did. I was the one who actually added the typing and doc string. I cleaned up the coco converter across the board and then added a depercation message but this function is causing trouble so I reverted it for some reason with python 3.8 unit test does not like line 38 having a default value. (I'm not sure if it is a problem with the test or just this python 3.8) I can add back in the typing but this entire module will likely be removed in the next few weeks. I added the deprecation messages it has a lot of other problems besides this.

libs/labelbox/src/labelbox/data/serialization/coco/annotation.py
Gabefire: Im honestly not sure it was how the original author had it. I added the None in as a default value because some of our label.annotations type don't have a name resulting in an error. But that broke a test. If we wanted it back I can look at the test

Gabefire: I figured it out the dictionary key was resolving weird since it could not pick up a value in the or state. I added logic to give a default value if name or image is not available and the test for python 3.8 works

attila-papai (approved):

PR #1594, opened 2024-05-07
[PLT-772] SDK Ontology, project, editor support for chat evaluation

vbrodsky:
Update methods that create ontology and project as well as setup editor and link ontology to project

Approach:

- use editor_task_type to identify chat evaluation ontologies and projects but hide this (internal) attribute from the user. Instead use Project ontology_kind and set it based on the editor_task_type
- implemented intelligent workarounds arounds, as necessary, to work around of limiting assumptions of our homegrown sdk graphql framework


libs/labelbox/src/labelbox/client.py
```diff
@@ -563,14 +565,16 @@ def get_labeling_frontends(self, where=None) -> List[LabelingFrontend]:
         """
         return self._get_all(Entity.LabelingFrontend, where)
 
-    def _create(self, db_object_type, data):
+    def _create(self, db_object_type, data, extra_params={}):
```
sfendell-labelbox, line 568: Could this extra_params arg just be **kwargs?

libs/labelbox/tests/integration/test_chat_evaluation_ontology_project.py
```diff
@@ -0,0 +1,105 @@
...
+    # here we are essentially testing the ontology creation which is a fixture
+    assert ontology
+    assert ontology.name
+    assert len(ontology.tools()) == 3
+    for tool in ontology.tools():
+        assert tool.schema_id
+        assert tool.feature_schema_id
+
+    project = chat_evaluation_project
+    # editor = list(
```
sfendell-labelbox: Nit: Delete?

libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -131,6 +133,9 @@ class Project(DbObject, Updateable, Deletable):
     #
     _wait_processing_max_seconds = 3600
 
+    def is_chat_evalution(self) -> bool:
```
sfendell-labelbox: Typo: `is_chat_evalution` -> `is_chat_evaluation`

libs/labelbox/src/labelbox/client.py
```diff
@@ -728,12 +743,46 @@ def create_project(self, **kwargs) -> Project:
...
+            Field.Boolean("append_to_existing_dataset"):
+                params.pop("append_to_existing_dataset", None),
+            Field.Int("data_row_count"):
+                params.pop("data_row_count", None),
+        }
+        extra_params = {k: v for k, v in extra_params.items() if v is not None}
+
+        return self._create(Entity.Project, params, extra_params)
+
+    def create_model_evalution_project(self,
```
sfendell-labelbox: `evalution` -> `evaluation`

libs/labelbox/src/labelbox/client.py
```diff
@@ -579,12 +583,14 @@ def _create(self, db_object_type, data):
         """
         # Convert string attribute names to Field or Relationship objects.
         # Also convert Labelbox object values to their UIDs.
+
```
sfendell-labelbox: Super nit: Revert unnecessary whitespace change

libs/labelbox/src/labelbox/client.py
vbrodsky: I understand that kwargs typically allows for passing any general pair of name and value parameters in bulk. However, in this instance, I specifically need a list of GraphQL filters. If I use kwargs here, I would need to structure my data as {extra_params: {...}}, which seems unnecessary. What do you think?

libs/labelbox/tests/integration/test_chat_evaluation_ontology_project.py
vbrodsky: yep thx

libs/labelbox/src/labelbox/client.py
```diff
@@ -1184,26 +1254,43 @@ def create_ontology(self, name, normalized, media_type=None) -> Ontology:
             name (str): Name of the ontology
             normalized (dict): A normalized ontology payload. See above for details.
             media_type (MediaType or None): Media type of a new ontology
+            ontology_kind (OntologyKind or None): set to OntologyKind.ModelEvaluation if the ontology is for chat evaluation,
+                leave as None otherwise.
+
         Returns:
             The created Ontology
+
+        NOTE caller of this method is expected to set media_type to Conversational if ontology_kind is ModelEvaluation
```
sfendell-labelbox, line 1263: I understand having the separate arguments for media_type and ontology_kind, but could we solve this by having ontology_kind read first before media_type? Like: If ontology_kind is set to ModelEvaluation, media_type is set to conversational, otherwise media_type is just the arg.

sfendell-labelbox: Also, if you disagree with that comment and this note remains relevant, can we move it to the docstring? (But I think it's better to refactor and delete it)

libs/labelbox/src/labelbox/client.py
sfendell-labelbox: Ok yeah, it's fine as is

sfendell-labelbox (commented):
Looks good to me overall, I'll stamp after comments get addressed

libs/labelbox/src/labelbox/client.py
vbrodsky: Let me ping you to understand and discuss your concern better

sfendell-labelbox (approved):

PR #1595, opened 2024-05-07
[PLT-0] Non breaking change

adrian-chang:
Set default in method body instead of parameter


vbrodsky (approved):

adrian-chang:
Merging, issue is related to MF BE (not SDK code change).

PR #1597, opened 2024-05-09
[DOC-455] Example on how to setup the html editor with lb-html-api

ovalle15:
- User needs to be on editor-html-api FF


Gabefire:
Closing this for right now since Chun Pong does not want this to be public at all. I will keep the branch open though to add if they do make it public

PR #1598, opened 2024-05-09
[PLT-0] Do not retry 422 errors

vbrodsky:
QA discovered this issue when trying to create a model config with an invalid model id. In this case, the model service return http 422 as follows:
```
[{'message': 'Validation Error', 'locations': [{'line': 2, 'column': 21}], 'extensions': {'code': 'INTERNAL_SERVER_ERROR', 'exception': {'name': 'ApiError', 'url': 'http://lb-model-service-primary/api/v1/model-configs', 'status': 422, 'statusText': 'Unprocessable Entity', 'body': {'message': '1 validation error for Request body -> modelId value is not a valid uuid (type=type_error.uuid)'}, 'request': {'method': 'POST', 'url': '/api/v1/model-configs', 'body': {'modelId': 'invalid_valid_model_id', 'inferenceParams': {'param': 'value'}, 'organizationId': 'cl4ywm5aw00ke077lh4mrgpw5', 'name': 'model_config'}, 'mediaType': 'application/json', 'errors': {'422': 'Validation Error'}}}}, 'path': ['createModelConfig']}]
```

This PR prevents unnecessary retries of http 422 errors. 

See https://labelbox.atlassian.net/browse/PLT-773?focusedCommentId=201400 for more

### DIFF
--- a/.github/workflows/python-package-develop.yml
+++ b/.github/workflows/python-package-develop.yml
@@ -45,9 +45,6 @@ jobs:
     environment: 
       name: Test-PyPI
       url: 'https://test.pypi.org/p/labelbox-test'
-    permissions:
-      # IMPORTANT: this permission is mandatory for trusted publishing
-      id-token: write
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


adrian-chang (approved):

.github/workflows/python-package-develop.yml
```diff
@@ -79,6 +76,9 @@ jobs:
   test-container:
     runs-on: ubuntu-latest    
     needs: ['build']
+    permissions:
```
adrian-chang, line 79: Val this should be moved from the previous lines,

Need to add the permission I mentioned on slack here as an addition

adrian-chang (changes requested):

vbrodsky:
Done in another PR

PR #1599, opened 2024-05-10
Improve usability of create_project for chat evaluation and add basic validations

vbrodsky:
This PR addresses several usability issues and minor bugs identified during the initial QA of SDK support for MOE:

Issue: Creating a model configuration with an invalid model ID resulted in retries for 120 seconds before an error was reported.
**Fix**: We have now stopped retrying for HTTP 422 errors.
Issue: The options available during project creation were unclear, leading to end-user confusion and bugs.
**Fix**: I have added overloaded functions and docstring examples to clarify the two scenarios—creating a new project or appending to an existing dataset.
Improvement: Added basic validation for parameters to prevent confusing API error messages.

### DIFF
--- a/.github/workflows/python-package-develop.yml
+++ b/.github/workflows/python-package-develop.yml
@@ -10,6 +10,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  id-token: write
+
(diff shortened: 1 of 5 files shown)


adrian-chang (approved):
Should we write a unit test? Integration test is fine but of course unit is simpler ...

PR #1604, opened 2024-05-11
[PLT-1205] Add user group management

adrian-chang:
* Add user group management via SDK
* Create, Update, Delete Groups
* Also add projects to groups, remove projects from groups


libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+                                    UserGroupUser(id=member["id"],
+                                                  email=member["email"])
+                                    for member in group["members"]["nodes"]
+                                },
+                                projects={
+                                    UserGroupProject(id=project["id"],
+                                                     name=project["name"])
+                                    for project in group["projects"]["nodes"]
+                                })
+            nextCursor = userGroups.get("nextCursor", None)
```
sfendell-labelbox: Nit: 'None' argument is redundant here, get defaults to None

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+                        }
+                        totalCount
+                    }
+                }
+            }        
+        """
+        params = {
+            "id": self.id,
+        }
+        result = self.client.execute(query, params)
```
sfendell-labelbox, line 194: Handle execute returning None here.

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+            "projectIds": [
+                project.id if hasattr(project, 'id') else project.uid
+                for project in self.projects
+            ],
+            "userIds": [
+                user.id if hasattr(user, 'id') else user.uid
+                for user in self.users
+            ]
+        }
+        self.client.execute(query, params)
```
sfendell-labelbox: Handle execute returning None here.

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+        """
+        query = """
+        mutation DeleteUserGroupPyApi($id: ID!) {
+            deleteUserGroup(where: {id: $id}) {
+                success
+            }
+        }
+        """
+        params = {"id": self.id}
+        result = self.client.execute(query, params)
```
sfendell-labelbox, line 349: Handle execute returning None here e.g. if the group with this ID doesn't exist).

libs/labelbox/tests/integration/schema/test_user_group.py
```diff
@@ -0,0 +1,114 @@
+import pytest
+import faker
+from labelbox import Client
+from labelbox.schema.user_group import UserGroup, UserGroupColor, UserGroupUser, UserGroupProject
+
+data = faker.Faker()
+
+
+@pytest.fixture
+def client(client):
```
sfendell-labelbox: I think this text fixture affects the client fixture in others. But even if it doesn't, I still think it's better to just instantiate a new client here instead of editing one from another test module. Chaining test modules like this quickly gets very very hard to reason about for a reader, and we intentionally don't hardcode the endpoints in the original test module to point to an environment so I don't think we should here either.

libs/labelbox/tests/integration/schema/test_user_group.py
```diff
@@ -0,0 +1,114 @@
...
+        user_group.update()
+
+        # Verify that the user is added to the group
+        assert users[0] in user_group.users
+        assert projects[0] in user_group.projects
+
+        user_group.delete()
+
+
+if __name__ == "__main__":
```
sfendell-labelbox, line 88: Remove this?

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+    _name: str = None
+    _color: UserGroupColor = None
+    _users: Set[Union[UserGroupUser, User]] = None
+    _projects: Set[Union[UserGroupProject, Project]] = None
+    _client: Client
+
+    def __init__(self,
+                 client: Client,
+                 reload=True,
+                 **kwargs: UserGroupParameters):
```
sfendell-labelbox: I think we should use default arguments instead of kwargs here. kwargs is mostly useful when you're passing in some set of optional arguments to some other API, and you don't want to copy/paste its typings and arguments to the caller function. Having a typed kwargs where each argument defaults to something is a roundabout way of just creating default named args.

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+        _users (Set[Union[UserGroupUser, User]]): The set of user IDs in the user group.
+        _projects (Set[Union[UserGroupProject, Project]]): The set of project IDs in the user group.
+        _client (Client): The Labelbox client.
+    """
+    _id: str = None
+    _name: str = None
+    _color: UserGroupColor = None
+    _users: Set[Union[UserGroupUser, User]] = None
+    _projects: Set[Union[UserGroupProject, Project]] = None
+    _client: Client
```
sfendell-labelbox: All of these that can be None should be typed Optional

libs/labelbox/src/labelbox/schema/user_group.py
sfendell-labelbox: In fact you can already see some of the cost here where you have to effectively set these default values in 2 places now, one when you specify the default in kwargs.get and one when you specify the default instance var values.

libs/labelbox/src/labelbox/schema/user_group.py
sfendell-labelbox: Wait, these can't be none because their value is set in init... But if they can't be none they shouldn't be set as such here. If you change the kwargs to named args with defaults I think this all gets fixed.

libs/labelbox/src/labelbox/pydantic_compat.py
```diff
@@ -31,4 +31,4 @@ def pydantic_import(class_name, sub_module_path: Optional[str] = None):
 conint = pydantic_import("conint")
 conlist = pydantic_import("conlist")
 constr = pydantic_import("constr")
-confloat = pydantic_import("confloat")
+confloat = pydantic_import("confloat")
```
sfendell-labelbox: Nit: Revert changes to this file entirely since all you did was remove the terminating newline

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+
+    def __init__(self,
+                 client: Client,
+                 reload=True,
```
sfendell-labelbox: Why do we need this argument? We use this class in 1 place in the code and set this to false.

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+                                    for member in group["members"]["nodes"]
+                                },
+                                projects={
+                                    UserGroupProject(id=project["id"],
+                                                     name=project["name"])
+                                    for project in group["projects"]["nodes"]
+                                })
+            nextCursor = userGroups.get("nextCursor", None)
+            # this doesn't seem to be used right now
+            if nextCursor is None:
```
sfendell-labelbox: Nit: `if nextCursor is None:` -> `if nextCursor:`

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+                                                  email=member["email"])
+                                    for member in group["members"]["nodes"]
+                                },
+                                projects={
+                                    UserGroupProject(id=project["id"],
+                                                     name=project["name"])
+                                    for project in group["projects"]["nodes"]
+                                })
+            nextCursor = userGroups.get("nextCursor", None)
+            # this doesn't seem to be used right now
```
sfendell-labelbox: What does this comment mean?

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,497 @@
...
+        super().__init__()
+        self.color = kwargs.get('color', UserGroupColor.BLUE)
+        self.users = kwargs.get('users', set())
+        self.projects = kwargs.get('projects', set())
+        self.client = client
+
+        # runs against _gql
+        if not client.enable_experimental:
+            raise RuntimeError(
+                "Experimental features are not enabled. Please enable them in the client to use this feature."
```
sfendell-labelbox: Nit: I'd be more specific in the error since we're throwing it from UserGroup specifically: `Please enable experimental in client to use UserGroups`

libs/labelbox/tests/integration/schema/test_user_group.py
```diff
@@ -0,0 +1,114 @@
...
+from labelbox import Client
+from labelbox.schema.user_group import UserGroup, UserGroupColor, UserGroupUser, UserGroupProject
+
+data = faker.Faker()
+
+
+@pytest.fixture
+def client(client):
+    client.enable_experimental = True
+    client.endpoint = "https://app.lb-stage.xyz/api/_gql/"
```
vbrodsky: why hardcoding to stage?
Also, similar to Sam's comments. we could have an experimental_client fixture next to the client fixture in libs/labelbox/tests/conftest.py

adrian-chang: This is temporary, just a way to deal with  internal gql, I do plan to remove

adrian-chang: Resolve and leave sam's comment comment

libs/labelbox/src/labelbox/schema/user_group.py
adrian-chang: The original idea is that people have a list of group ids, you want to pull the rest of the representation (eg I want to get the color, name, etc. even if I don't have it) automatically on UserGroup creation

User groups are updated via GQL via PUT, eg you need to pass the entire group data to the api and not via a PATCH

So imagine you pass an ID (partial representation of the group) only then call update ... you potentially would drop all projects / users etc. as the representation of the group is seen to be empty

At the same rate, I could warn people of this fact via a log message but it's too easy to miss.

libs/labelbox/src/labelbox/schema/user_group.py
adrian-chang: It returns none at the moment, it's not in the GQL schema yet the api returns it. i have yet to see it return a value other than null at the moment.

libs/labelbox/tests/integration/schema/test_user_group.py
adrian-chang: I think this is fine. It's there if you run the file directly (`python test_user_group.py`) instead of `pytest test_user_group.py`.

libs/labelbox/src/labelbox/schema/user_group.py
adrian-chang: Marking resolved

libs/labelbox/src/labelbox/client.py
```diff
@@ -543,6 +543,19 @@ def get_projects(self, where=None) -> PaginatedCollection:
...
+
+        >>> users = client.get_users(where=User.email == "<user_email>")
+
+        Args:
+            where (Comparison, LogicalOperation or None): The `where` clause
+                for filtering.
+        Returns:
+            An iterable of Users (typically a PaginatedCollection).
+        """
+        return self._get_all(Entity.User, where, filter_deleted=False)
```
vbrodsky, line 557: perhaps we should show how to create a `where` clause here?

libs/labelbox/tests/integration/schema/test_user_group.py
```diff
@@ -0,0 +1,107 @@
+import pytest
+import faker
+from labelbox import Client
+from labelbox.schema.user_group import UserGroup, UserGroupColor, UserGroupUser, UserGroupProject
+
+data = faker.Faker()
+
+
+class TestUserGroup:
+
```
vbrodsky: FYI I prefer functions for testing, just easier to deal is

libs/labelbox/tests/integration/schema/test_user_group.py
```diff
@@ -0,0 +1,107 @@
...
+        user_group.color = UserGroupColor.BLUE
+        user_group.create()
+
+        # Verify that the user group was created successfully
+        user_group_equal = UserGroup(client, id=user_group.id)
+        assert user_group.id == user_group_equal.id
+        assert user_group.name == user_group_equal.name
+        assert user_group.color == user_group_equal.color
+
+        user_group.delete()
```
vbrodsky: [required] I would convert user group to a fixture and just test the fixture. Otherwise we would have resource leaks if something goes wrong during the test

libs/labelbox/tests/integration/test_offline_chat_evaluation_project.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 
-
+@pytest.mark.skip(reason="Behavior changed, need to update test")
```
vbrodsky: [required] This should not be marked as skipped

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+from labelbox.schema.user import User
+from labelbox.schema.project import Project
+from labelbox.exceptions import UnprocessableEntityError, InvalidQueryError
+
+
+class UserGroupColor(Enum):
+    """
+    Enum representing the colors available for a group.
+
+    Attributes:
```
vbrodsky, line 16: nice docstring

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+    ORANGE = "FFB35F"
+    CYAN = "4ED2F9"
+    PINK = "FFAEA9"
+    LIGHT_PINK = "FFA9D5"
+    GREEN = "3FDC9A"
+    YELLOW = "E7BF00"
+    GRAY = "B8C4D3"
+
+
+class UserGroupUser(BaseModel):
```
vbrodsky, line 38: [required] should use pydantic_compat

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+    Represents a user in a group.
+
+    Attributes:
+        id (str): The ID of the user.
+        email (str): The email of the user.
+    """
+    id: str
+    email: str
+
+    def __hash__(self):
```
vbrodsky, line 49: good stuff, but why did you need it?

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+        return hash((self.id))
+
+    def __eq__(self, other):
+        if not isinstance(other, UserGroupUser):
+            return False
+        return self.id == other.id
+
+
+class UserGroupProject(BaseModel):
```
vbrodsky, line 58: [required] add pydantic_compat

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+
+    def _reload(self):
+        """
+        Reloads the user group information from the server.
+
+        This method sends a GraphQL query to the server to fetch the latest information
+        about the user group, including its name, color, projects, and members. The fetched
+        information is then used to update the corresponding attributes of the `Group` object.
+
+        Raises:
```
vbrodsky, line 162: nice to list `raises`

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+
+        Returns:
+            bool: True if the two GroupProject objects are equal, False otherwise.
+        """
+        if not isinstance(other, UserGroupProject):
+            return False
+        return self.id == other.id
+
+
+class UserGroup:
```
vbrodsky: why not use pydantic, we could make code simpler by not using getters / setters etc

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+
+        Returns:
+            Group: The newly created group. (self)
+
+        Raises:
+            ResourceCreationError: If the group already exists or if the creation fails.
+            ValueError: If the group name is not provided.
+        """
+        if self.id:
+            raise ResourceCreationError("Group already exists")
```
vbrodsky, line 275: seems like this error should come from the server?

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+        mutation DeleteUserGroupPyApi($id: ID!) {
+            deleteUserGroup(where: {id: $id}) {
+                success
+            }
+        }
+        """
+        params = {"id": self.id}
+        result = self.client.execute(query, params)
+        if not result:
+            raise UnprocessableEntityError("Failed to delete user group")
```
vbrodsky, line 351: should we add a more detailed error message from the server here too?

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+        return result["deleteUserGroup"]["success"]
+
+    @staticmethod
+    def get_user_groups(client: Client) -> Iterator["UserGroup"]:
```
vbrodsky, line 355: nice to use an iterator

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,523 @@
...
+        _name (str): The name of the user group.
+        _color (UserGroupColor): The color of the user group.
+        _users (Set[Union[UserGroupUser, User]]): The set of user IDs in the user group.
+        _projects (Set[Union[UserGroupProject, Project]]): The set of project IDs in the user group.
+        _client (Client): The Labelbox client.
+    """
+    _id: str
+    _name: str
+    _color: UserGroupColor
+    _users: Set[Union[UserGroupUser, User]]
```
vbrodsky: do not see why we include actual `User` and `Project` here and below, since when we get the data from the server we always make them `UserGroupUser`, etc.

vbrodsky (changes requested):

libs/labelbox/src/labelbox/__init__.py
```diff
@@ -40,4 +40,4 @@
 from labelbox.schema.identifiables import UniqueIds, GlobalKeys, DataRowIds
 from labelbox.schema.identifiable import UniqueId, GlobalKey
 from labelbox.schema.ontology_kind import OntologyKind
-from labelbox.schema.project_overview import ProjectOverview, ProjectOverviewDetailed
+from labelbox.schema.project_overview import ProjectOverview, ProjectOverviewDetailed
```
sfendell-labelbox: no changes besides deleting terminating whitespace, revert this file

libs/labelbox/tests/integration/schema/test_user_group.py
sfendell-labelbox: Agreed, but we do use classes in like...6 labelbox-python test files? We should clean that up for consistency (and also delete this one before merge)

libs/labelbox/tests/integration/test_offline_chat_evaluation_project.py
adrian-chang: @vbrodsky this is broken right now (will be blocking), how do you propose to fix?

libs/labelbox/src/labelbox/client.py
adrian-chang:
```
users = client.get_users(where=User.email == "<user_email>")
```
in the docstring

libs/labelbox/src/labelbox/schema/user_group.py
adrian-chang: I disagree

Example 
UserGroup(id="1", name="mygroup").create()  = create { name="mygroup", id="3"} 
UserGroup(name="mygroup")  = create {name ="mygroup", id="2"}

The only thing unique to a group it's an id, two groups can have the same name.

The createApiEndpoint would need to have an id exposed to validate this or you'd have to fetch twice (once to check and then once to create if the fetch results in no group found). Upsert behavior, to avoid this two call tax, which doesn't exist and I'd like to avoid creating an API for this.

Simple is 0 calls. 

If you provide an id, on usergroup initialization, the state is fetched. Creating an existing usergroup again  (with an ID) makes no sense.

libs/labelbox/src/labelbox/schema/user_group.py
adrian-chang: The server will raise the error specifically (e.g. client.execute itself will throw the error), this is just a general guard aginst execute returning none in any case

libs/labelbox/src/labelbox/schema/user_group.py
adrian-chang: No reason, just for convenience if people are iteracting with full user / project objects, I'll remove

libs/labelbox/src/labelbox/schema/user_group.py
vbrodsky: @adrian-chang ^

libs/labelbox/src/labelbox/schema/user_group.py
vbrodsky: @adrian-chang

libs/labelbox/src/labelbox/schema/user_group.py
vbrodsky: Oh so the id is not generated by the server?

libs/labelbox/src/labelbox/schema/user_group.py
```diff
@@ -0,0 +1,418 @@
...
+                }
+            }
+        """
+        nextCursor = None
+        while True:
+            userGroups = client.execute(
+                query, {"nextCursor": nextCursor})["userGroups"]
+            if not userGroups:
+                return
+                yield
```
sfendell-labelbox, line 397: Unreachable?

adrian-chang: See https://stackoverflow.com/questions/6266561/how-to-write-python-generator-function-that-never-yields-anything second answer

vbrodsky (approved):

PR #1605, opened 2024-05-13
Fix project creation error reporting

vbrodsky:
This update resolves an issue where the SDK did not correctly report errors during project or dataset creation. Specifically, it addresses the scenario where an invalid dataset ID is used in the creation of an MOE project.


adrian-chang (approved):
Do we need a test here or is it possible?

PR #1608, opened 2024-05-14
Converted TypedDict to BaseModel for client.send_to_annotate_from_catelog method

Gabefire:
* Previous implementation used a typed dictionary which could not check if either a project ID or model run ID is present
* If either one of these is not present the UI looks strange, and there are no errors also doc string specifies one of these two ids are required
* Making this params field a pydantic base model increases validation support with the added benefit of typo checking
* Also managed to clean up code and remove a lot of the .get methods for the dictionaries
* Integration tests already exist on this method and they pass if needed I can write a unit test to test if the source project id or model run id being present passes the validation

### DIFF
--- a/libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
+++ b/libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
@@ -3,8 +3,7 @@ import sys
 from typing import Optional, Dict
 
 from labelbox.schema.conflict_resolution_strategy import ConflictResolutionStrategy
-from pydantic import model_validator, ConfigDict
-from pydantic.main import BaseModel
+from labelbox import pydantic_compat
 
 if sys.version_info >= (3, 8):
     from typing import TypedDict
@@ -12,7 +11,7 @@ else:
     from typing_extensions import TypedDict
 
 
-class SendToAnnotateFromCatalogParams(BaseModel):
+class SendToAnnotateFromCatalogParams(pydantic_compat.BaseModel):
     """
     Extra parameters for sending data rows to a project through catalog. At least one of source_model_run_id or
     source_project_id must be provided.
@@ -31,7 +30,6 @@ class SendToAnnotateFromCatalogParams(BaseModel):
         ConflictResolutionStrategy.KEEP_EXISTING.
     :param batch_priority: Optional[int] - The priority of the batch. Defaults to 5.
     """
-    model_config = ConfigDict(extra='forbid')
 
     source_model_run_id: Optional[str] = None
     source_project_id: Optional[str] = None
@@ -42,12 +40,16 @@ class SendToAnnotateFromCatalogParams(BaseModel):
         ConflictResolutionStrategy] = ConflictResolutionStrategy.KeepExisting
     batch_priority: Optional[int] = 5
 
-    @model_validator(mode='after')
+    @pydantic_compat.root_validator
     def check_project_id_or_model_run_id(self):
         if not self.source_model_run_id and not self.source_project_id:
             raise ValueError(
-                'source_project_id or source_model_id are required inside params dictionary'
+                'Either source_project_id or source_model_id are required'
             )
+        if self.source_model_run_id and self.source_project_id:
+            raise ValueError(
+                'Provide either only a source_project_id or source_model_id'
+            ) 
 
 
 class SendToAnnotateFromModelParams(TypedDict):
--- a/libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
+++ b/libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
@@ -3,7 +3,7 @@ import sys
 from typing import Optional, Dict
 
 from labelbox.schema.conflict_resolution_strategy import ConflictResolutionStrategy
-from pydantic import model_validator
+from pydantic import model_validator, ConfigDict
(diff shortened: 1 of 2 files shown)


libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
```diff
@@ -3,14 +3,16 @@
...
+from pydantic.main import BaseModel
 
 if sys.version_info >= (3, 8):
     from typing import TypedDict
 else:
     from typing_extensions import TypedDict
 
 
-class SendToAnnotateFromCatalogParams(TypedDict):
+class SendToAnnotateFromCatalogParams(BaseModel):
```
vbrodsky: need to inherit from `pydantic_compat.BaseModel` to support compatibility with with pydantic v1 and v2

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
```diff
@@ -29,14 +31,23 @@ class SendToAnnotateFromCatalogParams(TypedDict):
...
     """
-
-    source_model_run_id: Optional[str]
-    predictions_ontology_mapping: Optional[Dict[str, str]]
-    source_project_id: Optional[str]
-    annotations_ontology_mapping: Optional[Dict[str, str]]
-    exclude_data_rows_in_project: Optional[bool]
-    override_existing_annotations_rule: Optional[ConflictResolutionStrategy]
-    batch_priority: Optional[int]
+    model_config = ConfigDict(extra='forbid')
```
vbrodsky: we generally do not enforce this, please remove

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
```diff
@@ -29,14 +31,23 @@ class SendToAnnotateFromCatalogParams(TypedDict):
...
+    source_model_run_id: Optional[str] = None
+    source_project_id: Optional[str] = None
+    predictions_ontology_mapping: Optional[Dict[str, str]] = {}
+    annotations_ontology_mapping: Optional[Dict[str, str]] = {}
+    exclude_data_rows_in_project: Optional[bool] = False
+    override_existing_annotations_rule: Optional[
+        ConflictResolutionStrategy] = ConflictResolutionStrategy.KeepExisting
+    batch_priority: Optional[int] = 5
+
+    @model_validator(mode='after')
```
vbrodsky: need to use pydantic_compat... are you locally using pydantic v2? I think v1 does not even have mode='after'

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
```diff
@@ -29,14 +31,23 @@ class SendToAnnotateFromCatalogParams(TypedDict):
...
+    def check_project_id_or_model_run_id(self):
+        if not self.source_model_run_id and not self.source_project_id:
+            raise ValueError(
+                'source_project_id or source_model_id are required inside params dictionary'
```
vbrodsky: I would just say `Either source_project_id or source_model_id are required`

vbrodsky (changes requested):

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
Gabefire: I can remove it but the only reason I added was because this was a unique parameter in our SDK. We pass in a dictionary of parameters instead of just the arguments which I think we do everywhere else only these methods are like this. From getting on a call with a customer one thing I saw was accidental typos with the hard strings like if I put source_project_ids as a key instead of source_project_id and they get confused why nothing happened since it wont error out and everything has a default value.

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
vbrodsky: Is there a check that both should not be present?

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
Gabefire: I removed it!

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
Gabefire: Added

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
Gabefire: Changed!

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
Gabefire: Changed!

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
Gabefire: Done!

libs/labelbox/src/labelbox/schema/send_to_annotate_params.py
vbrodsky: wait your argument made sense to me... I am good either way

vbrodsky (approved):

PR #1609, opened 2024-05-14
[SN-131] update_huggingface.ipynb

paultancre:
This is an overall rework of our HG notebook, added simplified methods removed outdated lib.


lb-pno (approved):
LGTM. Thanks

PR #1612, opened 2024-05-14
Added notebook workflow to remove metadata through ci

Gabefire:


.github/workflows/notebooks.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+          fetch-depth: 0
+      - name: Install databooks
+        run: sudo pip3 install databooks
```
adrian-chang: I think you can do "pip install databooks"

I doubt we have to worry about python 2

See 'https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md'

adrian-chang (approved):

PR #1615, opened 2024-05-16
Add get project_overview to retrieve the value of the Overview tab

lb-pno:
`project.get_overview()` provides a programmatic way to retrieve the content of the Overview tab of a project and in particular:

|attribute|description|
|-|-|
|to_label| Number of data rows to label|
|in_review| Dictionary with different review queues and associated number of data rows|
|in_rework|Number of data rows to rework|
|skipped|Number of skipped data rows|
|done|Number of done data rows|
|issues|Number of issues|
|labeled|Number of labeled data rows|
|all_in_data_rows|Total number of data rows|

This change introduces a new model: `Project_Overview`

Examples:

```project.get_overview()```

Result:

```
Project_Overview(to_label=99989, in_review={'Initial review task': 4, 'all': 4}, in_rework=0, skipped=4, done=0, issues=None, labeled=4, all_in_data_rows=99993)
```

Note: In order to retrieve the number of issues, `client.execute()` must run with `experimental=True`.

### CODE REVIEW

File: libs/labelbox/src/labelbox/schema/project_overview.py [hunk 1]
```diff
@@ -0,0 +1,15 @@
+from typing import Dict, Optional
+from labelbox import pydantic_compat
+
+class Project_Overview(pydantic_compat.BaseModel):
```
adrian-chang: This should be CamelCase `ProjectOverview`

libs/labelbox/src/labelbox/schema/project_overview.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    Class that represents a project overview.
```
adrian-chang: Do you think copilot can generate the detailed params comments here?

libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -1746,6 +1747,63 @@ def __check_data_rows_have_been_processed(
         return response["queryAllDataRowsHaveBeenProcessed"][
             "allDataRowsHaveBeenProcessed"]
 
+    def get_overview(self) -> Project_Overview:
```
adrian-chang: Should there be a unit test for this?

adrian-chang (changes requested):

libs/labelbox/src/labelbox/schema/project_overview.py
lb-pno: Thanks. The name of the class was updated.

libs/labelbox/src/labelbox/schema/project_overview.py
lb-pno: Thanks. I have added a few comments to provide additional details.

libs/labelbox/src/labelbox/schema/project.py
lb-pno: The current test (`test_get_overview()`) is part of `libs/labelbox/tests/integration/test_task_queue.py` because it is related since the new method provides the number of data rows per task queue.

libs/labelbox/src/labelbox/schema/project_overview.py
```diff
@@ -0,0 +1,18 @@
+from typing import Dict, Optional
+from labelbox import pydantic_compat
+
+class ProjectOverview(pydantic_compat.BaseModel):
```
adrian-chang: Do you just want to import the BaseModel, small style knit, also use code pilot to generate the param definitions?

On that note, it's probably worth updating the docs/.rst to include this new project so it shows up on readthedocs

adrian-chang (changes requested):

libs/labelbox/src/labelbox/schema/project_overview.py
lb-pno: Thanks for the comment.

All the attributes are now documented.
`project-overview.rst` was added.

I am not sure of what you mean by "_Do you just want to import the BaseModel, small style knit,_". Are you suggesting that a dataclass would be more appropriate in this case?

adrian-chang: `from labelbox.pydantic_compat import BaseModel`

lb-pno: OK but `from labelbox import pydantic_compat` is used everywhere else. I can change it but it's going to look off.

image

lb-pno: Import changed.

adrian-chang (approved):

PR #1617, opened 2024-05-16
[SN-121] Cleaned up readme created python file to auto generate

Gabefire:


Gabefire:
https://github.com/Labelbox/labelbox-python/tree/gu/clean_up_examples_readme/examples

Gabefire:
* HTML version for inline (like readme): https://github.com/Labelbox/labelbox-python/tree/gu/clean_up_examples_readme/examples

* MD Version like old readme: https://github.com/Labelbox/labelbox-python/blob/b2ec21ad7a4535f1227bdd59e8cd774628a96ee4/examples

@paulnoirel For you to check tomorrow but I made two versions of my generator script above the HTML one follows what readme has so we can use that instead to maybe automatically generate it on the readme custom page. I think the tables look almost identical TBH

adrian-chang (approved):

Gabefire:
Made a readme hidden page just to see what it looks like if you want to check https://dash.readme.com/project/labelbox-docs/v5/page/test

Changed it to full HTML for the table and header it seems to work a lot better in readme

PR #1619, opened 2024-05-20
[SN-0] Switched notebooks to RYE for better management cleaned up everything

Gabefire:
Modified notebook workflow to help with styling and automating some of our docs


adrian-chang:
Can you explain what happened here? I get the title, but I don't see how Rye, for example is used at all in the `examples/` folder

examples/pyproject.toml
Gabefire, line 1: @adrian-chang The pyproject file is right here. A couple of reasons for Rye:
1. manage dev dependencies. I'm using a few things to format
2. match the version of the main SDK to align
3. The Rye custom scripts are extremely powerful. I don't have to change workflow files anymore or add installs. Rye will literally do it for me. I just have to update the `rye run clean` script, and it runs the correct formatting files and commands. We can add or remove to it if needed and update dev dependencies without touching workflows or leaving the examples directory. 
4. Quick workflow installs
5. It also allows syncing dev dependencies, so you just run Rye sync, and you have all the dependencies to work on the example notebooks and main SDK. This results in a cleaner and consistent experience across the repo.

The [contributing](https://github.com/Labelbox/labelbox-python/blob/f8f9dba5acd193e8c8fb29ebf7dd8de13ce60d22/examples/CONTRIBUTING.md) guide that was added to this PR explains the scripts
[Rye](https://rye-up.com/guide/virtual/) actually mentions doing a virtual project for a doc-like project.

Gabefire: But the reason why all were updated was because the workflow script linted all the code and formatted. I’m having someone else review the notebooks with me to ensure they work but the more engineering focus stuff like workflows I’d definitely want your approval

.github/actions/notebooks/action.yml
```diff
@@ -0,0 +1,18 @@
+name: Labelbox Example Notebook Setup
```
adrian-chang: Are you sure you don't want to use the same one here https://github.com/Labelbox/labelbox-python/blob/develop/.github/actions/python-package-shared-setup/action.yml as it looks very similar?

adrian-chang (dismissed):
One small nit, otherwise looks fine

.github/actions/notebooks/action.yml
Gabefire: Done

adrian-chang (dismissed):

adrian-chang:
Please try to ensure commits get squashed into one (as there are many)

Gabefire:
Merged so tests will rerun

sfendell-labelbox (approved):

PR #1622, opened 2024-05-20
[PLT-1018] Add dataset.set_iam_integration() to select/deselect integrations

lb-pno:
Currently, users can create datasets with the SDK but can't modify the integration outside of the UI. This makes migration projects tedious since no automation is possible.

The current method brings a quick solution to this problem.

With this new method it is possible to:

- Set IAM integration for integration id
```python
# Get all IAM integrations
iam_integrations = client.get_organization().get_iam_integrations()

 # Get IAM integration id
iam_integration_id = [integration.uid for integration
   in iam_integrations
   if integration.name == "My S3 integration"][0]

dataset.add_iam_integration(iam_integration_id)
```

- Set IAM integration for IAM integration object
```python
# Get all IAM integrations
iam_integrations = client.get_organization().get_iam_integrations()

 # Get IAM integration object
iam_integration = [integration.uid for integration
   in iam_integrations
   if integration.name == "My S3 integration"][0]
 
dataset.add_iam_integration(iam_integration)
```
- Remove IAM integration
```python
dataset.remove_iam_integration()
```


libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,80 @@ def _convert_items_to_upsert_format(self, _items):
...
+          >>>
+          >>>   # Set IAM integration for IAMIntegrtion object
+          >>>   dataset.set_iam_integration(iam_integration)
+          >>>
+          >>>   # Unset IAM integration
+          >>>   dataset.set_iam_integration()
+          """
+
+          # Unset IAM integration if iam_integration is None
+          if iam_integration is None:
```
sfendell-labelbox: I'd rather this method was split into two, `set_iam_integration` and `remove_iam_integration`, and the iam_integration argument was made required.

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,80 @@ def _convert_items_to_upsert_format(self, _items):
...
+              query = """mutation DetachSignerPyApi($id: ID!) {
+                  clearSignerForDataset(where: {id: $id}) {
+                      id
+                      signer {
+                      id
+                      }
+                  }
+                  }"""
```
sfendell-labelbox: Nit: Can you fix the indenting on these two gql queries?

libs/labelbox/src/labelbox/schema/dataset.py
sfendell-labelbox: Actually, could you do `add_iam_integration` and `remove_iam_integration`? We don't have a standard for this yet, but I'd like to start having 'create/delete' semantics for creating/deleting objects, and `add/remove` for creating/deleting associations between objects.

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,80 @@ def _convert_items_to_upsert_format(self, _items):
...
+              response = self.client.execute(query, {"id": self.uid})
+              
+              if response:
+                 return response["clearSignerForDataset"]["signer"]
+              else:
+                raise LabelboxError("Can't unset IAM integration")
+              
+          else:
+
+              if isinstance(iam_integration, IAMIntegration):
```
sfendell-labelbox: Nit: I'd simplify this a bit into `iam_integration_id = iam_integration.uid if isInstance(..) else iam_integration`

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,80 @@ def _convert_items_to_upsert_format(self, _items):
...
+                  iam_integration_id = iam_integration.uid
+              else:
+                  iam_integration_id = iam_integration
+
+              query = """mutation AttachSignerPyApi($signerId: ID!, $datasetId: ID!) {
```
sfendell-labelbox: `AttachSignerPyApi` -> `SetSignerForDatasetPyApi`

We are almost totally consistent with this standard across GQL queries so I'd prefer to keep it that way. Same for removing signers.

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,80 @@ def _convert_items_to_upsert_format(self, _items):
...
+                  setSignerForDataset(data: {signerId: $signerId}, where: {id: $datasetId}) {
+                      id
+                      signer {
+                      id
+                      }
+                  }
+                  }"""
+              response = self.client.execute(query, {"signerId": iam_integration_id, "datasetId": self.uid})
+              
+              # Return IAM Integration object if
```
sfendell-labelbox: Delete this unfinished comment

sfendell-labelbox: Or finish it; I just don't think it's super useful (presumably the complete comment is "Return object if it is created" or something equivalent)

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,80 @@ def _convert_items_to_upsert_format(self, _items):
...
+                      id
+                      signer {
+                      id
+                      }
+                  }
+                  }"""
+              response = self.client.execute(query, {"signerId": iam_integration_id, "datasetId": self.uid})
+              
+              # Return IAM Integration object if  
+              if response:
```
sfendell-labelbox: Response can only be `None` if the resource is not found. So I'd prefer something like:

```
if not response:
    <raise some appropriate ResourceNotFoundError>
    
<rest of code here>
```

fwiw I hate this pattern, `execute` should just throw an error instead of returning None (since that's what its docstring says it does) and I'd like to to refactor this as of seeing it in this PR. But for now, this pattern is present in the rest of client, so stick to this.

libs/labelbox/tests/integration/test_delegated_access.py
```diff
@@ -77,3 +78,52 @@ def test_no_default_integration(client):
...
+    # Prepare dataset with an existing integration
+    integration = [
+        integration for integration 
+        in integrations 
+        if 'aws-da-test-bucket' in integration.name][0]
+
+    ds = client.create_dataset(iam_integration=integration, name=f"integration_change-{uuid.uuid4()}")
+
+    # Test unset integration
+    ds.set_iam_integration()
```
sfendell-labelbox: Split this test. it's testing 3 different codepaths (set with id, set with object, unset) in one test that could easily be separated.

sfendell-labelbox (changes requested):
Comments left. LGTM in concept, but some styling/consistency stuff.

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Done

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Done

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Done

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Done

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Done

libs/labelbox/tests/integration/test_delegated_access.py
lb-pno: Done

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Comment removed

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,104 @@ def _convert_items_to_upsert_format(self, _items):
...
+                        id
+                    }
+                }
+            }
+         """
+
+          response = self.client.execute(query, {"signerId": iam_integration_id, "datasetId": self.uid})
+
+          if not response:
+            raise LabelboxError(f"Can't set IAM integration {iam_integration_id}")
```
sfendell-labelbox: Replace with:

```raise ResourceNotFoundError(IAMIntegration, {"signerId": iam_integration_id, "datasetId": self.uid})```

Look at other uses of ResourceNotFound error for this pattern.

File: libs/labelbox/src/labelbox/schema/dataset.py [hunk 9]
```diff
@@ -912,3 +913,104 @@ def _convert_items_to_upsert_format(self, _items):
...
+            }
+        }
+        """
+
+        response = self.client.execute(query, {"id": self.uid})
+
+        if response:
+            return response.get("clearSignerForDataset", {}).get("signer")
+        else:
+            raise LabelboxError("Can't unset IAM integration")
```
sfendell-labelbox: ResourceNotFound error here too

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,104 @@ def _convert_items_to_upsert_format(self, _items):
...
+                    id
+                }
+            }
+        }
+        """
+
+        response = self.client.execute(query, {"id": self.uid})
+
+        if response:
+            return response.get("clearSignerForDataset", {}).get("signer")
```
sfendell-labelbox: Return statement conflicts with docstring and typing.

lb-pno: Actually, there is no conflict but I admit that it's not obvious.
This is an example of value for `response`:

```python
{"clearSignerForDataset": {"id": "clwepodbo002i0716bwufsxfr", "signer": None}}
```

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Done

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Done

libs/labelbox/src/labelbox/schema/dataset.py
sfendell-labelbox: Is the value signer always None? If so, I'd simplify it as:
if not response:
   raise LabelboxError
 <end fo function>
 
Also, if signer is always none, why specify its return value in the gql query?

lb-pno: Done.

I've updated the error to 
```raise ResourceNotFoundError(Dataset, {"id": self.uid})```
since response is typically null if the dataset has been deleted.

File: libs/labelbox/src/labelbox/schema/dataset.py [hunk 11]
```diff
@@ -912,3 +913,101 @@ def _convert_items_to_upsert_format(self, _items):
...
+                id
+            }
+        }
+        """
+
+        response = self.client.execute(query, {"id": self.uid})
+
+        if not response:
+            raise ResourceNotFoundError(Dataset, {"id": self.uid})
+
```
sfendell-labelbox: Final nit: Get rid of this extra whitespace at the end of the file.

sfendell-labelbox (dismissed):

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Done

sfendell-labelbox (dismissed):

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -912,3 +913,98 @@ def _convert_items_to_upsert_format(self, _items):
...
+            }
+         """
+
+          response = self.client.execute(query, {"signerId": iam_integration_id, "datasetId": self.uid})
+
+          if not response:
+            raise ResourceNotFoundError(IAMIntegration, {"signerId": iam_integration_id, "datasetId": self.uid})
+          
+          try:
+            iam_integration_id = response.get("setSignerForDataset", {}).get("signer", {})["id"]
```
adrian-chang, line 974: This seems oddly formatted here, can you format to 4 spaces?

adrian-chang (changes requested):

libs/labelbox/src/labelbox/schema/dataset.py
lb-pno: Method reformatted. The docstring was off compared to the other methods, and it messed up the all indentation.

adrian-chang:
@paulnoirel please fill out https://labelbox.atlassian.net/browse/PLT-1018 with some details for QA

adrian-chang (approved):

PR #1623, opened 2024-05-21
Throw resource not found error for model config methods.

sfendell-labelbox:


vbrodsky (dismissed):
should we add an integration test for 1 case?

sfendell-labelbox:
> should we add an integration test for 1 case?

Done.

vbrodsky (approved):

PR #1627, opened 2024-05-23
[PLT-43] Vb/create datarows chunking plt 43

vbrodsky:
Stories:
- https://labelbox.atlassian.net/browse/PLT-43
- https://labelbox.atlassian.net/browse/PLT-988

This pull request introduces two key improvements for the create_data_rows function:

**Increased Upload Limit:** We're removing the previous 150,000 data row limit, allowing for significantly larger datasets to be uploaded.
**Enhanced Efficiency**: The create_data_rows function now leverages our existing upsert code, which utilizes data chunking for improved performance. Additionally, it incorporates paginated results and error handling through lazy evaluation, already available in the SDK.

### 
**_Challenges Addressed_**:

**Legacy Upsert Code**: The upsert functionality relied heavily on outdated code for data generation.
Solution: The code has been refactored to separate data similarities and differences. A dedicated subclass was created specifically for the create_data_rows use case.
**Edge Cases in Data Row Creation:** The original create_data_rows function contained various edge cases that could lead to errors.
Solution: These edge cases were identified through our comprehensive integration test suite. To address them, we've added logic to support uploading data from local files within create_data_rows. Additionally, we've verified that the remaining edge cases are now handled correctly.

**_Technical Considerations_**:

To implement paginated result and error streaming while maintaining backward compatibility for create_data_rows, we created a subclass named DataUpsertTask that inherits from the Task class. Due to limitations in extensibility of the Task class, some workarounds were necessary. The functionality has been thoroughly tested through extensive SDK integration tests.: The create_data_rows function now leverages our existing upsert code, which utilizes data chunking for improved performance. Additionally, it incorporates paginated results and error handling through lazy evaluation, already available in the SDK.


libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -29,26 +29,17 @@
...
+from labelbox.schema.internal.data_row_uploader import DataRowUploader
+from labelbox.schema.internal.datarow_upload_constants import (
+    MAX_DATAROW_PER_API_OPERATION, FILE_UPLOAD_THREAD_COUNT, UPSERT_CHUNK_SIZE)
 
 logger = logging.getLogger(__name__)
 
-MAX_DATAROW_PER_API_OPERATION = 150_000
-
-
-class DataRowUpsertItem(BaseModel):
```
vbrodsky, line 41: refactored into a separate package

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -284,271 +280,39 @@ def create_data_rows(self, items) -> "Task":
             InvalidAttributeError: If there are fields in `items` not valid for
                 a DataRow.
             ValueError: When the upload parameters are invalid
-        """
```
vbrodsky, line 288: refactored into a separate package

libs/labelbox/src/labelbox/schema/task.py
```diff
@@ -49,6 +55,13 @@ class Task(DbObject):
     created_by = Relationship.ToOne("User", False, "created_by")
     organization = Relationship.ToOne("Organization")
 
+    def __eq__(self, task):
```
vbrodsky, line 58: To support the `DataUpsertTask` in the list `user.created_task()`
We really should refactor our tasks (sigh)

libs/labelbox/src/labelbox/schema/task.py
```diff
@@ -49,6 +55,13 @@ class Task(DbObject):
...
+        return isinstance(
+            task, Task) and task.uid == self.uid and task.type == self.type
+
+    def __hash__(self):
```
vbrodsky, line 62: to support lru caching

libs/labelbox/src/labelbox/schema/task.py
```diff
@@ -214,3 +227,170 @@ def get_task(client, task_id):
...
+    def errors(self) -> Optional[List[Dict[str, Any]]]:  # type: ignore
+        return self._errors_as_list()
+
+    @property
+    def created_data_rows(  # type: ignore
+            self) -> Optional[List[Dict[str, Any]]]:
+        return self.result
+
+    @property
+    def failed_data_rows(  # type: ignore
```
vbrodsky, line 266: had to do `# type: ingnore` because linter complained due to a change in return type (see the above comment re: need to refactor Task)

libs/labelbox/src/labelbox/schema/internal/data_row_create_upsert.py
```diff
@@ -0,0 +1,76 @@
+from abc import ABC, abstractmethod
+from typing import List, Tuple, Optional
+
+from labelbox.schema.identifiable import UniqueId, GlobalKey
+from labelbox.pydantic_compat import BaseModel
+
+
+class DataRowItemBase(BaseModel, ABC):
```
mrobers1982: This class does not look to be abstract - all methods are implemented.

I'm not sure you ultimately need/want inheritance, perhaps we can take advantage of another OO design pattern.

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
```diff
@@ -0,0 +1,294 @@
...
+from labelbox.exceptions import InvalidAttributeError
+from labelbox.exceptions import MalformedQueryException
+from labelbox.orm.model import Entity
+from labelbox.orm.model import Field
+from labelbox.schema.embedding import EmbeddingVector
+from labelbox.schema.internal.data_row_create_upsert import DataRowItemBase
+from labelbox.schema.internal.datarow_upload_constants import MAX_DATAROW_PER_API_OPERATION
+
+
+class UploadManifest:
```
mrobers1982: nit: Why not use a Pydantic model?

libs/labelbox/src/labelbox/schema/task.py
```diff
@@ -214,3 +227,188 @@ def get_task(client, task_id):
         task: Task = tasks[0]
         task._user = user
         return task
+
+
+class DataUpsertTask(Task):
+    """
+    Task class for data row upsert operations
+    """
+    __max_donwload_size: Final = MAX_DATAROW_PER_API_OPERATION
```
mrobers1982: nit: Constants are typically in UPPER_CASE (at the very least we should be consistent).

libs/labelbox/src/labelbox/schema/task.py
```diff
@@ -214,3 +227,188 @@ def get_task(client, task_id):
...
+    def result(self) -> Optional[List[Dict[str, Any]]]:  # type: ignore
+        """
+        Fetches maximum 150K results. If you need to fetch more, use `result_all` property
+        """
+        if self.status == "FAILED":
+            raise ValueError(f"Job failed. Errors : {self.errors}")
+        return self._results_as_list()
+
+    @property
+    def errors(self) -> Optional[List[Dict[str, Any]]]:  # type: ignore
```
mrobers1982, line 253: I'm curious why we are not simply supporting a single `results` and `errors` method which returns a generator?

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
```diff
@@ -0,0 +1,294 @@
...
+
+    @staticmethod
+    def create_descriptor_file(client,
+                               items,
+                               max_attachments_per_data_row=None,
+                               is_upsert=False):
+        """
+        This function is shared by `Dataset.create_data_rows`, `Dataset.create_data_rows_sync` and `Dataset.update_data_rows`.
+        It is used to prepare the input file. The user defined input is validated, processed, and json stringified.
+        Finally the json data is uploaded to gcs and a uri is returned. This uri can be passed to
```
sfendell-labelbox: Passed to what?

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
```diff
@@ -0,0 +1,294 @@
...
+            ids = list(map(lambda spec: spec.id.get("value"), empty_specs))
+            raise ValueError(
+                f"The following items have an empty payload: {ids}")
+
+        chunks = [
+            specs[i:i + upsert_chunk_size]
+            for i in range(0, len(specs), upsert_chunk_size)
+        ]
+
+        def _upload_chunk(_chunk):
```
sfendell-labelbox: Nit: I don't think we have this pattern of an underscore prefixing a variable anywhere else, can this just be `chunk`?

libs/labelbox/src/labelbox/schema/task.py
sfendell-labelbox: Also, `donwload` should be `download`

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
```diff
@@ -0,0 +1,294 @@
...
+            allowed_extra_fields = {
+                'attachments', 'media_type', 'dataset_id', 'embeddings'
+            }
+            invalid_keys = set(item) - {f.name for f in DataRow.fields()
+                                       } - allowed_extra_fields
+            if invalid_keys:
+                raise InvalidAttributeError(DataRow, invalid_keys)
+            return item
+
+        def formatLegacyConversationalData(item):
```
sfendell-labelbox: snake_case instead of camelCase

libs/labelbox/src/labelbox/schema/internal/datarow_upload_constants.py
```diff
@@ -0,0 +1,4 @@
+MAX_DATAROW_PER_API_OPERATION = 150_000
+FILE_UPLOAD_THREAD_COUNT = 20
```
sfendell-labelbox, line 2: Looks like we also define this in data_row_uploader, do we use this constant anywhere?

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
```diff
@@ -0,0 +1,294 @@
...
+
+        Raises:
+            InvalidQueryError: If the `items` parameter does not conform to
+                the specification above or if the server did not accept the
+                DataRow creation request (unknown reason).
+            InvalidAttributeError: If there are fields in `items` not valid for
+                a DataRow.
+            ValueError: When the upload parameters are invalid
+        """
+        file_upload_thread_count = 20
```
sfendell-labelbox: Can we use the existing constant here?

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
vbrodsky: I could ok

libs/labelbox/src/labelbox/schema/task.py
vbrodsky: this is code from Attila, I will update

libs/labelbox/src/labelbox/schema/task.py
vbrodsky: because it will not be backward-compatible

libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -264,13 +257,16 @@ def create_data_rows_sync(self, items) -> None:
             url_param: descriptor_url
         })
 
-    def create_data_rows(self, items) -> "Task":
+    def create_data_rows(self,
+                         items,
+                         file_upload_thread_count=FILE_UPLOAD_THREAD_COUNT
```
sfendell-labelbox, line 260: We should validate that thread_count is a positive integer here and raise an error if not

libs/labelbox/tests/integration/test_dataset.py
sfendell-labelbox, line 1: Maybe it's worthwhile to test with a non-default number of threads for create_data_rows?

libs/labelbox/src/labelbox/schema/task.py
mrobers1982: This is likely going to be confusing for users, i.e. `results` vs. `results_all`.

Returning a generator will likely not break things in practice.

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def to_dict(self):
+        return {
+            "source": self.source,
+            "item_count": self.item_count,
+            "chunk_uris": self.chunk_uris
+        }
+
+
+class DataRowUploader:
```
sfendell-labelbox, line 25: Correct me if i'm wrong, but this class has only static methods, no instance variables, and is never instantiated? If so, I think we don't need the class. Just export its static methods as functions directly

vbrodsky: I am just trying to follow what seems to be an sdk pattern: not to use functions, but to group like functions in a class... lmk if my understanding is not correct

libs/labelbox/src/labelbox/schema/task.py
vbrodsky: It will, I think, if someone is trying to use returns as a List

libs/labelbox/src/labelbox/schema/task.py
```diff
@@ -214,3 +227,169 @@ def get_task(client, task_id):
...
+        return PaginatedCollection(
+            client=self.client,
+            query=query_str,
+            params=params,
+            dereferencing=['failedDataRowImports', 'results'],
+            obj_class=convert_errors_to_legacy_format,
+            cursor_path=['failedDataRowImports', 'after'],
+        )
+
+    def _results_as_list(self) -> Optional[List[Dict[str, Any]]]:
```
mrobers1982, line 369: nit: I believe `PaginatedCollection` is an iterator, in which case you can reduce this function to a single line of code (or remove it entirely).

```python
def _results_as_list(self) -> Optional[List[Dict[str, Any]]]:
    return list(self._download_results_paginated()) or None
```

mrobers1982 (approved):

PR #1628, opened 2024-05-24
[PLT-1008] Export content parses incorrectly 

adrian-chang:
* Introduce Buffered Stream
* Prep deprecation of non-buffered methods
* Fixup errors / result in export.task to use bufferedstream


libs/labelbox/src/labelbox/schema/export_task.py
```diff
@@ -520,6 +476,144 @@ def start(
...
+            f"{{task(where: $where)"
+            f"{{{'exportFileFromOffset'}(streamType: $streamType, offset: $offset)"
+            f"{{offsets {{start end}} lines {{start end}} file}}"
+            f"}}}}")
+        variables = {
+            "where": {
+                "id": self._ctx.task_id
+            },
+            "streamType": self._ctx.stream_type.value,
+            "offset": str(self._current_offset),
```
vbrodsky, line 510: offset is declared as UInt64! So does this conversion do it?

vbrodsky (dismissed):
I see the approach and agree with it. The code looks right, not enough deep understanding on my end to dig into each and every line.

libs/labelbox/src/labelbox/schema/export_task.py
adrian-chang: I'd assume so, this is from Klaus, i Just moved it / reused it

Gabefire:
Are you technically deprecating the [FileConverterType](https://github.com/Labelbox/labelbox-python/blob/705275b141518cfade1720429e4f748e2d0cffb8/libs/labelbox/src/labelbox/schema/export_task.py#L181) with this change since the buffered stream does not look like you can pass custom converters?

Gabefire:
NIT: one last thing: would it make sense to change the default value for the convertor in the [get_stream method](https://github.com/Labelbox/labelbox-python/blob/705275b141518cfade1720429e4f748e2d0cffb8/libs/labelbox/src/labelbox/schema/export_task.py#L899) to the buffered version so the deprecation messages do not fire twice/ clean up the method easily? This is for this [workflow](https://docs.labelbox.com/reference/label-export#advanced-usage).

sfendell-labelbox (approved):

adrian-chang:
> Are you technically deprecating the [FileConverterType](https://github.com/Labelbox/labelbox-python/blob/705275b141518cfade1720429e4f748e2d0cffb8/libs/labelbox/src/labelbox/schema/export_task.py#L181) with this change since the buffered stream does not look like you can pass custom converters?

Sort of. I think the way it is written is a bit incorrect (yielding a file constantly is very odd in my view) but the main idea of yielding an entire file is correct. 

The Export_Task code is very over engineered and buggy and needs to be simplified. 

We should aim to give customers basically 3 outputs: fully outputted in memory, line by line in memory, or an entire file by disk and that's it. 

Allowing for a lot of customization such as choosing what chunk to export at (offset) or line makes etc. just makes our life more difficult maintaining the SDK.

adrian-chang:
> NIT: one last thing: would it make sense to change the default value for the convertor in the [get_stream method](https://github.com/Labelbox/labelbox-python/blob/705275b141518cfade1720429e4f748e2d0cffb8/libs/labelbox/src/labelbox/schema/export_task.py#L899) to the buffered version so the deprecation messages do not fire twice/ clean up the method easily? This is for this [workflow](https://docs.labelbox.com/reference/label-export#advanced-usage).

No. BufferedJsonOutput has a schema of ` { json: any }` versus ` { json_str: str } `. Changing the convertor is a potential breaking change.

This is also why buffered_stream was introduced.

PR #1630, opened 2024-05-24
PLT-1016 - Add methods for External Workforce management

lb-pno:
This PR adds three methods to manage external workforces (aka BPO organizations) from a project, in particular:

```python
project.add_external_workforce(workforce_id)

project.remove_external_workforce(workforce_id)

project.get_external_workforces()
```

All three methods return a list of `ExternalWorkforce` associated with the project at the end of the operation or an empty list.


libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -1824,6 +1825,115 @@ def clone(self) -> "Project":
...
+        Raises:
+            LabelboxError: If the external workforce with the given ID cannot be found.
+
+        Note:
+            This method adds an external workforce (organization) to the current project. 
+            The `workforce_id` parameter can be either a string representing the organization ID or an instance of the `ExternalWorkforce` class.
+        """
+        workforce_id = workforce_id.uid if isinstance(workforce_id, ExternalWorkforce) else workforce_id
+
+        mutation = """
```
adrian-chang, line 1903: Does this require experimental client mode on? Just wondering don't know myself

adrian-chang (changes requested):
See comments

libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -1840,6 +1841,115 @@ def clone(self) -> "Project":
...
+                        name
+                    }
+                }
+            }
+            """
+        
+        result = self.client.execute(mutation, {"projectId": self.uid, "organizationId": workforce_id})
+        
+        if not result:
+            raise LabelboxError(f"Can't find External Workforce {workforce_id}")
```
sfendell-labelbox: Change to `ResourceNotFoundError`

libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -1840,6 +1841,115 @@ def clone(self) -> "Project":
...
+                        name
+                    }
+                    }
+            }
+        """
+
+        result = self.client.execute(mutation, {"projectId": self.uid, "organizationId": workforce_id})
+
+        if not result:
+            raise LabelboxError(f"Can't find External Workforce {workforce_id}")
```
sfendell-labelbox: Change to `ResourceNotFoundError`

libs/labelbox/tests/integration/test_external_workforce.py
```diff
@@ -0,0 +1,41 @@
+from labelbox import Project, ExternalWorkforce
```
sfendell-labelbox, line 1: Remove trailing whitespace from this file (we have some throughout the codebase but prefer to not add more, we really should add this to our lint rules)

libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -42,6 +42,7 @@
 from labelbox.schema.task_queue import TaskQueue
 from labelbox.schema.ontology_kind import (EditorTaskType, OntologyKind)
 from labelbox.schema.project_overview import ProjectOverview, ProjectOverviewDetailed
+from labelbox.schema.external_workforce import ExternalWorkforce
```
sfendell-labelbox: Remove trailing whitespace from this file

sfendell-labelbox: Also, I know this is such a nit, but as long as you're editing it...Could you alphabetize this import to match the rest?

libs/labelbox/tests/integration/test_external_workforce.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+import sys
+
+def _get_workforce_id_for_test() -> str:
+    # Basic function to provide an organization id for the test
+    # org id from https://labelbox.atlassian.net/wiki/spaces/PLT/pages/2110816271/How+to+labelbox-python+SDK+CI+Tests
+    current_version = sys.version_info[:2] 
+
+    if os.environ['LABELBOX_TEST_ENVIRON'] == "staging":
```
vbrodsky, line 25: what is the meaning of this? can you add comments?

lnash4 (changes requested):
so per @manuaero 's request we would not like to use the add and remove endpoints anymore, now that we have built boost requests within the product 

the mutation's name is upsertProjectBoostWorkforce, this will automatically add the alignerr org as an external workforce

then if the project is properly set up, which means they have 

1. ontology configured 
2. data added
3. there are labeling instructions for the ontology

then you can move the project to the 'REQUESTED' status with the 
updateProjectBoostWorkforceStatus mutation

lb-pno:
I am closing the current PR since it is not relevant with Alignerr.

PR #1631, opened 2024-05-24
[SN-144] Removed uneeded notebook ths is the last one that needs gone

Gabefire:
Notebook is not needed anymore its outdated and their is no links to our docs


lb-pno (approved):
LGTM

PR #1639, opened 2024-05-29
Add example for project.clone()

lb-pno:


mina-ebeid:
Running into an error when using `project.clone()` returning a project object rather than the assumed projectID of the cloned project. Which leads the line `destination_project = client.get_project(destination_project_id)` to break

mina-ebeid (approved):
Looks good to go!

PR #1644, opened 2024-05-31
[PLT-821][PLT-822] Throw user readable errors when creating a custom embedding.

sfendell-labelbox:
Fixes # (PLT-981) (PLT-982)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## All Submissions

- [X] Have you followed the guidelines in our Contributing document?
- [X] Have you provided a description?
- [X] Are your changes properly formatted?

## New Feature Submissions

- [X] Does your submission pass tests?
- [X] Have you added thorough tests for your new feature?

## Changes to Core Features

- [X] Have you written new tests for your core changes, as applicable?
- [X] Have you successfully run tests with your changes locally?

### DIFF
--- a/libs/labelbox/src/labelbox/adv_client.py
+++ b/libs/labelbox/src/labelbox/adv_client.py
@@ -3,6 +3,7 @@ import json
 import logging
 from typing import Dict, Any, Optional, List, Callable
 from urllib.parse import urlparse
+from labelbox.exceptions import LabelboxError
 
 import requests
 from requests import Session, Response
@@ -59,7 +60,12 @@ class AdvClient:
                                         url,
                                         data=requests_data,
                                         headers=headers)
-        response.raise_for_status()
+        if response.status_code != requests.codes.ok:
+            message = response.json().get('message')
+            if message:
+                raise LabelboxError(message)
+            else:
+                response.raise_for_status()
         return response.json()
 
     def _send_ndjson(self,


libs/labelbox/src/labelbox/adv_client.py
```diff
@@ -59,7 +60,18 @@ def _request(self,
...
+            if response_json.get('status') == 'CONFLICT':
+                raise ADVError("A custom embedding with this name already exists. Custom embeddings must have unique names.")
+            message = response_json.get('message')
+            if message:
+                if message == "must be greater than or equal to 8":
+                    raise ADVError("Custom embeddings require at least 8 dimensions.")
+                else:
+                    raise ADVError(response.json()['message'])
+            else:
+                response.raise_for_status()
```
mrobers1982, line 68: I find this approach brittle for a couple reasons: 1) `_request` is a generic function and no assumption should be made wrt the calling function based on the value of `status` and 2) translating exception messages unnecessarily couples this implementation to the backend service, if we decide to change the exception message on the backend this will break - if you are not happy with the received message we should update on the backend.

sfendell-labelbox: I hate parsing error messages too, I just didn't want to have a broader impact than what the ticket specified. I would much prefer to change the messages on ADV side to these. If you don't mind me doing that I'll remove these changes and edit the appropriate errors in ADV, and parse neither 'status' nor 'message' here.

mrobers1982: Sounds good, I can also work with you to update on the ADV side too, let me know.

At the end of the day the exception message in ADV are meant to be human-readable and if they are not we should update them.

libs/labelbox/src/labelbox/exceptions.py
```diff
@@ -15,6 +15,9 @@ def __init__(self, message, cause=None):
     def __str__(self):
         return self.message + str(self.args)
 
+class ADVError(LabelboxError):
```
mrobers1982: nit: Not sure I see the value in surfacing an `ADVError` to the customer as opposed to the more generic error sub-types - it's an implementation detail.

mrobers1982 (dismissed):

libs/labelbox/src/labelbox/exceptions.py
sfendell-labelbox: Good point

mrobers1982 (approved):

PR #1648, opened 2024-06-03
[PLT-999] Vb/chunk by size plt 999

vbrodsky:
# Description

This PR updates our chunking logic for Dataset `create_data_rows` and `upsert_data_rows` to chunk by size of file, not the number of data rows. This will result in generating files of similar size, making processing more predictable and reducing a chance of error or a performance issue due to a very large file

We have set the file size limit to 10MB based on the following considerations:

- in an average case, each row in the files consists of a hash with row_data and, possibly, global or external key. In our estimate the json-encoded size of such a hash is around 500 bytes
- current chunk limit of a file is 10K lines
- we have set the chunk size for a file to be 10MB

Fixes # (issue)

## Type of change

- [x] Internal optimization

## All Submissions

## New Feature Submissions

## Changes to Core Features


libs/labelbox/src/labelbox/schema/internal/descriptor_file_creator.py
```diff
@@ -0,0 +1,327 @@
...
+    This class is used to upload a list of dict and return a url.
+    It will create multiple files if the size of upload the max_chunk_size in bytes,
+        upload the files to gcs in parallel, and return a list of urls
+
+    Args:
+        client (Client): The client object
+        is_upsert (bool): Whether the upload is an upsert. This is a legacy parameter and should always be True because this class will only support upsert
+        max_chunk_size_bytes (int): The maximum size of the file in bytes
+
+    TODO: Remove is_upsert parameter
```
sfendell-labelbox: Can we do this now and remove the TODO? Unless I'm mistaken this doesn't seem very hard to do now. Maybe change the name to `UpsertDescriptorFileCreator` to make it clear that it only supports upserts.

libs/labelbox/src/labelbox/schema/internal/descriptor_file_creator.py
```diff
@@ -0,0 +1,327 @@
...
+    """
+
+    def __init__(self, client: "Client"):
+        self.client = client
+
+    """"
```
sfendell-labelbox: Docstring should go inside method, not above.

libs/labelbox/src/labelbox/schema/internal/descriptor_file_creator.py
```diff
@@ -0,0 +1,327 @@
...
+            return [future.result() for future in as_completed(futures)]
+
+    def create_one(self,
+                   items,
+                   max_attachments_per_data_row=None,
+                   is_upsert=False) -> List[str]:
+        items = self._prepare_items_for_upload(items,
+                                               max_attachments_per_data_row,
+                                               is_upsert)
+        # Prepare and upload the desciptor file
```
sfendell-labelbox: Nit: 'desciptor' -> 'descriptor'

But also, I don't think this comment is really useful at all so I'd just delete it.

libs/labelbox/src/labelbox/schema/internal/descriptor_file_creator.py
```diff
@@ -0,0 +1,327 @@
...
+        return self.client.upload_data(data,
+                                       content_type="application/json",
+                                       filename="json_import.json")
+
+    def _prepare_items_for_upload(self,
+                                  items,
+                                  max_attachments_per_data_row=None,
+                                  is_upsert=False):
+        """
+        This function is shared by `Dataset.create_data_rows`, `Dataset.create_data_rows_sync` and `Dataset.update_data_rows`.
```
sfendell-labelbox: Nit: I'd remove this line, since it's a private method none of these methods actually access, and this comment will go obsolete quickly as we change `Dataset`

libs/labelbox/src/labelbox/schema/internal/descriptor_file_creator.py
```diff
@@ -0,0 +1,327 @@
...
+        >>> dataset.create_data_rows([
+        >>>     {DataRow.row_data:"http://my_site.com/photos/img_01.jpg"},
+        >>>     {DataRow.row_data:"/path/to/file1.jpg"},
+        >>>     "path/to/file2.jpg",
+        >>>     {DataRow.row_data: {"tileLayerUrl" : "http://", ...}}
+        >>>     {DataRow.row_data: {"type" : ..., 'version' : ..., 'messages' : [...]}}
+        >>>     ])
+
+        For an example showing how to upload tiled data_rows see the following notebook:
+            https://github.com/Labelbox/labelbox-python/blob/ms/develop/model_assisted_labeling/tiled_imagery_mal.ipynb
```
sfendell-labelbox: I get a 404 for this url

libs/labelbox/tests/integration/test_data_rows.py
```diff
@@ -2,6 +2,7 @@
 import uuid
 from datetime import datetime
 import json
+import math
```
sfendell-labelbox: Unused?

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
```diff
@@ -1,286 +1,33 @@
...
-            return DataRowUploader.create_descriptor_file(client,
-                                                          chunk,
-                                                          is_upsert=True)
-
-        with ThreadPoolExecutor(file_upload_thread_count) as executor:
-            futures = [
-                executor.submit(_upload_chunk, chunk) for chunk in chunks
-            ]
-            chunk_uris = [future.result() for future in as_completed(futures)]
+        chunk_uris = DescriptorFileCreator(client).create(
```
sfendell-labelbox: This smells "unpythonic" to me. We have this class, that takes in a single other class, but only has one public method, and we instantiate it every time we use it, but we only use it in one place? And that one place is a static method...I know I commented before and you said we have this pattern in SDK, and I believe you, but I still think it's an antipattern and I think this is getting unnecessarily convoluted with OO patterns we don't really need.

Like, if we're instantiating a class every time just to call a single method, we don't really need a class. Let's just make that method a function and export it. 

My thought: A single method, `create_file_descriptor`, that takes in a client , specs, and max_chunk_size, and returns an upload manifest. both descriptor_file_creator and data_row_uploader become wholely stateless, and the file descriptor creation logic can still be shared where it's needed. Wdyt?

libs/labelbox/src/labelbox/schema/internal/descriptor_file_creator.py
sfendell-labelbox: I feel very strongly we should remove this, actually.

libs/labelbox/src/labelbox/schema/internal/descriptor_file_creator.py
```diff
@@ -0,0 +1,327 @@
...
+            futures = [executor.submit(convert_item, item) for item in items]
+            items = [future.result() for future in as_completed(futures)]
+
+        return items
+
+    def _chunk_down_by_bytes(self, items: List[dict],
+                             max_chunk_size: int) -> Generator[str, None, None]:
+        """
+        Recursively chunks down a list of items into smaller lists until each list is less than or equal to max_chunk_size bytes
+        NOTE: of one data row is large than max_chunk_size, it will be returned as one chunk
```
sfendell-labelbox: Nit: `of one data row is large` -> `if one data row is larger`

libs/labelbox/src/labelbox/schema/internal/data_row_uploader.py
```diff
@@ -1,287 +1,30 @@
...
-        with ThreadPoolExecutor(file_upload_thread_count) as executor:
-            futures = [
-                executor.submit(_upload_chunk, chunk) for chunk in chunks
-            ]
-            chunk_uris = [future.result() for future in as_completed(futures)]
-
-        return UploadManifest(source="SDK",
-                              item_count=len(specs),
-                              chunk_uris=chunk_uris)
+    return UploadManifest(source="SDK",
```
sfendell-labelbox: Sorry I missed this in this and previous review, but can this be an enum?

sfendell-labelbox (dismissed):

adrian-chang (approved):

PR #1654, opened 2024-06-05
Provide ability to get model configs with a model id

Gabefire:
# Description

* I noticed there is no way to get model config ids from the UI or easily from SDK without attaching it to a project. So I made a method to get model configs for Live model evaluation projects based on a model id you can get from the UI.

* This allows the flexibility to quickly add them to projects or just manage them in general, which is not even available on the UI yet. 

Loom to explain issue: 
https://www.loom.com/share/d086b0f8b4e5456c97a5077cf395f1b3?sid=e0ac0a7a-a48f-42fa-9725-0ef268a65d5f
## Type of change

## All Submissions

## New Feature Submissions

### DIFF
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -654,6 +654,29 @@ class Client:
         if not result:
             raise labelbox.exceptions.ResourceNotFoundError(Entity.ModelConfig, params)
         return result['deleteModelConfig']['success']
+    
+    def get_model_configs(self, model_id: str) -> List[ModelConfig]:
+        """ Gets all model configs attached to a given model id
+
+        Args:
+            model_id (str): ID of the model associated with the model configs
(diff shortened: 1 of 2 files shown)


Gabefire:
Quick question: What org do tests inside Labelbox-python sdk run against? Is it just the default stageing org or an SDK-specific one? I ask because for this method, I'm trying to test, and I don't think we are cleaning the model configs between other tests. It should be assert two for the valid model ID given, but it has 297 model configs attached.  Tests work against my prod org. basically, I attached some model configs and then tested if it returns two model configs. If you don't mind could I make the [model id fixture](https://github.com/Labelbox/labelbox-python/blob/9dff1c2f5164632289df45966d727d449b98a6f3/libs/labelbox/tests/conftest.py#L1061) delete all the model configs attached to a model when it tears down?

PR #1675, opened 2024-06-12
[PLT-1135] Alias these commands to each other so users don't have to remember which one is valid.

sfendell-labelbox:


Gabefire:
Can you also do this one? https://github.com/Labelbox/labelbox-python/blob/c1a3ac2f3f469eed5a32bb94bfd90e9dcc334d6e/libs/labelbox/src/labelbox/schema/annotation_import.py#L95

adrian-chang (approved):

libs/labelbox/src/labelbox/schema/create_batches_task.py
```diff
@@ -23,6 +23,9 @@ def __init__(self, client, project_id: str, batch_ids: List[str],
             Entity.Task.get_task(self.client, task_id) for task_id in task_ids
         ]
 
+    def wait_until_done(self, timeout_seconds: int = 300) -> None:
+        self.wait_till_done(timeout_seconds)
```
adrian-chang, line 27: Might consider writing a unit tests (e.g.) copy and paste to ensure if they are changed accidentally, nothing breaks

adrian-chang (dismissed):

adrian-chang (approved):

PR #1682, opened 2024-06-18
Vb/model offline eval plt 1107

vbrodsky:
# Description

Story: [SDK support for offline model evaluation use cases
](https://labelbox.atlassian.net/browse/PLT-1107)

This story adds support for creation of offline chat model evaluation projects and verifies the following offline and live chat evalutation properties using sdk integration tests.

For an offline project:
- an ontology can be added
- rows can be added after project creation
- a model config can not be added

For a live chat model evaluation project we now verify that
- rows can not be added after project creation

I refactored create project calls to make sure users will not, inadvertently, cause an issue by supplying incorrect parameters or missing important parameters.

- For each project type - live chat evaluation, offline chat evaluation and 'other' projects there is a separate method
- Each method sets defaults for certain important parameters that a user can not override
- All methods share a common `_create_project()` method

Additionally I have updated handling of EditorTaskType. In a previous release, a new editor task typed cause client get_project() method to break. Now we will set editor task type to `Missing`=None

## Type of change

## All Submissions

## New Feature Submissions

There failing integration tests not related to my code

## Changes to Core Features

- [x] Have you successfully run tests with your changes locally?

### DIFF
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -821,7 +821,8 @@ class Client:
 
     def create_offline_model_evaluation_project(self, **kwargs) -> Project:
         kwargs["media_type"] = MediaType.Conversational
-        kwargs["editor_task_type"] = EditorTaskType.OfflineModelChatEvaluation.value
+        kwargs[
+            "editor_task_type"] = EditorTaskType.OfflineModelChatEvaluation.value
 
         return self.create_project(**kwargs)
 
@@ -900,8 +901,6 @@ class Client:
                 params.pop("dataset_name_or_id", None),
             Field.Boolean("append_to_existing_dataset"):
                 params.pop("append_to_existing_dataset", None),
-            Field.Int("data_row_count"):
-                params.pop("data_row_count", None),
         }
         extra_params = {k: v for k, v in extra_params.items() if v is not None}
 
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -743,6 +743,10 @@ class Client:
         Raises:
             InvalidAttributeError: If the Project type does not contain
                 any of the attribute names given in kwargs.
+
+        NOTE: the following attributes are used only chat model evaluation projects:
+            dataset_name_or_id, append_to_existing_dataset, data_row_count, editor_task_type
+            They are not used for general projects and not supported in this method
         """
         kwargs.pop("dataset_name_or_id", None)
         kwargs.pop("append_to_existing_dataset", None)
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -744,7 +744,6 @@ class Client:
             InvalidAttributeError: If the Project type does not contain
                 any of the attribute names given in kwargs.
         """
-        kwargs.pop("ontology_kind", None)
         kwargs.pop("dataset_name_or_id", None)
         kwargs.pop("append_to_existing_dataset", None)
         kwargs.pop("data_row_count", None)
@@ -823,6 +822,9 @@ class Client:
         kwargs["media_type"] = MediaType.Conversational
         kwargs[
             "editor_task_type"] = EditorTaskType.OfflineModelChatEvaluation.value
+        kwargs.pop("dataset_name_or_id", None)
+        kwargs.pop("append_to_existing_dataset", None)
+        kwargs.pop("data_row_count", None)
 
         return self.create_project(**kwargs)
 
@@ -865,15 +867,6 @@ class Client:
                 "Creating a project without specifying media_type"
                 " through this method will soon no longer be supported.")
             media_type_value = None
-
-        ontology_kind = kwargs.pop("ontology_kind", None)
-        if ontology_kind and OntologyKind.is_supported(ontology_kind):
-            editor_task_type_value = EditorTaskTypeMapper.to_editor_task_type(
-                ontology_kind, media_type).value
-        elif ontology_kind:
-            raise OntologyKind.get_ontology_kind_validation_error(ontology_kind)
-        else:
-            editor_task_type_value = None
 
         quality_mode = kwargs.get("quality_mode")
         if not quality_mode:
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -821,7 +821,7 @@ class Client:
 
     def create_offline_model_evaluation_project(self, **kwargs) -> Project:
         kwargs["media_type"] = MediaType.Conversational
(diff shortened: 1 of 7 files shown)


libs/labelbox/src/labelbox/client.py
```diff
@@ -716,6 +717,8 @@ def create_dataset(self,
             raise e
         return dataset
 
+# **** Create Project begin ****
```
sfendell-labelbox: Remove

libs/labelbox/src/labelbox/client.py
```diff
@@ -740,85 +743,16 @@ def create_project(self, **kwargs) -> Project:
...
-                params.pop("dataset_name_or_id", None),
-            Field.Boolean("append_to_existing_dataset"):
-                params.pop("append_to_existing_dataset", None),
-            Field.Int("data_row_count"):
-                params.pop("data_row_count", None),
-        }
-        extra_params = {k: v for k, v in extra_params.items() if v is not None}
 
-        return self._create(Entity.Project, params, extra_params)
+        NOTE: the following attributes are used only chat model evaluation projects:
```
sfendell-labelbox: "are used only chat" -> "are used only in chat"

libs/labelbox/src/labelbox/client.py
```diff
@@ -716,6 +717,8 @@ def create_dataset(self,
             raise e
         return dataset
 
+# **** Create Project begin ****
+
     def create_project(self, **kwargs) -> Project:
```
sfendell-labelbox: Why does this take kwargs and then specify arg names and typings in the docstring? I know this isn't your doing, but now that the pattern has infected _create_project I think we should actually type these arguments.

In fact this docstring is already out of date, since auto_audit_percentage is a valid arg that's not specified here.

libs/labelbox/src/labelbox/client.py
```diff
@@ -881,13 +815,97 @@ def create_model_evaluation_project(self,
...
+                params.pop("dataset_name_or_id", None),
+            Field.Boolean("append_to_existing_dataset"):
+                params.pop("append_to_existing_dataset", None),
+        }
+        extra_params = {k: v for k, v in extra_params.items() if v is not None}
+
+        return self._create(Entity.Project, params, extra_params)
+
+
+# **** Create Project end ****
```
sfendell-labelbox: Remove this line + extra whitespace

libs/labelbox/src/labelbox/client.py
```diff
@@ -881,13 +815,97 @@ def create_model_evaluation_project(self,
...
-        kwargs["ontology_kind"] = OntologyKind.ModelEvaluation
         kwargs["dataset_name_or_id"] = dataset_name_or_id
         kwargs["append_to_existing_dataset"] = append_to_existing_dataset
         kwargs["data_row_count"] = data_row_count
+        kwargs["editor_task_type"] = EditorTaskType.ModelChatEvaluation.value
+
+        return self._create_project(**kwargs)
+
+    def create_offline_model_evaluation_project(self, **kwargs) -> Project:
+        kwargs["media_type"] = MediaType.Conversational
```
sfendell-labelbox: We should type these arguments and add a docstring

libs/labelbox/src/labelbox/schema/ontology_kind.py
```diff
@@ -26,12 +26,28 @@ def get_ontology_kind_validation_error(cls, ontology_kind):
...
     ResponseCreation = "RESPONSE_CREATION"
+    OfflineModelChatEvaluation = "OFFLINE_MODEL_CHAT_EVALUATION"
     Missing = None
 
     @classmethod
     def is_supported(cls, value):
         return isinstance(value, cls)
 
+    @classmethod
+    def _missing_(cls, name) -> 'EditorTaskType':
```
sfendell-labelbox, line 37: TIL (no action required)

libs/labelbox/src/labelbox/client.py
```diff
@@ -881,13 +815,97 @@ def create_model_evaluation_project(self,
...
+        kwargs[
+            "editor_task_type"] = EditorTaskType.OfflineModelChatEvaluation.value
+        kwargs.pop("dataset_name_or_id", None)
+        kwargs.pop("append_to_existing_dataset", None)
+        kwargs.pop("data_row_count", None)
 
         return self.create_project(**kwargs)
 
+    def _create_project(self, **kwargs) -> Project:
+        """
```
sfendell-labelbox: Kind of a nit, but I don't think any part of this docstring is necessary. It's indicated as internal by the leading underscore, the args should be replaced with real typings (I don't think we generally have docstrings for internal methods), and the note is sort of redundant with the fact that it's internal.

libs/labelbox/src/labelbox/client.py
sfendell-labelbox: Wait sorry, auto_audit_percentage is not really a valid arg, I see why it's not documented...I'm guessing we include it in here for historical reasons. Still we should just type this method signature with keyworded args instead of this.

vbrodsky: Yes, you're absolutely right. Using a TypedDict or a Pydantic model, etc. for the arguments would be a cleaner and more type-safe approach. But I am concerned about the potential risk of introducing such a change within this scope of this specific story/MOE Phase 3. To capture this improvement for future consideration, I propose to create a story under https://labelbox.atlassian.net/browse/PLT-554

libs/labelbox/src/labelbox/client.py
vbrodsky: Re: typed arguments, see my proposal above https://github.com/Labelbox/labelbox-python/pull/1682#discussion_r1645233862
I will add a docstring (forgot to add it). Since these arguments are not user-facing, I propose to clarify them as code comments

libs/labelbox/src/labelbox/client.py
sfendell-labelbox: Fair enough, thanks

libs/labelbox/src/labelbox/client.py
sfendell-labelbox: is `create_offline_model_evaluation_project` not user facing? it's part of the client right?

libs/labelbox/src/labelbox/client.py
sfendell-labelbox: Could you create a ticket and link a comment here (just like //TODO: PLT-### works)

sfendell-labelbox (approved):
LGTM

libs/labelbox/src/labelbox/client.py
vbrodsky: https://labelbox.atlassian.net/browse/PLT-1154

libs/labelbox/src/labelbox/client.py
vbrodsky: it is user facing, but the user is not expected to pass in / set those args

PR #1685, opened 2024-06-20
[PLT-1151] Vb/model setup complete

vbrodsky:
# Description

Provides support for a call to 'complete' chat model evaluation project, thus preventing new model configs from being added / removed. This will make sure sdk can support full model evaluation projects lifecycle on par with the UI

## Type of change

## All Submissions

## New Feature Submissions

## Changes to Core Features

### DIFF
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -840,7 +840,7 @@ class Client:
         kwargs.pop("append_to_existing_dataset", None)
         kwargs.pop("data_row_count", None)
 
-        return self.create_project(**kwargs)
+        return self._create_project(**kwargs)
 
     def _create_project(self, **kwargs) -> Project:
         auto_audit_percentage = kwargs.get("auto_audit_percentage")
--- a/libs/labelbox/src/labelbox/exceptions.py
+++ b/libs/labelbox/src/labelbox/exceptions.py
@@ -152,7 +152,7 @@ class ProcessingWaitTimeout(Exception):
     """Raised when waiting for the data rows to be processed takes longer than allowed"""
 
 
-def _error_message_for_unparsed_graphql_error(error_string: str) -> str:
+def error_message_for_unparsed_graphql_error(error_string: str) -> str:
     """
     Since our client only parses certain graphql errors, this function is used to
     extract the error message from the error string when the error is not
--- a/libs/labelbox/src/labelbox/exceptions.py
+++ b/libs/labelbox/src/labelbox/exceptions.py
@@ -1,3 +1,6 @@
+import re
+
+
 class LabelboxError(Exception):
     """Base class for exceptions."""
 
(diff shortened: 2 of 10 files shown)


libs/labelbox/src/labelbox/client.py
```diff
@@ -839,7 +839,7 @@ def create_offline_model_evaluation_project(self, **kwargs) -> Project:
         kwargs.pop("append_to_existing_dataset", None)
         kwargs.pop("data_row_count", None)
 
-        return self.create_project(**kwargs)
+        return self._create_project(**kwargs)
```
vbrodsky, line 843: fixing typo from a previous PR

libs/labelbox/src/labelbox/exceptions.py
```diff
@@ -147,3 +150,21 @@ class CustomMetricsNotSupportedException(Exception):
 
 class ProcessingWaitTimeout(Exception):
     """Raised when waiting for the data rows to be processed takes longer than allowed"""
+
+
+def _error_message_for_unparsed_graphql_error(error_string: str) -> str:
```
sfendell-labelbox: If we're exporting this and using it outside this module, we should remove the leading underscore

libs/labelbox/src/labelbox/exceptions.py
```diff
@@ -147,3 +150,21 @@ class CustomMetricsNotSupportedException(Exception):
...
+    """
+    Since our client only parses certain graphql errors, this function is used to
+    extract the error message from the error string when the error is not
+    parsed by the client.
+    """
+    # Regex to find the message content
+    pattern = r"'message': '([^']+)'"
```
sfendell-labelbox, line 162: Is this string a json blob with a set structure? If so, can we parse it with json and navigate the object instead of this? This is sorta brittle

sfendell-labelbox: I see, from usage it looks like it's not a json blob...Unfortunate but I understand using this for now

libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -1271,7 +1269,18 @@ def add_model_config(self, model_config_id: str) -> str:
...
             "modelConfigId": model_config_id,
         }
-        result = self.client.execute(query, params)
+        try:
+            result = self.client.execute(query, params)
+        except LabelboxError as e:
+            if e.message.startswith(
+                    "Unknown error: "
+            ):  # unfortunate hack to handle unparsed graphql errors
+                error_content = _error_message_for_unparsed_graphql_error(
```
sfendell-labelbox: Nit: We can push all this logic into the _error_message_for_unparsed_graphql_error function since we use it both (e.g. the function takes in any error, returns extracted content if it starts with `Unknown error:` and has `"message":`, otherwise returns the message. Up to you

libs/labelbox/tests/unit/test_exceptions.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+from labelbox.exceptions import _error_message_for_unparsed_graphql_error
+
+
+@pytest.mark.parametrize('exception_message, expected_result', [
+    ("Unparsed errors on query execution: [{'message': 'Cannot create model config for project because model setup is complete'",
```
sfendell-labelbox: Very small nit: Can we add the closing brackets }] to this message? I know it doesn't affect functionality but i assume the real gql errors have them

sfendell-labelbox (dismissed):

libs/labelbox/src/labelbox/exceptions.py
vbrodsky: right another thing I hope we improve one day

sfendell-labelbox (approved):

PR #1707, opened 2024-07-01
Add upload_type to Project

vbrodsky:
# Description

This PR replaces logic mirrored from the api in a custom method with an attribute set by the api directly. This makes it more reliable

## Type of change

## All Submissions

## New Feature Submissions

## Changes to Core Features

### DIFF
--- a/libs/labelbox/src/labelbox/schema/project.py
+++ b/libs/labelbox/src/labelbox/schema/project.py
@@ -37,7 +37,8 @@ from labelbox.schema.queue_mode import QueueMode
 from labelbox.schema.resource_tag import ResourceTag
 from labelbox.schema.task import Task
 from labelbox.schema.task_queue import TaskQueue
-from labelbox.schema.ontology_kind import (EditorTaskType, OntologyKind)
+from labelbox.schema.ontology_kind import (EditorTaskType, OntologyKind,
+                                           UploadType)
 from labelbox.schema.project_overview import ProjectOverview, ProjectOverviewDetailed
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -145,8 +147,7 @@ def is_chat_evaluation(self) -> bool:
         return self.media_type == MediaType.Conversational and self.editor_task_type == EditorTaskType.ModelChatEvaluation
 
     def is_auto_data_generation(self) -> bool:
-        return self.media_type == MediaType.LLMPromptCreation or self.media_type == MediaType.LLMPromptResponseCreation or self.is_chat_evaluation(
-        )
+        return (self.upload_type == UploadType.Auto) is True
```
adrian-chang: reduntant, return self.upload == UploadType.Auto

Also not sure if it's needed, maybe the field is good enough instead of needing a function

adrian-chang (changes requested):

libs/labelbox/src/labelbox/schema/project.py
vbrodsky: Screenshot 2024-07-01 at 7 36 41 PM
this is what you get if you try to do it directly... our `Field` stuff is not real Bool. Also the method help to wrap it all and since I already have it in place, makes it work better

vbrodsky: @adrian-chang I have updated as you suggested... this lint warning is invalid, I have verified in code self.upload_type is an Enum, not a Field

adrian-chang (approved):

PR #1708, opened 2024-07-02
[PLT-1205] address feedback from part 1

adrian-chang:
# Description

Update error messages to be more descriptive


vbrodsky (approved):
why is this message better? I.e. it is not clear from it what is the cause. Is there no better way to identify the cause?

adrian-chang:
Check line 292 in client, we'd need to change it everywhere for resource not found

PR #1714, opened 2024-07-05
[PLT-1271] Improve annotation import test library

Gabefire:
# Description

* Made huge QOL and improved to annotation import library 
* Added connect_ontology in place of setup method
* Documented different methods

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## All Submissions


libs/labelbox/tests/data/annotation_import/test_conversation_import.py
```diff
@@ -1,45 +0,0 @@
-import uuid
```
Gabefire, line 1: Deleted this file since we are testing already for conversation entity import with test_generic_data_types

libs/labelbox/tests/data/annotation_import/test_data_types.py
```diff
@@ -1,12 +1,8 @@
 import datetime
```
Gabefire, line 1: Heavily reduced the test on this file and focused on unit testing serializing data type labels to NDJSON format. Generic data types tests the importing of NDJSON. I feel like it was a waste to double test this considering how long importing annotations takes when you consider set up and such

Gabefire: Also data types is marked as depercated

adrian-chang (approved):
Ideally this is broken up into more smaller changes to derisk a large change (even though these are tests, better to change them in smaller units of work). Note to consider next time to prevent rework right now.

PR #1724, opened 2024-07-11
[PLT-0] Update to dummy emails

adrian-chang:
# Description

Fix tests to match single pending invite per email

## Type of change

### DIFF
--- a/libs/labelbox/tests/integration/test_user_management.py
+++ b/libs/labelbox/tests/integration/test_user_management.py
@@ -7,7 +7,7 @@ faker = Faker()
 
 def test_org_invite(client, organization, environ, queries):
     role = client.get_roles()['LABELER']
-    dummy_email = "none+{}@labelbox.com".format(faker.random_letters(26))
+    dummy_email = "none+{}@labelbox.com".format("".join(faker.random_letters(26)))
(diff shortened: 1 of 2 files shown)


libs/labelbox/tests/integration/test_user_management.py
```diff
@@ -1,10 +1,13 @@
...
 
+from labelbox import ProjectRole
+from faker import Faker
+
+faker = Faker()
 
 def test_org_invite(client, organization, environ, queries):
     role = client.get_roles()['LABELER']
-    dummy_email = "none@labelbox.com"
+    dummy_email = faker.email()
```
midhun-pm: I just hope this doesn't generate email accounts that exist.

midhun-pm (dismissed):

libs/labelbox/tests/integration/test_user_management.py
midhun-pm: As in, I don't want emails to go out to users.

adrian-chang: Good point, let me use a random string at labelbox

midhun-pm (approved):

PR #1726, opened 2024-07-12
[PLT-1266] Added project and ontology creation for prompt response projects

Gabefire:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context.

* Added support for creating ontologies and projects for prompt response
* Added new classification type since prompt classifications are completely different then normal classifications
* Follows same pattern as modal evualation projects

## Type of change

## All Submissions

## New Feature Submissions

- [x] Does your submission pass tests?

### DIFF
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -891,7 +891,9 @@ class Client:
             **kwargs: Additional parameters to pass see the create_project method
         Returns:
             Project: The created project
-
+            
+        NOTE: Only a dataset_name or dataset_id should be included 
+        
         Examples:
             >>> client.create_prompt_response_generation_project(name=project_name, dataset_name="new data set", project_kind=MediaType.LLMPromptResponseCreation)
             >>>     This creates a new dataset with a default number of rows (100), creates new project and assigns a batch of the newly created datarows to the project.
@@ -908,8 +910,14 @@ class Client:
         """
         if not dataset_id and not dataset_name:
             raise ValueError(
-                "dataset_name or data_set_id must be present and not be an empty string."
+                "dataset_name or dataset_id must be present and not be an empty string."
             )
+        
+        if dataset_id and dataset_name:
+            raise ValueError(
+                "Only provide a dataset_name or dataset_id, not both."
+            )    
+            
         if data_row_count <= 0:
             raise ValueError("data_row_count must be a positive integer.")
 
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -885,8 +885,8 @@ class Client:
 
     @overload
     def create_prompt_response_generation_project(self,
-                                                  dataset_id: str,
                                                   dataset_name: str = None,
+                                                  dataset_id: str,
                                                   data_row_count: int = 100,
                                                   **kwargs) -> Project:
         pass
@@ -903,7 +903,7 @@ class Client:
             dataset_name: When creating a new dataset, pass the name
             dataset_id: When using an existing dataset, pass the id
             data_row_count: The number of data row assets to use for the project
-            **kwargs: Additional parameters to pass to the the create_project method
+            **kwargs: Additional parameters to pass see the create_project method
         Returns:
             Project: The created project
 
@@ -917,7 +917,7 @@ class Client:
             >>> client.create_prompt_response_generation_project(name=project_name, dataset_id="clr00u8j0j0j0", project_kind=MediaType.LLMPromptCreation)
             >>>     This creates a new project, and adds 100 datarows to the dataset with id "clr00u8j0j0j0" and assigns a batch of the newly created data rows to the project.
 
-            >>> client.create_prompt_response_generation_project(name=project_name, dataset_id="clr00u8j0j0j0", data_row_count=10, project_kind=MediaType.LLMPromptCreation)
+            >>> client.create_prompt_response_generation_project(name=project_name, dataset_id="clr00u8j0j0j0", data_row_count=10, project_kind=MediaType.LLMPromptResponseCreation)
             >>>     This creates a new project, and adds 100 datarows to the dataset with id "clr00u8j0j0j0" and assigns a batch of the newly created 10 data rows to the project.
 
         """
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -41,7 +41,7 @@ from labelbox.schema.model import Model
 from labelbox.schema.model_config import ModelConfig
 from labelbox.schema.model_run import ModelRun
 from labelbox.schema.ontology import Ontology, DeleteFeatureFromOntologyResult
-from labelbox.schema.ontology import Tool, Classification, FeatureSchema
+from labelbox.schema.ontology import Tool, Classification, FeatureSchema, PromptResponseClassification
 from labelbox.schema.organization import Organization
 from labelbox.schema.project import Project
 from labelbox.schema.quality_mode import QualityMode, BENCHMARK_AUTO_AUDIT_NUMBER_OF_LABELS, \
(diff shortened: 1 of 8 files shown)


adrian-chang (commented):
This is a fairly large change in an area I'm less familar with, let's have @vbrodsky or @sfendell-labelbox confirm the changes.

Gabefire:
The vast majority of this code is very similar to multimodal chat eval project set up and just testing everything which is also the same as multimodal chat. The only thing different is a need classification type since response and prompts take different params and different name compared to classification

libs/labelbox/src/labelbox/client.py
```diff
@@ -872,6 +872,99 @@ def create_offline_model_evaluation_project(self, **kwargs) -> Project:
         kwargs.pop("data_row_count", None)
 
         return self._create_project(**kwargs)
+    
+    @overload
+    def create_prompt_response_generation_project(self,
```
vbrodsky: You should have 2 overload methods with 2 different set of params, one for the case of new and one for existing data set. `@overload` is just a syntactic sugar for readability

libs/labelbox/src/labelbox/client.py
```diff
@@ -872,6 +872,99 @@ def create_offline_model_evaluation_project(self, **kwargs) -> Project:
...
+                                                  data_row_count: int = 100,
+                                                  **kwargs) -> Project:
+        """
+        Use this method exclusively to create a prompt and response generation project. 
+        
+        Args:
+            dataset_name: When creating a new dataset, pass the name
+            dataset_id: When using an existing dataset, pass the id
+            data_row_count: The number of data row assets to use for the project
+            **kwargs: Additional parameters to pass to the the create_project method
```
vbrodsky: I would add to the **kwargs description something like
See `create_project` documentation for the list of supported parameters

libs/labelbox/src/labelbox/client.py
```diff
@@ -872,6 +872,99 @@ def create_offline_model_evaluation_project(self, **kwargs) -> Project:
...
+            >>> client.create_prompt_response_generation_project(name=project_name, dataset_name="new data set", project_kind=MediaType.LLMPromptResponseCreation)
+            >>>     This creates a new dataset with a default number of rows (100), creates new project and assigns a batch of the newly created datarows to the project.
+
+            >>> client.create_prompt_response_generation_project(name=project_name, dataset_name="new data set", data_row_count=10, project_kind=MediaType.LLMPromptCreation)
+            >>>     This creates a new dataset with 10 data rows, creates new project and assigns a batch of the newly created datarows to the project.
+
+            >>> client.create_prompt_response_generation_project(name=project_name, dataset_id="clr00u8j0j0j0", project_kind=MediaType.LLMPromptCreation)
+            >>>     This creates a new project, and adds 100 datarows to the dataset with id "clr00u8j0j0j0" and assigns a batch of the newly created data rows to the project.
+
+            >>> client.create_prompt_response_generation_project(name=project_name, dataset_id="clr00u8j0j0j0", data_row_count=10, project_kind=MediaType.LLMPromptCreation)
```
vbrodsky: looks like a dupe example... perhaps you want to update it with a different media type?

libs/labelbox/src/labelbox/client.py
```diff
@@ -872,6 +872,99 @@ def create_offline_model_evaluation_project(self, **kwargs) -> Project:
...
+        Returns:
+            Project: The created project
+        """
+        kwargs[
+            "media_type"] = MediaType.Text  # Only Text is supported
+        kwargs[
+            "editor_task_type"] = EditorTaskType.ResponseCreation.value  # Special editor task type for response creation projects
+
+        # The following arguments are not supported for response creation projects
+        kwargs.pop("dataset_name_or_id", None)
```
vbrodsky, line 958: oh so in the case of response creation project, we do not precreate data rows?

libs/labelbox/src/labelbox/client.py
```diff
@@ -1175,10 +1270,13 @@ def create_ontology_from_feature_schemas(
                         f"Tool `{tool}` not in list of supported tools.")
             elif 'type' in feature_schema.normalized:
                 classification = feature_schema.normalized['type']
-                try:
+                if classification in Classification.Type._value2member_map_.keys():
                     Classification.Type(classification)
                     classifications.append(feature_schema.normalized)
-                except ValueError:
+                elif classification in PromptResponseClassification.Type._value2member_map_.keys():
```
vbrodsky, line 1276: why not just add all classification type to the same `Classification` class

libs/labelbox/src/labelbox/schema/ontology_kind.py
```diff
@@ -21,6 +21,30 @@ def get_ontology_kind_validation_error(cls, ontology_kind):
         return TypeError(f"{ontology_kind}: is not a valid ontology kind. Use"
                          f" any of {OntologyKind.__members__.items()}"
                          " from OntologyKind.")
+    
+    @staticmethod
+    def evaluate_ontology_kind_with_media_type(ontology_kind,
+                                               media_type: Union[MediaType, None]) -> Union[MediaType, None]:
```
vbrodsky:
```suggestion
                                               media_type: Optional[MediaType]) -> Union[MediaType, None]:
```

libs/labelbox/src/labelbox/client.py
Gabefire: Done!

libs/labelbox/src/labelbox/client.py
Gabefire: Done!

libs/labelbox/src/labelbox/client.py
Gabefire: Done!

vbrodsky: Here is how they should look like

first overload should have only data_name, not dataset_id
second only dataset_id

or just renove both overloads

libs/labelbox/src/labelbox/schema/ontology.py
```diff
@@ -221,6 +221,183 @@ def add_option(self, option: Option) -> None:
                 f"Duplicate option '{option.value}' "
                 f"for classification '{self.name}'.")
         self.options.append(option)
+        
+@dataclass
+class ResponseOption:
```
mihhail-m: this looks like `Option` but renamed. Any particular reasons for duplication? Couldn't `ResponseOption` be just alias for `Option`?

libs/labelbox/src/labelbox/schema/ontology_kind.py
```diff
@@ -21,6 +21,30 @@ def get_ontology_kind_validation_error(cls, ontology_kind):
...
+                    )
+        
+        elif ontology_kind == OntologyKind.ResponseCreation:
+            if media_type is None:
+                media_type = MediaType.Text
+            else:
+                if media_type is not MediaType.Text:
+                    raise ValueError(
+                        "For response creation, media_type must be Text."
+                    )
```
mihhail-m: this can be greatly simplified and the level of unnecessary `if` nesting can be removed. Quick gpt prompting gives a nice suggestion for this:

```python
ontology_to_media = {
    OntologyKind.ModelEvaluation: (MediaType.Conversational, "For chat evaluation, media_type must be Conversational."),
    OntologyKind.ResponseCreation: (MediaType.Text, "For response creation, media_type must be Text.")
}

if ontology_kind in ontology_to_media:
    expected_media_type, error_message = ontology_to_media[ontology_kind]

    if media_type is None:
        media_type = expected_media_type
    else:
        raise ValueError(error_message)
```

Gabefire: Clever! Added

libs/labelbox/src/labelbox/schema/ontology.py
Gabefire: I added it for consistency to separate it from normal classifications. Prompt response classifications are different from normal classifications (names and some params), and I want users to understand that, but I switched it to an alias since you are right. It has a lot of duplication.

libs/labelbox/src/labelbox/schema/ontology_kind.py
Gabefire: I did add on to this:
```python
    if media_type is None or media_type == expected_media_type:
        media_type = expected_media_type
    else:
        raise ValueError(error_message)
```
 This is added for users who already gave the correct expected_media_type through how it is set up, but just in case.

vbrodsky (approved):

mihhail-m (approved):

PR #1727, opened 2024-07-13
Huge optimization to annotation import library 

Gabefire:
# Description

Please include a summary of the changes and the related issues. Please also include relevant motivation and context.

* Swapped a long integration label import test for a unit test. As a comparison against staging, each of these tests can take up to 5 minutes, each now takes a few seconds. 
* Tests if a validation error was thrown while creating an object. My pr just tests the function that is used to validate with the annotation NDJSON improvements I made
* Removed a couple of unneeded tests and will add comments on why
* Also, this validation seems off. It validates the name of the annotation. You can still go over 5000 annotations as long as they all have different names; this is just validating there is not 5k of the same annotation on the label. Im not sure if that is intended

## Type of change

## All Submissions


libs/labelbox/tests/data/annotation_import/test_bulk_import_request.py
```diff
@@ -219,110 +219,3 @@ def test_delete(module_project, predictions):
...
-                                           value=Rectangle(
-                                               start=Point(x=100, y=100),
-                                               end=Point(x=105, y=105),
-                                           ))
-                 ])
-
-
-@pytest.mark.parametrize("configured_project", [MediaType.Video], indirect = True)
-@patch('labelbox.schema.annotation_import.ANNOTATION_PER_LABEL_LIMIT', 20)
-def test_below_annotation_limit_on_single_data_row(
```
Gabefire, line 242: Remove this we are constantly testing for this case everytime we do a label import

libs/labelbox/tests/data/annotation_import/test_bulk_import_request.py
```diff
@@ -219,110 +219,3 @@ def test_delete(module_project, predictions):
...
-        project_id=configured_project.uid,
-        name=f"import {str(uuid.uuid4())}",
-        predictions=labels)
-
-    assert import_annotations.errors == []
-
-
-@pytest.mark.parametrize("configured_project", [MediaType.Video], indirect = True)
-@patch('labelbox.schema.annotation_import.ANNOTATION_PER_LABEL_LIMIT', 20)
-def test_above_annotation_limit_divided_among_labels_on_one_row(
```
Gabefire, line 306: This seems to do almost the same thing above, so it was removed. I can add this back in as a unit test instead if wanted

adrian-chang (approved):

PR #1759, opened 2024-07-31
[PLT-600] Remove hacks and simplified annotation import library 

Gabefire:
# Description

* Pulling pieces out of Pydantic v2 converstion. This one is REQUIRED for pydantic v2 because it does not support a previous [hack](https://github.com/Labelbox/labelbox-python/blob/37f4386fb15a151317fcfcb95d1f983dff136be5/libs/labelbox/src/labelbox/data/serialization/ndjson/objects.py#L776-L778)
* This solution replaces the hack mentioned with a registry system that infers base models. This is still slightly complicated, but I put plenty of comments. It is also less complicated than depending on the order of models for stuff to work or having complicated parts of the code spread among multiple files. (it's contained to this one area) 
* All test run without changing anything inside the test library
* I think getting base models from dicts is unnecessary, but we need to keep backwards compat

## Type of change

## All Submissions


vbrodsky:
Migrated diffs to https://github.com/Labelbox/labelbox-python/pull/1738 and closing this PR

PR #1771, opened 2024-08-18
Vb/search projects filters plt 1399

vbrodsky:
# Description

Add search filters to project list

Story: https://labelbox.atlassian.net/browse/PLT-1399

Methods updated:
- Client get_labeling_service_dashboards
- LabelingServiceDashboard.get_all

Adds API/ES search filters as pydantic classes

## Type of change

## All Submissions

## New Feature Submissions

## Changes to Core Features

- [x] Have you successfully run tests with your changes locally?

### DIFF
--- a/libs/labelbox/src/labelbox/schema/search_filters.py
+++ b/libs/labelbox/src/labelbox/schema/search_filters.py
@@ -107,16 +107,24 @@ class WorkforceStageUpdatedRangeFilter(BaseSearchFilter):
     value: DateRangeValue
 
 
-SearchFilters = Union[OrganizationFilter, WorkspaceFilter, TagFilter,
-                      ProjectStageFilter, WorkforceRequestedDateFilter,
-                      WorkforceStageUpdatedFilter,
-                      WorkforceRequestedDateRangeFilter,
-                      WorkforceStageUpdatedRangeFilter]
+SearchFilter = Union[OrganizationFilter, WorkspaceFilter, TagFilter,
+                     ProjectStageFilter, WorkforceRequestedDateFilter,
+                     WorkforceStageUpdatedFilter,
+                     WorkforceRequestedDateRangeFilter,
+                     WorkforceStageUpdatedRangeFilter]
 
 
-def _build_search_filter(filter: List[SearchFilters]):
-    operation_types = {f.operation for f in filter}
-    if len(operation_types) < len(filter):
-        raise ValueError("Only one filter per operation type is allowed")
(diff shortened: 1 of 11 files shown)


tomislav-peharda (dismissed):

tomislav-peharda (dismissed):

libs/labelbox/src/labelbox/schema/search_filters.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+SearchFilter = Union[OrganizationFilter, WorkspaceFilter, TagFilter,
+                     ProjectStageFilter, WorkforceRequestedDateFilter,
+                     WorkforceStageUpdatedFilter,
+                     WorkforceRequestedDateRangeFilter,
+                     WorkforceStageUpdatedRangeFilter]
+
+
+def _dict_to_graphql_string(d: Union[dict, list]) -> str:
```
sfendell-labelbox: How is this different from json.dumps? and if we still need it, can d be Union[dict, list, str, number] (or just any?) for more accurate type?

sfendell-labelbox: Oh I see, keys don't have quotes around them as they do in json dumps

libs/labelbox/src/labelbox/schema/search_filters.py
```diff
@@ -0,0 +1,187 @@
...
+    if isinstance(d, dict):
+        return "{" + ", ".join(
+            f'{k}: {_dict_to_graphql_string(v)}' for k, v in d.items()) + "}"
+    elif isinstance(d, list):
+        return "[" + ", ".join(
+            _dict_to_graphql_string(item) for item in d) + "]"
+    else:
+        return f'"{d}"' if isinstance(d, str) else str(d)
```
sfendell-labelbox, line 179: Do we need to surround numbers with quotations too? Maybe doesn't matter currently since I don't think any of the filters currently use one, but if we do we should do that now

sfendell-labelbox (dismissed):

libs/labelbox/src/labelbox/schema/search_filters.py
vbrodsky: the quotations are only if `isinstance(d, str)`. You are right we need to test this code with numbers. Currently we do not have any, but I will do it in a follow up story for a new filter we are creating

tomislav-peharda (approved):

PR #1796, opened 2024-09-09
Removed pydantic V2 protected messages to avoid unnecessary warnings to users

Gabefire:
# Description

* If a Pydantic model has an attribute beginning with `model_`, it will spam the user with a protected message. You can avoid this message by inside the config dict https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.protected_namespaces
* Also copy_on_model_validation is removed as a option entirely in Pydantic V2 so removed this to avoid unnecessary warnings https://docs.pydantic.dev/latest/migration/#changes-to-config

## Type of change

## All Submissions

### DIFF
--- a/libs/labelbox/src/labelbox/data/annotation_types/data/raster.py
+++ b/libs/labelbox/src/labelbox/data/annotation_types/data/raster.py
@@ -24,7 +24,8 @@ class RasterData(BaseModel, ABC):
     uid: Optional[str] = None
     global_key: Optional[str] = None
     arr: Optional[TypedArray[Literal['uint8']]] = None
-    model_config = ConfigDict(extra="forbid", copy_on_model_validation="none")
+    
+    model_config = ConfigDict(extra="forbid")
 
     @classmethod
     def from_2D_arr(cls, arr: Union[TypedArray[Literal['uint8']],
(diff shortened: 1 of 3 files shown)


libs/labelbox/src/labelbox/data/annotation_types/data/raster.py
```diff
@@ -24,7 +24,8 @@ class RasterData(BaseModel, ABC):
     uid: Optional[str] = None
     global_key: Optional[str] = None
     arr: Optional[TypedArray[Literal['uint8']]] = None
-    model_config = ConfigDict(extra="forbid", copy_on_model_validation="none")
+    
+    model_config = ConfigDict(extra="forbid")
```
vbrodsky, line 28: why was `copy_on_model_validation` removed

Gabefire: image
Pydantic V2 removed it :( with no alternative

Gabefire: It throws a warning to users if you import this library and technically does nothing

Gabefire: image
This is what it used to do. We don't have a way to add this behavior anymore but Im not sure why the original author added this setting it is the only place that includes this in the config settings

vbrodsky (approved):

PR #1800, opened 2024-09-10
[PLT-1392] [PLT-1350] Remove SDK methods for exports v1

tomislav-peharda:


vbrodsky (approved):

vbrodsky (changes requested):
@tomislav-peharda I think Project label_generator() ought to be removed, it uses the methods you have deleted behind scenes

vbrodsky (approved):

adrian-chang (approved):

PR #1806, opened 2024-09-11
[PLT-1495] Migrated sdk to use Ruff for formattor

Gabefire:
# Description

* Swapping Yapf to Ruff, which is built into Rye
* Ruff is 100x times faster then Yapf with more features
* The reason why every file was touched is people have not been using Yapf, our checker has not been enforcing it in the workflow. If you run rye run yapf-lint it touchs almost all files so I don't think it has been running.
* I added on to the limiting workflow to start enforcing the formattor 

## Type of change


libs/labelbox/pyproject.toml
```diff
@@ -87,9 +89,8 @@ unit = "pytest tests/unit"
 # LABELBOX_TEST_BASE_URL="http://host.docker.internal:8080" \
 integration = { cmd = "pytest tests/integration" }
 data = { cmd = "pytest tests/data" }
-yapf-lint = "yapf tests src -i --verbose --recursive --parallel --style \"google\""
 mypy-lint = "mypy src --pretty --show-error-codes --non-interactive --install-types"
-lint = { chain = ["yapf-lint", "mypy-lint"] }
```
adrian-chang, line 92: Should we add fmt to the chain here?

adrian-chang (changes requested):

libs/labelbox/pyproject.toml
Gabefire: Yea added this and removed my extra workflow job

adrian-chang (approved):

PR #1810, opened 2024-09-12
Added experiment to execute client for user_groups

### LINKED ISSUE #1808: User Group Management doesn't support `experimental_enabled` on Client
# Describe the bug

The `Client` class has the `enable_experimental` argument, but the `execute()` method does not check the `self.enable_experimental` value.

This means that `labelbox.schema.user_group.UserGroup` class's methods ignore the `enable_experimental` even though it's defined as required by that class.

# To reproduce

1. Use method `labelbox.schema.user_group.UserGroup()`
2. Pass parameters `client=labelbox.Client(api_key="...", enable_experimental=True)`
3. See error: `labelbox.exceptions.NetworkError: HTTPSConnectionPool(host='api.labelbox.com', port=443): Max retries exceeded with url: /graphql`

# Expected behavior

The error should be indicating the experimental URL: `/_gql`

# Screenshots / Videos

N/A

# Environment (please complete the following information
 - OS: `macOS`
 - Python Version: `3.11.x`
 - SDK + SDK Version: `labelbox@4.0.0`
 - Python Environment: `poetry.lock`

# Additional context

N/A

# Proposed Fix

Every call to `self.client.execute(...)` like here:

https://github.com/Labelbox/labelbox-python/blob/45109d855c32796a309eae23a9dc53750c3ae9a5/libs/labelbox/src/labelbox/schema/user_group.py#L141

could be updated with `self.client.execute(..., experimental=self.client.enable_experimental)`

Gabefire:
# Description

* See the attached issue. The user is complaining about the endpoint that is used for user groups. 
* TLDR: client.execute has an experiment parameter outside the client. Setting the enable_experiment parameter swaps the endpoint used in the SDK from "/graphql" to "/_gql." From my understanding of these different URLs, they don't mean much anymore. The user must have some SSL certificate verification check in his local environment that monitors if the experiment is turned on for the correct URL because his other issue he has submitted is similar.
* I just changed what is passed to the experiment parameter inside user_groups for consistency. I thought about slightly modifying client.execute to override the experiment parameter if enabled_experiment is its own, but not sure of downstream effects or the exact reason the two experiment places are different (maybe some legacy feature)

### DIFF
--- a/libs/labelbox/src/labelbox/schema/user_group.py
+++ b/libs/labelbox/src/labelbox/schema/user_group.py
@@ -6,7 +6,11 @@ from labelbox import Client
 from labelbox.exceptions import ResourceCreationError
 from labelbox.schema.user import User
 from labelbox.schema.project import Project
-from labelbox.exceptions import UnprocessableEntityError, MalformedQueryException, ResourceNotFoundError
+from labelbox.exceptions import (
+    UnprocessableEntityError,
+    MalformedQueryException,
+    ResourceNotFoundError,
+)
 from labelbox.schema.queue_mode import QueueMode
 from labelbox.schema.ontology_kind import EditorTaskType
 from labelbox.schema.media_type import MediaType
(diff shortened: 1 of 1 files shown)


Gabefire:
https://github.com/Labelbox/labelbox-python/issues/1808

PR #1811, opened 2024-09-12
Added merge environmental settings to prepared request

### LINKED ISSUE #1809: Labelbox Client usage of `requests` Library doesn't support the `verify` argument so it ignores `REQUESTS_CA_BUNDLE` Environment Variable
# Describe the bug

In the `execute()` method of the `Client` class, the low-level `requests.Request` class is instantiated to create a `requests.PreparedRequest`, but this bypasses the `requests.merge_environment_settings()` method.

This means that `Client.execute()` will not collect and use the `REQUESTS_CA_BUNDLE` environment variable if present, and will result in SSL errors.

# To reproduce

1. Use method `client_obj = labelbox.Client(api_key="...")`
2. Pass parameters `client_obj.execute(...)`
3. See error: `labelbox.exceptions.NetworkError: HTTPSConnectionPool(host='api.labelbox.com', port=443): Max retries exceeded with url: /_gql (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1006)')))`

# Expected behavior

The `REQUESTS_CA_BUNDLE` if set properly should allow the SSL certificate chain to be verified through a proxy or other deferred networking path where a custom certificate chain is needed.

# Screenshots / Videos

N/A

# Environment (please complete the following information
 - OS: `macOS`
 - Python Version: `3.11.x`
 - SDK + SDK Version: `labelbox@4.0.0`
 - Python Environment: `poetry.lock`

# Additional context

N/A

# Proposed Fix

In the `Client.execute(...)` method at the lines:

https://github.com/Labelbox/labelbox-python/blob/45109d855c32796a309eae23a9dc53750c3ae9a5/libs/labelbox/src/labelbox/client.py#L223-L225

the update could be:

```python
            prepped: requests.PreparedRequest = request.prepare()

            response = self._connection.send(
                prepped,
                timeout=timeout,
                verify=os.getenv("REQUESTS_CA_BUNDLE", None),
            )
```

However, a more robust fix could be like this, though this wouldn't allow direct pass-through setting of the values, it would only rely on the Environment Variables:

```python
            prepped: requests.PreparedRequest = request.prepare()

            req_settings: dict = self._connection.merge_environment_settings(
                url=None,
                proxies=None,
                stream=None,
                verify=None,
                cert=None,
            )

            response = self._connection.send(
                prepped,
                timeout=timeout,
                **req_settings,
            )
```

Gabefire:
# Description

* User mentions they are wanting to use self-signed SSL certs for the SDK but unable to do so since our SDK does not take account of the environmental variable `REQUESTS_CA_BUNDLE` 
* reference these [docs](https://requests.readthedocs.io/en/latest/user/advanced/#prepared-requests) for more details 
* But I essentially implemented the recommended solution


libs/labelbox/src/labelbox/client.py
```diff
@@ -214,15 +214,20 @@ def convert_value(value):
...
             request = requests.Request('POST',
                                        endpoint,
                                        headers=headers,
                                        data=data,
                                        files=files if files else None)
 
-            prepped: requests.PreparedRequest = request.prepare()
-
-            response = self._connection.send(prepped, timeout=timeout)
+            prepped: requests.PreparedRequest = self._connection.prepare_request(
```
vbrodsky: I do not believe we should be using `prepare_request` on the Session itself... unfortunately so. This is because we are modifying headers above and if you use prepare_request on the session it will prb merge those headers back.

vbrodsky (changes requested):

libs/labelbox/src/labelbox/client.py
Gabefire: So just to test I printed out both the request headers and prepped headers
```
{
    "User-Agent": "python-requests/2.32.3",
    "Accept-Encoding": "gzip, deflate",
    "Accept": "application/json",
    "Connection": "keep-alive",
    "Authorization": "Bearer <API_KEY>",
    "Content-Type": "application/json",
    "X-User-Agent": "python-sdk 4.0.0",
    "X-Python-Version": "3.8.18-final",
}
{
    "User-Agent": "python-requests/2.32.3",
    "Accept-Encoding": "gzip, deflate",
    "Accept": "application/json",
    "Connection": "keep-alive",
    "Authorization": "Bearer <API_KEY>",
    "Content-Type": "application/json",
    "X-User-Agent": "python-sdk 4.0.0",
    "X-Python-Version": "3.8.18-final",
    "Content-Length": "733",
}
```

Gabefire: Ahh I see what your saying. I modified it to add the correct headers to the prepped response. I believe that is what the docs recommended to allow the merge_environment_settings

Gabefire: So I swapped it to add the headers to the prepped response instead of on the request maintaining the correct headers

Gabefire: Moved this back to just prepped but did the merge setting still

vbrodsky (approved):
@Gabefire we really need to confirm this via testing... how can we do that?

PR #1818, opened 2024-09-16
[PLT-1463] Remove deserialize completely

Gabefire:
# Description

* Removed NDJsonConverter.deserialize from sdk
* Notes on changes to some tests included


libs/labelbox/tests/data/annotation_import/test_generic_data_types.py
```diff
@@ -28,76 +28,6 @@ def validate_iso_format(date_string: str):
...
-        (MediaType.Text, GenericDataRowData),
-        (MediaType.Video, GenericDataRowData),
-        (MediaType.Conversational, GenericDataRowData),
-        (MediaType.Document, GenericDataRowData),
-        # (MediaType.LLMPromptResponseCreation, GenericDataRowData),
-        # (MediaType.LLMPromptCreation, GenericDataRowData),
-        (OntologyKind.ResponseCreation, GenericDataRowData),
-    ],
-)
-def test_generic_data_row_type_by_global_key(
```
Gabefire, line 81: Removed the generic tests on this file in place of dedicated unit tests libs/labelbox/tests/data/serialization/ndjson/test_generic_data_row_data.py

libs/labelbox/tests/data/annotation_import/test_data_types.py
Gabefire, line 1: Remove this test file completely since we are removing data classes besides the generic one in V6

libs/labelbox/tests/data/annotation_import/test_bulk_import_request.py
```diff
@@ -87,27 +60,6 @@ def test_create_from_objects(
     )
 
 
-def test_create_from_label_objects(
```
Gabefire: Remove this test completely since we are removing bulk import in V6. This would be complicated to get to work, so not worth our time to fix

.github/workflows/lbox-develop.yml
```diff
@@ -2,9 +2,9 @@ name: LBox Develop
 
 on:
   push:
-    branches: [develop]
+    branches: [develop, v6]
```
Gabefire, line 5: Add our V6 branch here just so tests can run we can remove once V6 is merged onto develop

libs/labelbox/tests/data/annotation_import/test_bulk_import_request.py
vbrodsky: I will have a separate PR on BulkImportRequest

libs/labelbox/tests/data/serialization/ndjson/test_generic_data_row_data.py
```diff
@@ -0,0 +1,79 @@
+from labelbox.data.annotation_types.data.generic_data_row_data import (
```
vbrodsky, line 1: Question: on the removed test, we tested all annotation types `annotations_by_media_type`. But here we just test with 2?

Gabefire: I think testing withevery annotation was a bit overkill and would be complicated to add in without deserialization existing. I would have to rewrite that `annotations_by_media_type` fixture to provide Pydantic models over just the JSON and that fixture is pretty large. I wanted to simplify it by just seeing if the GenericDataRow serializes correctly. We also test for the GenericDataRow type with just the dictionary data={"global_key":global_key} and data={"uid":data_row_id} in our end to end testing

Gabefire: Id have to rewrite most of this file https://github.com/Labelbox/labelbox-python/blob/develop/libs/labelbox/tests/data/annotation_import/conftest.py to make it work :disappointed: specifically the interfaces. Since we cant deserialize the JSON anymore directly to pydantic objects.

vbrodsky (commented):
Can you add a story to this epic as a tech debt to add more unit testing for annotation serialization... perhaps Co-pilot can help to reduce work

vbrodsky (approved):

PR #1822, opened 2024-09-17
[PLT-1506] Added Ruff linting to SDK

Gabefire:
# Description
* This adds in the Ruff Linter that is the same package as the Ruff formator we now use
* if this merges SDK devs will HAVE to use Ruff for linting this is the same tool for formatting
* Note mypy type checker is separate from this linter 
* Using default rules besides Flake which is showing 300+ violations right now mostly unused imports. I also ignore one rule for bare excepts since that is a rule we break a lot
* View this doc for all [rules](https://docs.astral.sh/ruff/rules/)


vbrodsky:
@Gabefire is it linting? or removing ND deserialization? Or both?

libs/labelbox/src/labelbox/data/serialization/ndjson/base.py
```diff
@@ -8,18 +8,6 @@
 
 from ....annotated_types import Cuid
 
-subclass_registry = {}
-
-
-class _SubclassRegistryBase(BaseModel):
```
vbrodsky: Yeah!!!

vbrodsky (approved):
Great job

vbrodsky (changes requested):
Had to remove approval due to failing tests for p3.12 version

libs/labelbox/pyproject.toml
```diff
@@ -74,6 +74,10 @@ dev-dependencies = [
 [tool.ruff]
 line-length = 80
 
+[tool.ruff.lint]
+ignore = ["F", "E722"]
+exclude = ["**/__init__.py"]
```
Gabefire, line 79: This needs to be excluded for future rules (no unused imports)

libs/labelbox/pyproject.toml
```diff
@@ -74,6 +74,10 @@ dev-dependencies = [
 [tool.ruff]
 line-length = 80
 
+[tool.ruff.lint]
+ignore = ["F", "E722"]
```
Gabefire, line 78: I have PyFlake and a bareexcept rule ignored here. Later we should look at not ignoring these rules and possiblly include more then just the defaults.

libs/labelbox/pyproject.toml
```diff
@@ -89,9 +93,10 @@ unit = "pytest tests/unit"
 # LABELBOX_TEST_BASE_URL="http://host.docker.internal:8080" \
 integration = { cmd = "pytest tests/integration" }
 data = { cmd = "pytest tests/data" }
+rye-lint = "rye lint"
 rye-fmt-check = "rye fmt --check"
 mypy-lint = "mypy src --pretty --show-error-codes --non-interactive --install-types"
-lint = { chain = ["mypy-lint", "rye-fmt-check"] }
+lint = { chain = ["rye-fmt-check", "mypy-lint", "rye-lint"] }
```
vbrodsky, line 99: do we still need mypy here?

vbrodsky (approved):

libs/labelbox/pyproject.toml
Gabefire: Yea I think so Mypy is a [type checker](https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-mypy-or-pyright-or-pyre)

PR #1830, opened 2024-09-18
[PLT-1485] Removed Datatypes

Gabefire:
# Description

* Removed datatypes from SDK besides generic and mask data (needed for mask annotations) and geotile
Geotile has to stay because it has parameters that are need for the label like zoom level


libs/labelbox/src/labelbox/data/annotation_types/collection.py
```diff
@@ -40,26 +40,6 @@ def _assign_ids(label: Label):
         self._fns["assign_feature_schema_ids"] = _assign_ids
         return self
 
-    def add_url_to_data(
```
Gabefire, line 43: Removed this function. This is not supported by GenericDataType and I do not think this is relevant any longer. Check doc string for what it does

libs/labelbox/tests/data/annotation_types/test_collection.py
```diff
@@ -91,37 +93,23 @@ def test_adding_schema_ids():
     assert next(generator).annotations[0].feature_schema_id == feature_schema_id
 
 
-def test_adding_urls(signer):
```
Gabefire, line 94: Removed due to removing function

libs/labelbox/src/labelbox/data/annotation_types/label.py
```diff
@@ -128,19 +90,6 @@ def frame_annotations(
                 frame_dict[annotation.frame].append(annotation)
         return frame_dict
 
-    def add_url_to_data(self, signer) -> "Label":
```
Gabefire, line 131: This is only applicable for text data types which no longer exists in this PR

libs/labelbox/src/labelbox/data/annotation_types/collection.py
```diff
@@ -40,47 +40,6 @@ def _assign_ids(label: Label):
...
-        """
-
-        def _add_url_to_data(label: Label):
-            label.add_url_to_data(signer)
-            return label
-
-        self._fns["add_url_to_data"] = _add_url_to_data
-        return self
-
-    def add_to_dataset(
```
Gabefire, line 63: Same concept of add_url_to_data. This was only used in TextData and GenericDataRow does not support this method

Gabefire: I can add it to GenericDataRow but I think methods like this were never used and are complicated

libs/labelbox/src/labelbox/data/serialization/ndjson/classification.py
```diff
@@ -232,7 +232,7 @@ def from_common(
         name: str,
         feature_schema_id: Cuid,
         extra: Dict[str, Any],
-        data: Union[TextData, ImageData],
+        data: Union[GenericDataRowData],
```
vbrodsky: do we really need `Union` here?

libs/labelbox/src/labelbox/data/serialization/ndjson/classification.py
```diff
@@ -264,7 +264,7 @@ def from_common(
         name: str,
         feature_schema_id: Cuid,
         extra: Dict[str, Any],
-        data: Union[VideoData, TextData, ImageData],
+        data: Union[GenericDataRowData],
```
vbrodsky: Union?

vbrodsky (approved):
left 2 tiny comments

PR #1838, opened 2024-09-23
[PLT-1471] Remove pydantic v2 deprecate warnings

Gabefire:
# Description

* Removed pydantic warnings from SDK. 
* Removed datetime.utcnow from test which is also deprecated


libs/labelbox/src/labelbox/data/annotation_types/types.py
```diff
@@ -5,19 +5,22 @@
...
 
 DType = TypeVar("DType")
 DShape = TypeVar("DShape")
 
 
 class _TypedArray(np.ndarray, Generic[DType, DShape]):
     @classmethod
-    def __get_validators__(cls):
-        yield cls.validate
+    def __get_pydantic_core_schema__(
```
Gabefire, line 17: I have a story to simplify this class

libs/labelbox/src/labelbox/schema/data_row_metadata.py
```diff
@@ -49,7 +56,7 @@ class DataRowMetadataKind(Enum):
 # Metadata schema
 class DataRowMetadataSchema(BaseModel):
     uid: SchemaId
-    name: str = Field(strip_whitespace=True, min_length=1, max_length=100)
```
Gabefire, line 52: strip_whitespace is depercated in pydantic V2 and gives warnings to users

libs/labelbox/src/labelbox/schema/data_row_metadata.py
```diff
@@ -36,6 +37,12 @@
     format_iso_from_string,
 )
 
+Name = Annotated[
```
vbrodsky, line 40: good

vbrodsky (approved):
Can we also remove some more warnings re 
DeprecationWarning: datetime.datetime.utcnow() 

Also can you squash these 53 commits?
Also can you add to the end of the epic to make sure we remove / do not have sdk deprecation warnings

PR #1839, opened 2024-09-24
[PLT-1489] Removed get stream and corresponding classes

Gabefire:
# Description

* get_stream is deprecated
* Removed get_stream and corresponding classes for just get_buffered_stream


mihhail-m:
hey @Gabefire just want to double check as when `get_stream` got deprecated? Because currently this is the way Labelbox application suggests users to download their exports, especially for the big projects, datasets etc. Is the substitution for this deprecation already in-place or being developed?

Screenshot 2024-09-24 at 14 00 08

Gabefire:
> hey @mihhail-m  just want to double check as when `get_stream` got deprecated? Because currently this is the way Labelbox application suggests users to download their exports, especially for the big projects, datasets etc. Is the substitution for this deprecation already in-place or being developed?
> 
> Screenshot 2024-09-24 at 14 00 08

Yeah, we need to write a ticket to replace that code block inside the UI. But `get_stream` is deprecated along with all its file converters in the SDK and our docs. https://docs.labelbox.com/reference/export-overview#streaming-results-and-errors The replaced method is `get_buffered_stream` we are cutting a new version of the SDK V6 specifically to remove all deprecated methods and get a fresh slate

vbrodsky (approved):

PR #1843, opened 2024-09-24
[PLT-1490] Removed project.setup

Gabefire:
# Description

* Removed project.setup and migrated tests to connect_ontology

### DIFF
--- a/libs/labelbox/tests/data/export/conftest.py
+++ b/libs/labelbox/tests/data/export/conftest.py
@@ -117,13 +117,6 @@ def ontology(client: Client):
         "color": "#008941",
         "classifications": [],
     }
-    entity_tool = {
-        "required": False,
-        "name": "entity--",
-        "tool": "named-entity",
-        "color": "#006FA6",
-        "classifications": [],
-    }
     raster_segmentation_tool = {
         "required": False,
         "name": "segmentation_mask",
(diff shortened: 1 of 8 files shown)


libs/labelbox/tests/integration/conftest.py
```diff
@@ -80,24 +80,27 @@ def project_pack(client):
...
 @pytest.fixture
-def project_with_empty_ontology(project):
-    editor = list(
-        project.client.get_labeling_frontends(
-            where=LabelingFrontend.name == "editor"
-        )
-    )[0]
-    empty_ontology = {"tools": [], "classifications": []}
-    project.setup(editor, empty_ontology)
+def project_with_one_feature_ontology(project, client: Client):
```
Gabefire, line 80: Can not have ontologies with no features and use the connect_ontology method

libs/labelbox/tests/data/export/conftest.py
```diff
@@ -118,18 +118,174 @@ def ontology():
...
+    ontology = client.create_ontology(
+        "image ontology",
+        {"tools": tools, "classifications": classifications},
+        MediaType.Image,
+    )
+    return ontology
+
+
+@pytest.fixture
+def video_ontology(client: Client):
```
Gabefire, line 189: I had to make a specific video ontology with the correct features. Since you can not have ontologies anymore with all features

Gabefire: We also might move away from not having a media type specified with ontology

libs/labelbox/tests/data/export/conftest.py
```diff
@@ -118,18 +118,174 @@ def ontology():
         "color": "#008941",
         "classifications": [],
     }
-    entity_tool = {
+    raster_segmentation_tool = {
```
vbrodsky, line 121: Why was this change, i.e. removed entity_tool and segmentation_tool

libs/labelbox/tests/data/export/conftest.py
vbrodsky: Do we need a story for that

vbrodsky (changes requested):
Minor questions... but the issues is tests failing for pythong 3.12 (latest code)

libs/labelbox/tests/data/export/conftest.py
Gabefire: Good point ill write a ticket. But I think we should I was talking to perception and we have had issues revolving around media types not being in ontologies

libs/labelbox/tests/data/export/conftest.py
Gabefire: Oh so regular masks (raster-segmentation) are here I just removed super pixel because it is not only deprecated but completed removed in the app so no point of making ontologies with it anymore. Entity I had to remove this ontology is meant for images and images don't support entity. The old workflow was to just add every single option to an ontology. You cant really do that anymore with regular connect_editor

vbrodsky (approved):

PR #1845, opened 2024-09-25
[PLT-46] Add method / heurisrtics to extract original sdk method name from python trace

vbrodsky:
# Description

This is passed on to the api in the header and from there logged to DD in order to build dashboards

Fixes # (issue)

## Type of change

## All Submissions

## New Feature Submissions

## Changes to Core Features

### DIFF
--- a/libs/lbox-clients/src/lbox/request_client.py
+++ b/libs/lbox-clients/src/lbox/request_client.py
@@ -6,7 +6,7 @@ import re
 import sys
 from datetime import datetime, timezone
 from types import MappingProxyType
-from typing import Callable, Dict, Optional
+from typing import Callable, Dict, Optional, TypedDict
 
 import requests
 import requests.exceptions
@@ -25,6 +25,13 @@ def python_version_info():
 
 
 LABELBOX_CALL_PATTERN = re.compile(r"/labelbox/")
+TEST_FILE_PATTERN = re.compile(r".*test.*\.py$")
+
+
+class _RequestInfo(TypedDict):
+    prefix: str
+    class_name: str
+    method_name: str
 
 
 def call_info():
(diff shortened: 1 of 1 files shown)


libs/lbox-clients/src/lbox/request_client.py
```diff
@@ -22,6 +24,39 @@ def python_version_info():
...
+        for stack in reversed(inspect.stack()):
+            if LABELBOX_CALL_PATTERN.search(stack.filename):
+                call_info = stack
+                method_name = call_info.function
+                class_name = call_info.frame.f_locals.get(
+                    "self", None
+                ).__class__.__name__
+
+                if method_name not in skip_methods and class_name not in skip_classes:
+                    if "test" in call_info.filename:
```
sfendell-labelbox: Can we test for a match against `.*_test.py` or something, so we don't accidentally catch `latest_whatever.py` (e.g.) here

libs/lbox-clients/src/lbox/request_client.py
```diff
@@ -186,6 +221,9 @@ def convert_value(value):
             if files:
                 del headers["Content-Type"]
                 del headers["Accept"]
+            headers["X-SDK-Method"] = (
```
sfendell-labelbox: Kind of a nit: This is the only place we use call_info right? Can we just have it create the string directly? Or if you prefer it like this, at least replace the three calls with a single call?

tomislav-peharda (approved):

libs/lbox-clients/src/lbox/request_client.py
vbrodsky: good point ok

sfendell-labelbox (approved):

PR #1852, opened 2024-10-02
[PLT-1490] Removed data row ids list on some project methods and removed get queue_modes

Gabefire:
# Description
Removed the following
1. _update_queue_mode
2. get_queue_mode
3. set_labeling_parameter_overrides (parameter non data identifiers)
4. update_data_row_labeling_priority(parameter non data identifiers)
5. move_data_rows_to_task_queue(parameter non data identifiers)


libs/labelbox/tests/unit/test_unit_project_validate_labeling_parameter_overrides.py
```diff
@@ -6,19 +6,6 @@
 from labelbox.schema.project import validate_labeling_parameter_overrides
 
 
-def test_validate_labeling_parameter_overrides_valid_data():
```
Gabefire, line 9: Not needed anymore since you only can use DataRowIdentifiers

Gabefire: No overrides

adrian-chang (approved):

PR #1853, opened 2024-10-02
[PLT-1487] Remaining sdk deprecated items removed

Gabefire:
# Description

Removing:
get_data_row_ids_for_global_keys - decide whether to update the deprecation notice and rdoc if the change in the api took place OR remove the deprecation notice

LabelGenerator assign_feature_schema_ids

AttachmentType remove TEXT

DataRowMetadataOntology bulk_delete. bulk_export  - we will keep strings, remove deprecation warnings

Dataset create_data_rows_sync

QueueMode (value=='CATALOG')

CatalogSlice get_data_row_ids

NOTE: I completely removed queue mode as a field on this PR. No project exists in Labelbox that is not Batch since we migrated all projects. You can not have any other queue mode besides Batch so no point of including. The API side was giving me some weird responses


libs/labelbox/src/labelbox/schema/dataset.py
```diff
@@ -166,47 +166,13 @@ def create_data_row(self, items=None, **kwargs) -> "DataRow":
...
-        return None  # Return None if no exception is raised
-
     def _create_data_rows_sync(
         self, items, file_upload_thread_count=FILE_UPLOAD_THREAD_COUNT
     ) -> "DataUpsertTask":
         max_data_rows_supported = 1000
         if len(items) > max_data_rows_supported:
             raise ValueError(
-                f"Dataset.create_data_rows_sync() supports a max of {max_data_rows_supported} data rows."
+                f"Dataset._create_data_rows_sync() supports a max of {max_data_rows_supported} data rows."
```
vbrodsky: We can remove the check for `max_data_rows_supported` and the exception, because all data row creation is async under the hood
Besides, this method is now only used in `create_data_row` that creates just 1 data row

vbrodsky (approved):

PR #1864, opened 2024-10-10
Added a member model to manage users better

Gabefire:
# Description

* This works similar to user groups
* Need to find a clever way to test this since it looks like staging orgs only have current users and you can not edit current user
* Unable to test deleting users for obvious reason
* All test run locally and deleting works locally

## Type of change

## All Submissions

## New Feature Submissions


libs/labelbox/src/labelbox/schema/member.py
```diff
@@ -0,0 +1,433 @@
...
+    name: Optional[str] = Field(default=None, frozen=True)
+    nickname: Optional[str] = Field(default=None, frozen=True)
+    picture: Optional[str] = Field(default=None, frozen=True)
+    is_viewer: bool = Field(default=False, frozen=True)
+    is_external_user: bool = Field(default=False, frozen=True)
+    accessible_project_count: Optional[int] = Field(default=None, frozen=True)
+    project_memberships: Set[ProjectMembership] = Field(default=set())
+    can_access_all_projects: bool = Field(default=False)
+    default_role: Optional[Role] = Field(default=None)
+    user_group_ids: Set[str] = Field(default=set())
```
Gabefire, line 83: Avoided making this user group objects since it would sort of get out of control.

libs/labelbox/src/labelbox/schema/member.py
```diff
@@ -0,0 +1,433 @@
...
+    field_validator,
+    model_serializer,
+    model_validator,
+)
+from labelbox.utils import _CamelCaseMixin
+from labelbox.schema.role import Role
+from labelbox import Client
+
+
+class ProjectMembership(_CamelCaseMixin):
```
mihhail-m, line 17: There is already exist some sort `ProjectRole` class for the same purpose. Can't that be refactored and re used? Otherwise SDK once again would have similar methods/classes for the same purpose. Or is there plan to deprecate the former?

libs/labelbox/src/labelbox/schema/member.py
```diff
@@ -0,0 +1,433 @@
...
+
+    @model_serializer()
+    def serialize_model(self):
+        return {
+            "projectId": self.project_id,
+            "roleId": None if self.role is None else self.role.uid,
+        }
+
+
+class Member(_CamelCaseMixin):
```
mihhail-m, line 39: SDK already have [class User](https://github.com/Labelbox/labelbox-python/blob/a75c402c2a6854b59ebab61d4950db5cb1202db7/libs/labelbox/src/labelbox/schema/user.py#L9). What would be the difference? Can be `User` class just refactored to have more fields and support for new methods? Otherwise there is again this weird state where SDK going to have two types to represent the same object.

libs/labelbox/src/labelbox/schema/member.py
```diff
@@ -0,0 +1,433 @@
...
+                }
+            }
+        """
+        params = {"userId": self.id}
+
+        result = self.client.execute(query, params, experimental=True)
+        if not result:
+            raise ResourceNotFoundError(
+                message="Failed to find user as user does not exist"
+            )
```
mihhail-m: i don't think this is needed here. `client.execute` already handles `ResourceNotFoundError`. In such case i think this code is simply not reachable

libs/labelbox/src/labelbox/schema/member.py
```diff
@@ -0,0 +1,433 @@
...
+            raise ValueError("Unable to update self")
+
+        try:
+            result = self.client.execute(query, params, experimental=True)
+            if not result:
+                raise ResourceNotFoundError(
+                    message="Failed to update member as member does not exist"
+                )
+        except MalformedQueryException as e:
+            raise UnprocessableEntityError("Failed to update member") from e
```
mihhail-m: this is already handled by `.execute` and raising new exceptions would just hide original error messages

libs/labelbox/src/labelbox/schema/member.py
```diff
@@ -0,0 +1,433 @@
...
+            """
+
+        if self.id == self._current_user_id:
+            raise ValueError("Unable to delete self")
+
+        params = {"id": self.id}
+
+        result = self.client.execute(query, params, experimental=True)
+        if not result:
+            raise ResourceNotFoundError(
```
mihhail-m: same here

libs/labelbox/src/labelbox/schema/member.py
Gabefire: I attempted this it does not work since we have to use a GetMemberRework call to make sure we get all the fields including ProjectMemberships. We also need to setUserAccess Call which is not supported by the built in entities user groups had a similar issues

libs/labelbox/src/labelbox/schema/member.py
Gabefire: https://github.com/Labelbox/labelbox-python/blob/6fb945c73baf94b428de3a234e6fe9a315ebd7a5/libs/labelbox/src/labelbox/schema/project.py#L1652
This won't work for this class since the relationships are off. Plus, ProjectRole and ProjectMemberships are technically two different objects in our backend.

libs/labelbox/src/labelbox/schema/member.py
Gabefire: Also this is 100% experimentally by definition can be deleted and removed any time. We should deprecate user objects if we decided to remove Members from experiments but right now the SDK is severally lacking on member management so for Alignerr for example i have to heavily use GQL direclty

libs/labelbox/src/labelbox/schema/member.py
```diff
@@ -0,0 +1,433 @@
...
+        previous_batch = 0
+        batch_size = 100
+        while True:
+            params = {
+                "first": batch_size,
+                "skip": previous_batch,
+                "search": search,
+                "roleIds": [role.uid for role in roles] if roles else None,
+                "groupIds": group_ids,
+                "complexFilters": None,
```
Gabefire, line 403: These complexFilters are associated with Alignerr and can be added to support getting members based on filters

libs/labelbox/src/labelbox/schema/member.py
Gabefire: Fixed

libs/labelbox/src/labelbox/schema/member.py
Gabefire: Fixed

PR #1876, opened 2024-10-17
[PLT-1587] Removed all unused imports

Gabefire:
# Description

* Added in Flake rule that does not allow unused imports
* Removed all unused imports
* Ignore other rules will add those in a separate PR

### DIFF
--- a/libs/labelbox/pyproject.toml
+++ b/libs/labelbox/pyproject.toml
@@ -74,7 +74,7 @@ dev-dependencies = [
 line-length = 80
 
 [tool.ruff.lint]
-ignore = ["F", "E722"]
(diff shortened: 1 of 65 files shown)


libs/labelbox/pyproject.toml
```diff
@@ -74,7 +74,7 @@ dev-dependencies = [
 line-length = 80
 
 [tool.ruff.lint]
-ignore = ["F", "E722"]
+ignore = ["F841", "E722", "F811", "F402", "F601", "F403", "F821", "F541"]
```
Gabefire, line 77: Will add the rest of F rules in the next PR and then we will be down to the defaults for Ruff besides bare exception statement

sfendell-labelbox (approved):

PR #1909, opened 2024-11-16
[PLT-1677] Added get mal import functions to replace old BulkImportRequest class

Gabefire:
# Description

* Two distinct annotation jobs
   1. Label imports
   2. Prediction imports
* included get methods for both and delete path for prediction import

## Type of change

## All Submissions

### DIFF
--- a/libs/labelbox/src/labelbox/schema/annotation_import.py
+++ b/libs/labelbox/src/labelbox/schema/annotation_import.py
@@ -587,6 +587,20 @@ class MALPredictionImport(CreatableAnnotationImport):
         """
         return self.project().uid
 
+    def delete(self) -> None:
+        """
+        Deletes a MALPredictionImport job
+        """
+
+        query_string = """
+            mutation deleteModelAssistedLabelingPredictionImportPyApi($id: ID!) {
+                deleteModelAssistedLabelingPredictionImport(where: { id: $id }) {
+                        id
+                    }
+                }
+            """
+        self.client.execute(query_string, {"id": self.uid})
+
     @classmethod
     def create_from_file(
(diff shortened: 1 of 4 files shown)


libs/labelbox/src/labelbox/schema/project.py
```diff
@@ -710,6 +711,56 @@ def _connect_default_labeling_front_end(self, ontology_as_dict: dict):
...
+
+        return PaginatedCollection(
+            self.client,
+            query_str,
+            {id_param: self.uid},
+            ["modelAssistedLabelingPredictionImports"],
+            MALPredictionImport,
+        )
+
+    def get_label_imports(self) -> PaginatedCollection:
```
Gabefire, line 739: Note: Not going to add a delete path for regular label imports. This is because the mutation is internal and it can also put the data row in a bad state i.e. label with no annotations. We also already have SDk methods to delete and requeue labels so not needed.

libs/labelbox/src/labelbox/schema/annotation_import.py
```diff
@@ -587,6 +587,20 @@ def parent_id(self) -> str:
         """
         return self.project().uid
 
+    def delete(self) -> None:
```
Gabefire, line 590: I can't use the deletable inheritance to make this work for this class due to this being a hard delete, not flipping a field to False.

tomislav-peharda:
Im okay approving because of codeowner requirement, but might be useful getting 👀 from someone from Model team

apollonin (approved):
lgtm

tomislav-peharda (approved):

PR #1910, opened 2024-11-17
[PLT-1725] Changed ontology dataclasses to pydantic dataclasses

Gabefire:
# Description

* Pydantic V2 has a standin replacement for [python dataclasses](https://docs.pydantic.dev/latest/concepts/dataclasses/) but with more features and better validation.
* Changed ontology creation to take advantage of pydantic features

For example for this code:
```python
classification_features = [
    lb.Classification(
        class_type="test",
        name=1,
        options=[
            lb.Option(value="blurry", label="Blurry"),
            lb.Option(value="distorted", label="Distorted"),
        ],
    )
]

ontology_builder = lb.OntologyBuilder(classifications=classification_features).asdict()
```
`class_type` and `name` is given the wrong type for the classification feature (string instead of an enum value and int instead of string). Before change, it would give this error
image
With pydantic validation it gives this error (catching the name being an int as well)
image

## Type of change

## All Submissions

### DIFF
--- a/libs/labelbox/src/labelbox/schema/slice.py
+++ b/libs/labelbox/src/labelbox/schema/slice.py
@@ -1,5 +1,7 @@
-from dataclasses import dataclass
-from typing import Optional, Tuple, Union
+from pydantic import ConfigDict, GetCoreSchemaHandler
+from pydantic.dataclasses import dataclass
+from pydantic_core import core_schema
+from typing import Any, Optional, Tuple, Union
 import warnings
 from labelbox.orm.db_object import DbObject
 from labelbox.orm.model import Field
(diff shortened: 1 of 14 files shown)


libs/labelbox/src/labelbox/schema/slice.py
```diff
@@ -47,6 +49,34 @@ def to_hash(self):
                 "global_key": self.global_key.key if self.global_key else None,
             }
 
+        model_config = ConfigDict(arbitrary_types_allowed=True)
+
+        @classmethod
+        def __get_pydantic_core_schema__(
```
vbrodsky: what does this do

Gabefire: This essentially mainly validated the model. Th issue is the global_key and unique id classes are ABC which is causing issues with pydantic. I went ahead and remove this since it is getting to complicated and losing the advantages of going all Pydantic.

PR #1918, opened 2024-11-22
[PLT-1993] Fixed relationship label bug

Gabefire:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context.

* Relationship annotations are broken if the source is used in more then one relationship
* this pr looks to simplify everyting
* All test are working besides the MEA one this PR will not allow extra keys in object annotations anymore but this was never mentioned in our docs and is not required for anything

## Type of change

## All Submissions

## New Feature Submissions

## Changes to Core Features

### DIFF
--- a/libs/labelbox/src/labelbox/data/serialization/ndjson/label.py
+++ b/libs/labelbox/src/labelbox/data/serialization/ndjson/label.py
@@ -1,7 +1,9 @@
 from collections import defaultdict
+import copy
 from itertools import groupby
 from operator import itemgetter
-from typing import Dict, Generator, List, Tuple, Union
+from typing import Generator, List, Tuple, Union
+from uuid import uuid4
 
 from pydantic import BaseModel
 
@@ -9,8 +11,7 @@ from ...annotation_types.annotation import (
     ClassificationAnnotation,
     ObjectAnnotation,
 )
-from ...annotation_types.collection import LabelCollection, LabelGenerator
-from ...annotation_types.data.generic_data_row_data import GenericDataRowData
+from ...annotation_types.collection import LabelCollection
 from ...annotation_types.label import Label
 from ...annotation_types.llm_prompt_response.prompt import (
     PromptClassificationAnnotation,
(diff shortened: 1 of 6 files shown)


libs/labelbox/tests/data/annotation_import/test_relationships.py
```diff
@@ -0,0 +1,217 @@
+import datetime
```
Gabefire, line 1: Relationships work different then other relationship so need a dedicated integration test

libs/labelbox/src/labelbox/data/serialization/ndjson/label.py
```diff
@@ -60,143 +60,19 @@
 class NDLabel(BaseModel):
     annotations: AnnotationType
 
-    class _Relationship(BaseModel):
```
Gabefire, line 63: This stuff was used when we use to deserialized so not needed anymore also is very complicated

libs/labelbox/tests/data/serialization/ndjson/test_relationship.py
```diff
@@ -1,165 +1,194 @@
...
-    Point,
-    Rectangle,
     RelationshipAnnotation,
     Relationship,
+    TextEntity,
 )
 
 
-def test_relationship():
-    with open("tests/data/assets/ndjson/relationship_import.json", "r") as file:
```
Gabefire, line 19: Simplified this test into two

libs/labelbox/tests/data/annotation_import/test_mea_prediction_import.py
```diff
@@ -233,19 +230,13 @@ def test_create_from_label_objects(
                     ),
                     ObjectAnnotation(
                         name="bbox",
-                        extra={
```
Gabefire, line 236: This could be breaking but want a second opinion. The extra field breaks the annotation if a customer decides to add it. It is not shown that they need to do this in the docs, and I don't see why they would. The old complicated logic removed these but to simplify this I think it is best to just not support this type of extra key

adrian-chang (approved):

PR #1971, opened 2025-04-28
[SDK-0] Disabling test

mihhail-m:
Disabling test that has been failing for weeks and disrupting deployments process. 
Test appears to be incomplete as there is no fixture for provisioning client wit restricted permissions for now hence skipping until further notice.


lb-pno (commented):
If the test is not required anymore, it's probably better to delete it.

midhun-pm (approved):

PR #2074, opened 2026-08-21
[PLT-4347] Route embedding metadata operations through GraphQL

abacchilb:
## Problem

New API keys are opaque `lbx_` secrets. lb-api accepts both legacy JWT keys and opaque keys, but the SDK's embedding methods bypass lb-api and send the customer's bearer token directly to ADV. ADV assumes every token is a JWT, returns a non-JSON 401, and `AdvClient` surfaces that as `JSONDecodeError`.

Fixes [PLT-4347](https://labelbox.atlassian.net/browse/PLT-4347).

## Change

Keep the public SDK API unchanged, but route embedding metadata operations through lb-api's existing GraphQL resolvers:

- `Client.create_embedding()` → `createEmbedding`
- `Client.get_embeddings()` → `embeddings`
- `Client.get_embedding_by_id()` → filters the bounded GraphQL list (organizations are limited to 10 embeddings)
- `Embedding.delete()` → `deleteEmbedding`

`Embedding` now receives the top-level `Client`, which implements the small interface needed by the model.

No platform/ADV authentication changes and no nginx routing changes are required.

## Follow-up: remaining direct ADV calls

These data-plane operations intentionally remain on `AdvClient`:

- `PUT /adv/v1/embeddings/{id}/_import_ndjson` (`Embedding.import_vectors_from_file`)
- `GET /adv/v1/embeddings/{id}/vectors/_count` (`Embedding.get_imported_vector_count`)

They still require a legacy JWT API key today. We are checking Datadog usage before deciding whether to add narrow lb-api proxy endpoints, migrate them another way, or deprecate them. `AdvClient` has been reduced to only those remaining operations so the boundary is explicit.

## Verification

- Full `libs/labelbox/tests/unit` suite with data extras: **291 passed**
- Added coverage for GraphQL create/list/lookup/delete and for the two operations intentionally retained on ADV
- Ruff format check passes for all changed files
- IDE diagnostics: no errors

Made with [Cursor](https://cursor.com)

[PLT-4347]: https://labelbox.atlassian.net/browse/PLT-4347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/labelbox/src/labelbox/adv_client.py
+++ b/libs/labelbox/src/labelbox/adv_client.py
@@ -1,7 +1,7 @@
 import io
 import json
 import logging
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, Optional
 from urllib.parse import urlparse
 
 import requests
@@ -16,19 +16,6 @@ class AdvClient:
         self.endpoint = endpoint
         self.api_key = api_key
         self.session = self._create_session()
-
-    def create_embedding(self, name: str, dims: int) -> Dict[str, Any]:
-        data = {"name": name, "dims": dims}
-        return self._request("POST", "/adv/v1/embeddings", data)
-
-    def delete_embedding(self, id: str):
-        return self._request("DELETE", f"/adv/v1/embeddings/{id}")
-
-    def get_embedding(self, id: str) -> Dict[str, Any]:
-        return self._request("GET", f"/adv/v1/embeddings/{id}")
-
-    def get_embeddings(self) -> List[Dict[str, Any]]:
-        return self._request("GET", "/adv/v1/embeddings").get("results", [])
 
     def import_vectors_from_file(self, id: str, file_path: str, callback=None):
         self._send_ndjson(
--- a/libs/labelbox/src/labelbox/client.py
+++ b/libs/labelbox/src/labelbox/client.py
@@ -2150,8 +2150,21 @@ class Client:
         Returns:
             A new Embedding object.
         """
-        data = self._adv_client.create_embedding(name, dims)
-        return Embedding(self._adv_client, **data)
+        mutation = """
+        mutation CreateEmbeddingPyApi($data: CreateEmbeddingInput!) {
+            createEmbedding(data: $data) {
+                id
+                name
+                dims
+                custom
+            }
+        }
+        """
+        data = self.execute(
+            mutation,
+            {"data": {"name": name, "dims": dims}},
+        )["createEmbedding"]
+        return Embedding(self, **data)
 
     def get_embeddings(self) -> List[Embedding]:
         """
@@ -2160,8 +2173,18 @@ class Client:
         Returns:
             A list of embedding objects.
         """
-        results = self._adv_client.get_embeddings()
-        return [Embedding(self._adv_client, **data) for data in results]
+        query = """
+        query GetEmbeddingsPyApi {
+            embeddings {
+                id
+                name
+                dims
+                custom
+            }
+        }
+        """
+        results = self.execute(query)["embeddings"]
+        return [Embedding(self, **data) for data in results]
 
     def get_embedding_by_id(self, id: str) -> Embedding:
         """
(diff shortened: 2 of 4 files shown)


abacchilb:
## Datadog follow-up on remaining direct ADV calls

I queried `service:lb-catalog-adv-primary` across hot + Flex logs with a requested 90-day window. Effective log visibility began around 2026-08-06 (~15 days), and no matching APM spans were indexed, so this is evidence within retention rather than a proof of lifetime non-use.

- **NDJSON import:** 36 successful `saving raw embeddings data` events, all in staging (`lb-staging-na-us-env0`), from 2026-08-07 through 2026-08-19. Zero successful import events in production. This pattern is consistent with the SDK integration suite, though available logs do not expose user-agent/caller metadata to prove attribution.
- **Vector count:** no successful request signal found. The only production ADV match was one unauthenticated probe on 2026-08-07 (`/v1/embeddings/x/vectors/_count`); a separate browser request hit a nonexistent lb-api path and returned 404.
- **Metadata endpoints:** direct `/adv/v1/embeddings` traffic is clearly active and failing with opaque keys: 873 logged errors in staging and 410 in production during the visible window. Those errors are what this PR removes for create/list/lookup/delete.

Recommendation: keep NDJSON import and vector count as an explicit follow-up rather than blocking PLT-4347. Before removing them, either run a longer-retention ingress/access-log analysis or announce deprecation; if they must support opaque keys, add narrow lb-api data-plane endpoints rather than reopening broad ADV customer-key auth.

abacchilb:
## Staging CI: red checks are pre-existing, not this PR

The GraphQL embedding path is working. After cleanup, all live embedding tests passed against staging:

- `test_get_embedding_by_id`
- `test_get_embedding_by_name_not_found`
- `test_import_vectors_from_file[10]`
- `test_get_imported_vector_count`

Lint, Bugbot, TruffleHog, Snyk, Test-PyPI, and test-container are also green.

### What we tried

1. Original run hit `ApiLimitError: You have exceeded 15000 requests in a 60 second time window` (5 Python versions × xdist, one staging org).
2. Re-ran failed jobs after the window cooled. Rate limits cleared. Remaining noise: `Max limit of custom embeddings reached : 10 per organization` at fixture setup.
3. Dispatched [Staging Embedding Cleanup](https://github.com/Labelbox/labelbox-python/actions/workflows/staging-embedding-cleanup.yml) (dry-run then real). Deleted one stale embedding (`sdk-int-ci-v2-1785232085-46ebef0e35`). Re-ran failed jobs again.

3.11 after cleanup ([run 32502258457](https://github.com/Labelbox/labelbox-python/actions/runs/32502258457)):

| Job | Result |
| --- | --- |
| integration | **5 failed, 306 passed, 0 errors** |
| unit-data | **9 failed, 272 passed, 0 errors** |

Zero embedding errors remaining.

### Remaining failures (not this change)

**unit-data** — all 9 in `tests/data/annotation_import/test_mea_prediction_import.py`:

```
LabelboxError: 9 FAILED_PRECONDITION: MEA prediction import has been deprecated
```

Staging turned the feature off; tests were never updated.

**integration** — same 5 tests as on `develop` historically:

- `test_delete_non_existent_schema_id[data_row_global_key / data_row_id_as_str / data_row_unique_id]` (failing since at least 2026-07-22)
- `test_semantic_error`, `test_query_complexity_error` → `InvalidQueryError: GraphQL validation error` (staging schema drift)

### Historical `develop` (this is not new)

The last `develop` runs that actually executed SDK tests all failed. The two later “green” runs (Jul 29, Aug 3) skipped the `build` job via `path-filter` because they only touched Alignerr/publish files.

| Run | Date | Integration |
| --- | --- | --- |
| [29933222859](https://github.com/Labelbox/labelbox-python/actions/runs/29933222859) | Jul 22 | 4 failed, 307 passed — same `test_delete_non_existent_schema_id` trio |
| [30362386905](https://github.com/Labelbox/labelbox-python/actions/runs/30362386905) | Jul 28 | 4 failed, 307 passed |
| this PR | Aug 21 | 5 failed, 306 passed after cleanup |

This branch is not regressing the suite. Further re-runs will not go green until those staging tests are fixed on `develop`.

mrobers1982 (approved):

